PR #3, opened 2016-03-09
added: multithreaded assembly for mixed structured models

akva2:
This adds multithreaded assembly for mixed models.


akva2:
merged manually with suggestions and fixes applied.

PR #4, opened 2016-03-09
Add support for dune-istl solvers

akva2:
Supported in serial only. We assemble into a sparse matrix, and then copy into a dune matrix for solve.


akva2:
included in #7

PR #5, opened 2016-03-09
fixed: mark functions as non-copyable

akva2:
lots of pointer members so these cannot be copied


akva2:
pushed manually with doxy.

PR #9, opened 2016-04-06
Inhomogeneous Dirichlet

timovanopstal:


timovanopstal:
jenkins build this please

timovanopstal:
jenkins build this please

timovanopstal:
test could use a closer look, I'm not familiar with gtest and its likely things can be done in a better way.

timovanopstal:
merging manually

timovanopstal:
awaiting @kmokstad 's approval

src/LinAlg/SAM.C
```diff
@@ -809,7 +809,8 @@ bool SAM::applyDirichlet (Vector& dofVec) const
     {
       int ip = mpmceq[iceq-1];
       dofVec[idof] = ttcc[ip-1];
-    }
+    } else if (iceq == 0)
+      dofVec[idof] = 0.0;
```
kmokstad: Have to think about this one, when I have the source code in front of me. It looks OK, but I can't see what this has to do with mixed problems? Looks more like a general error (if it is an error). iceq is zero when a DOF is fixed and I though in this case it should be necessary to do anything since dofVec should already have a zero value. Or can you explain why this is not so, only when doing mixed.

akva2: yes it's general and not related to mixed.

kmokstad:
IC, this method was introduced with the explicit time integration drivers, for which the input-vector is a linear combination of linear solution vectors (which all have or should have zero value for the fixed dofs). But if it now also is used on initial conditions read from file or generated by other means, this is no longer guarantied and the fix is in place. I was confused since I thought originally this was one of my methods used after SAM::expandVector, but that was not the case.

So the fix is good, only the commit message is not, as this has nothing to do with mixed. It is to enforce homogeneous Dirichlet conditions on initial conditions, isn't it.

akva2:
precisely

timovanopstal:
merging manually

PR #12, opened 2016-04-13
added: method to add a global lagrange multiplier to the system

timovanopstal:


timovanopstal:
jenkins build this please

timovanopstal:
jenkins build this please

src/ASM/ASMs3Dmx.h
```diff
@@ -199,6 +199,14 @@ class ASMs3Dmx : public ASMs3D, private ASMmxBase
...
   //! \param[in] basis Which basis to return size parameters for
   virtual bool getSize(int& n1, int& n2, int& n3, int basis = 0) const;
+protected:
+  //! \brief Returns the volume in the parameter space for an element.
+  //! \param[in] iel Element index
+  double getParametricVolume(int iel) const;
+  //! \brief Returns boundary face area in the parameter space for an element.
+  //! \param[in] iel Element index
+  //! \param[in] dir Local face index of the boundary face
+  double getParametricArea(int iel, int dir) const;
```
kmokstad: This two methods overload the simular methods in ASMs3D, but then they need to be declared virtual both here and there, or else I'm nor sure you get the right one when invoked from a parent class.

akva2: virtual is only needed in the class being overridden as such.

-------- Opprinnelig melding --------
Fra: kmokstad <notifications@github.com>
Dato: 14.04.2016 17:43 (GMT+01:00)
Til: OPM/IFEM <IFEM@noreply.github.com>
Emne: Re: [OPM/IFEM] added: method to add a global lagrange multiplier to the system (#12)


In src/ASM/ASMs3Dmx.h<https://github.com/OPM/IFEM/pull/12#discussion_r59740756>:

> @@ -199,6 +199,14 @@ class ASMs3Dmx : public ASMs3D, private ASMmxBase
>    //! \param[out] n3 Number of nodes in third (w) direction
>    //! \param[in] basis Which basis to return size parameters for
>    virtual bool getSize(int& n1, int& n2, int& n3, int basis = 0) const;
> +protected:
> +  //! \brief Returns the volume in the parameter space for an element.
> +  //! \param[in] iel Element index
> +  double getParametricVolume(int iel) const;
> +  //! \brief Returns boundary face area in the parameter space for an element.
> +  //! \param[in] iel Element index
> +  //! \param[in] dir Local face index of the boundary face
> +  double getParametricArea(int iel, int dir) const;


This two methods overload the simular methods in ASMs3D, but then they need to be declared virtual both here and there, or else I'm nor sure you get the right one when invoked from a parent class.

akva2: And it is only called in specialized methods (integrate) so there is no need for the method to be virtual. But if you think it will ever be called in the parent context later..

-------- Opprinnelig melding --------
Fra: kmokstad <notifications@github.com>
Dato: 14.04.2016 17:43 (GMT+01:00)
Til: OPM/IFEM <IFEM@noreply.github.com>
Emne: Re: [OPM/IFEM] added: method to add a global lagrange multiplier to the system (#12)


In src/ASM/ASMs3Dmx.h<https://github.com/OPM/IFEM/pull/12#discussion_r59740756>:

> @@ -199,6 +199,14 @@ class ASMs3Dmx : public ASMs3D, private ASMmxBase
>    //! \param[out] n3 Number of nodes in third (w) direction
>    //! \param[in] basis Which basis to return size parameters for
>    virtual bool getSize(int& n1, int& n2, int& n3, int basis = 0) const;
> +protected:
> +  //! \brief Returns the volume in the parameter space for an element.
> +  //! \param[in] iel Element index
> +  double getParametricVolume(int iel) const;
> +  //! \brief Returns boundary face area in the parameter space for an element.
> +  //! \param[in] iel Element index
> +  //! \param[in] dir Local face index of the boundary face
> +  double getParametricArea(int iel, int dir) const;


This two methods overload the simular methods in ASMs3D, but then they need to be declared virtual both here and there, or else I'm nor sure you get the right one when invoked from a parent class.

kmokstad: I know, but I prefer to declare virtual in the subclass also, such that it is clear that it is indeed virtual, without looking at the parent class definition.

akva2: Which has more meaning than just cosmetics. But anyways my main question is why it should be virtual in the first place

-------- Opprinnelig melding --------
Fra: kmokstad <notifications@github.com>
Dato: 14.04.2016 17:51 (GMT+01:00)
Til: OPM/IFEM <IFEM@noreply.github.com>
Ko: Arne Morten Kvarving <Arne.Morten.Kvarving@sintef.no>
Emne: Re: [OPM/IFEM] added: method to add a global lagrange multiplier to the system (#12)


In src/ASM/ASMs3Dmx.h<https://github.com/OPM/IFEM/pull/12#discussion_r59741957>:

> @@ -199,6 +199,14 @@ class ASMs3Dmx : public ASMs3D, private ASMmxBase
>    //! \param[out] n3 Number of nodes in third (w) direction
>    //! \param[in] basis Which basis to return size parameters for
>    virtual bool getSize(int& n1, int& n2, int& n3, int basis = 0) const;
> +protected:
> +  //! \brief Returns the volume in the parameter space for an element.
> +  //! \param[in] iel Element index
> +  double getParametricVolume(int iel) const;
> +  //! \brief Returns boundary face area in the parameter space for an element.
> +  //! \param[in] iel Element index
> +  //! \param[in] dir Local face index of the boundary face
> +  double getParametricArea(int iel, int dir) const;


I know, but I prefer to declare virtual in the subclass also, such that it is clear that it is indeed virtual, without looking at the parent class definition.

timovanopstal:
Ok, I guess there are two things going on:
1. Discrepancy with the 2D counterparts (not mentioned in this thread)
-> So the question is whether that does what we want...
2. Confusion over what happens when a function is virtual/non virtual (and what we want)
-> In my understanding the virtual keyword tells derived instances to check if they have their own implementation, so a) Adding virtual in `ASMs3Dmx` doesn't change behavior but is a helpful convention acc to kmokstad. b) On the other hand adding virtual in the parent changes what is being called in the parent context (if ever this is done in the future)

I've been thinking, I can understand we would like virtuals in `ASMs3D` to be "future proof", but I'm not sure about adopting a convention which hasn't been done consistently (cf. eg. `ASMs2Dmx::getParametricLength()` which is a derived's implementation of a virtual of the parent too). Since this convention is not yet established I would prefer not being sent around all the place by virtuals which are pointing nowhere (i.e. when classes aren't derived/have children reimplementing those members) and starting a github issue on it instead?

timovanopstal:
removed superfluous virtuals in ASM2D{,mx}::getParam* as discussed

timovanopstal:
jenkins build this please

timovanopstal:
manual merge alert

PR #17, opened 2016-04-25
fixed: fix boundary condition output to VTF for mixed bases

akva2:
These are only generated for the first basis, not for the geometry basis.

### DIFF
--- a/src/SIM/SIMoutput.C
+++ b/src/SIM/SIMoutput.C
@@ -333,7 +333,7 @@ bool SIMoutput::writeGlvBC (int& nBlock, int iStep) const
 
     geomID++;
     size_t nbc = myModel[i]->getNoFields(1);
-    Matrix bc(nbc,myModel[i]->getNoNodes(-1));
+    Matrix bc(nbc,myModel[i]->getNoNodes(1));
     RealArray flag(3,0.0);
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

kmokstad:
Bør vel fikses tilsvarende i ASMs3Dmx også i samme commit

akva2:
det har du helt rett i. lagt til.

akva2:
jenkins build this please

timovanopstal:
looking great! merging manually

src/SIM/SIMoutput.C
```diff
@@ -333,7 +333,7 @@ bool SIMoutput::writeGlvBC (int& nBlock, int iStep) const
 
     geomID++;
     size_t nbc = myModel[i]->getNoFields(1);
-    Matrix bc(nbc,myModel[i]->getNoNodes(-1));
+    Matrix bc(nbc,myModel[i]->getNoNodes(1));
```
kmokstad: This one actually not so great. It causes one (or at least one) of the FiniteDeformation tests with contact to crash, because the bc array now becomes too small. Was not caught because these tests have not been in the loop after the repo-split.

PR #18, opened 2016-04-26
Add a dedicated DomainComposition helper

akva2:
this administers the parallel finite element model.
    - establishes global node numbers
    - establishes global equation numbers
    - establishes global DOF numbers
    - generates subdomains for schwarz preconditioners (if using
      multiple subdomains for each patch)

It is substantially tested, though coverage for mixed 3D is still a bit sparse. But I'm fairly confident it works fine.

Some preparatory commits are included;
1) Change getBoundaryNodes to allow passing which basis to grab nodes for.
2) Fix &lt;refine&gt; and &lt;raiseorder&gt; keywords to work with parallel (renumbering of patches).
3) change from PARALLEL_PETSC to a generic HAVE_MPI preprocessor token. MPI support will  not be tied to PETSc in the future.
4) add a simple framework for parallel unit tests

I have not yet tested that 1) does not break FSI (which is the other user of these functions). I will get to that asap.

This is the first major step in the parallel framework reboot. It does not use the new helper yet (but adds a member where it is intended to enable the unit tests).


akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
tested FSI. it breaks, but in a very trivial way. the old getBoundaryCodes used to check the old map and only push if a node was unique. i fixed that externally where the uniquification makes more sense in SIMALE, using
``` 
// need to make these unique
std::vector<int> glbNodes2;
for (const auto& it : glbNodes)
  if (std::find(glbNodes2.begin(), glbNodes2.end(), it) == glbNodes2.end())
    glbNodes2.push_back(it);
```

akva2:
scratch that. fixed it in the SIM wrapper method instead as the SIMALE solution had problems elsewhere (SIMBeam et al). that update is included in this PR now, as well as a test documenting the expected behavior.

akva2:
superseeded by #7

PR #23, opened 2016-05-20
Supg mixed

timovanopstal:
a few things needed in the NavierStokes app (shortly)

### DIFF
--- a/src/Utility/Tensor.C
+++ b/src/Utility/Tensor.C
@@ -710,6 +710,18 @@ Tensor operator* (const Tensor& A, const Tensor& B)
 }
 
 
+Tensor operator*(const double c, const Tensor& T)
+{
+  Tensor C(T.n);
+
+  for (int i = 1; i <= C.n; ++i)
+    for (int j = 1; j <= C.n; ++j)
+      C(i,j) = c*T(i,j);
+
+  return C;
+}
+
+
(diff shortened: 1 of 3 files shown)


timovanopstal:
jenkins build this please

src/Utility/Tensor.C
```diff
@@ -710,6 +710,18 @@ Tensor operator* (const Tensor& A, const Tensor& B)
...
+  Tensor C(T.n);
+
+  for (int i = 1; i <= C.n; ++i)
+    for (int j = 1; j <= C.n; ++j)
+      C(i,j) = c*T(i,j);
+
+  return C;
+}
+
+
```
kmokstad: Since we already have the *= operator defined, this can be done better(?) just like this:
{
  Tensor C(T);
  return C *= c;
}

and use Real instead of double to be consistent with the rest of this class.

akva2:
hessian change probably need a 3D analogue.

timovanopstal:
have been tied down and unable to redo this properly, was thinking that the commented changes should actually enter IFEM in #16 though?

kmokstad:
OK, I can put it into #16 before that one is merged, before the end of the day.

kmokstad:
The items have now been merged through #16.
The commit message was misleading, as it as G-matrix and not the Hessian that was added.

PR #26, opened 2016-06-04
Added: Outprint of simulator title and the executing command.

kmokstad:
Such that it does not need to repeated in all the main programs,
and also to reflect the correct command for parallel runs.

Bare et forslag, må testes i parallel setting også.


kmokstad:
jenkins build this please

akva2:
Jeg er fan..

PR #27, opened 2016-06-14
Allow ramping up rtol in NonLinSIM

akva2:
This is a POC and not intended for merge as-is.

@kmokstad i need to know what a kosher version of this looks like.

### DIFF
--- a/src/SIM/NonLinSIM.C
+++ b/src/SIM/NonLinSIM.C
@@ -33,6 +33,9 @@ NonLinSIM::NonLinSIM (SIMbase& sim, CNORM n) : MultiStepSIM(sim), iteNorm(n)
   nupdat  = 20;
   prnSlow = 0;
   rTol    = 0.000001;
+  rTolSteps = 0;
+  rTolStep = 1.0;
+  rTolRampStart = 0;
   aTol    = 0.0;
   divgLim = 10.0;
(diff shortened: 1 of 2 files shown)


kmokstad:
Not sure when this is needed. If only in rare occasions, could it be implemented through a sub-class instead (we use to have a NSNonLinSim before, for instance), associated with that application only?

src/SIM/NonLinSIM.C
```diff
@@ -190,6 +196,9 @@ ConvStatus NonLinSIM::solveStep (TimeStep& param, SolutionMode mode,
   if (solution.empty())
     return FAILURE;
 
+  if (rTolSteps > 0 && param.step == 1)
+    rTol *= std::pow(rTolStep, -rTolSteps);
```
kmokstad: At least, consider moving this statement into the parse-method (lines 114-117) then no need to check param.step == 1

akva2:
kunne det vært en idè å skille ut dette (= sånn ca rtol, atol, dtol i NonLinSIM, NewmarkSIM, iterative lineærløsere) i et objekt, og gi en bruker-SIM mulighet til å instansere en versjon av denne klassen for bruk (med en enkel default) ?

timovanopstal:
this pr was just open for the question...

PR #29, opened 2016-06-24
fixed: accidentially removed code for non-homogenous dirichlet with LR

akva2:
this hunk got lost during refactoring in
b3ac7b81886f2bc81efd36910377b0292d47c7fc


akva2:
better fix incoming.

PR #30, opened 2016-06-27
fix: follow standard definition of G-matrix.

timovanopstal:
Afaik the G-matrix definition uses a (-1,1)^d reference element, but this is mostly CFD-motivated, and this is the only place where the G-matrix is used in IFEM.


timovanopstal:
jenkins build this please

PR #31, opened 2016-06-28
Fixed: Adaptive linear elasticity was broken in the attempt to generalize ...

kmokstad:
AdaptiveSIM to also handle time-dependent problems. This is now reverted
as this class is not well suited for that class of problems.
It was intended for static linear problems only.

Essentially, the change is that AdaptiveSIM::parse invokes model.parse for all tags except "adaptive". The rest is cosmetics changes and clean-up of unused stuff.


akva2:
ja, det ser greit ut. SIMSolverAdap og Darcy-appen må oppdateres litt når vi kommer dit men det tar jeg da.

PR #32, opened 2016-06-30
added: support for second derivatives in 3D mixed ASM

akva2:
Pretty much cut-and-paste


akva2:

akva2:
i chose to ignore the immense amounts of feedback and just push it. complaints go to /dev/null.

PR #33, opened 2016-06-30
changed: hdf5 serial/parallel switch support

akva2:
requires a ubuntu 16.04 or equivalent modern distro.

this for a newer cmake, and for the ability to have the serial
and parallel hdf5 libraries coexist on the system.

finally this mess has been resolved.

@TheBB would be interesting to hear how it fares on arch (no idea if arch has such support or not).


TheBB:
The two hdf5 packages conflict on arch so I can't install both. I'd have to compile on my own.

akva2:
okay, so seems this is a debian science (ubuntu specific) solution. in either case it's a big improvement there.

akva2:
pushed.

PR #35, opened 2016-07-08
added: additional convergence check in SIMCoupledSI

akva2:
you can use this to add additional criteria for convergence


kmokstad:
Merged manually, in a slightly rewritten form.

PR #39, opened 2016-07-20
Add absolute convergence tolerance in NewmarkSIM

TheBB:
In stationary solutions with no change the relative method is problematic.


kmokstad:
Included this in 4216595c7b3843b281a0595210bdeb0b8b3daeeb in a slightly different way, but similar as in NonLinSIM (such that the behaviour is comparable), although this feature probably is not that needed for dynamic problems (only in the initial step(s), when there is no external load whatsoever, due to program/model error). But it is OK to be consistent. Also added the maxIncr parameter (number of steps with increasing convergence norm before divergence is flagged).

PR #40, opened 2016-08-04
Fixed: NavierStokes did not build because of missing includes in AppCommon.h

kmokstad:
Resolve this by moving the parse implementation to a C-file (not a template).
Also added a default constructor initialising the dim member.
Removed obsolete file ResidualOperators.C


kmokstad:

timovanopstal:
`ResidualOperators.C` is not obsolete, `Divergence` is used in a branch intended for merge with `NavierStokes`.

kmokstad:
The Divergence method is already defined in the header file. @akva2 changed all methods into templates in c200bf3e68d64a763120c5667ca2fc459ebdc4b5 but for unknown reason did not remove the C-file, only took it out from the CMakeLists.txt such that it is no longer compiled. Then it is obsolete in my view.

timovanopstal:
My bad, you're right.

PR #50, opened 2016-08-29
Fixed: Allow using the <refine> and <raiseorder> keywords in parallel

kmokstad:
Translates to local patch index and ignores patch entries on other processes.
The nGlPatches member of SIMbase is now default equal to myModel.size() for
serial runs, such that we don't need to check that it is non-zero all the time
(except for in getLocalPatchIndex, which may be called before it is set).
The latter is now treated as an error conditions and indicates missing
partitioning information for parallel runs.

Alternative to #42, checking.


kmokstad:

PR #57, opened 2016-09-05
Add periodicity support in domain decomposition

akva2:
This adds periodicity support in domain decomposition class.

This sits on top of #56 please process after that one.


akva2:
merged manually.

PR #58, opened 2016-09-05
Lower dim connections

akva2:
This adds support for lower-dimensionality connection tags. Builds on #57 so process after that.


src/ASM/ASMs3D.h
```diff
@@ -204,6 +204,9 @@ class ASMs3D : public ASMstruct, public ASM3D
   //! \brief Returns the node index for a given corner.
   virtual int getCorner(int I, int J, int K, int basis = 1) const;
 
+  //! \brief Returns the node index for a given corner.
```
kmokstad: Skal vel være: Returns the node indices for a given edge.

akva2: 22b9888cd1207765078a2303cfb214eab7a120b8

akva2:
merged manually.

PR #59, opened 2016-09-14
fix: add missing include<array>

timovanopstal:


timovanopstal:
following up on my inline comment on #56

PR #60, opened 2016-09-15
fixed: due to the order of things, we need a model generator member

akva2:
due to the structure of parsing there is no way around this.
we first need to generate the model (and topology sets) in the SIMbase
parse function. we then need to refine the model in the SIMxD parse
functions which are called after the SIMbase parsing. thus,
we need to hand the model generator instanced in the SIMbase parsing
to the SIMxD to call at the appropriate time.


akva2:
@kmokstad i see no other way here, sorry..

timovanopstal:
Ser bra ut for min del!
Det eneste jeg tenker på er at vi trenger test-dekning i Apps/Common (det er kun noen tester lagt til i en IFEM-Stokes PR nå)

akva2:
merged manually.

PR #64, opened 2016-10-06
fixed: also check opposite dof-mismatch condition

akva2:
We checked if master had an equation, and local was fixed, but not the other way around.


akva2:
merged manually.

PR #65, opened 2016-10-06
added: support centered geometric refinement 

akva2:
centered as in refinement towards both sides of the patch.


akva2:
ja, den ser bedre ut.

kmokstad:
Merged manually the alternative solution from kmokstad:geom_centered instead

PR #69, opened 2016-10-18
Fix two LR issues

akva2:
- fix order of solution vector registry in SIMSolverAdap
- use a sparse system for the interpolation problem in ASMu2D. problem is sparse and large.

### DIFF
--- a/src/ASM/LR/ASMu2Drecovery.C
+++ b/src/ASM/LR/ASMu2Drecovery.C
@@ -414,19 +414,23 @@ LR::LRSplineSurface* ASMu2D::regularInterpolation (const RealArray& upar,
     return nullptr;
   }
 
-  DenseMatrix    A(nBasis,nBasis);
-  Matrix         B(points,true);
+  SparseMatrix   A(SparseMatrix::SUPERLU);
+  A.resize(nBasis, nBasis);
+  Matrix B2(points,true); // transpose to get one vector per field
+  StdVector B(B2);
   Go::BasisPtsSf splineValues;
 
   // Evaluate all basis functions at all points, stored in the A-matrix
   // (same row = same evaluation point)
   for (size_t i = 0; i < nBasis; i++)
   {
-    lrspline->computeBasis(upar[i],vpar[i],splineValues);
-    // optimization note:
-    // without an element id, splineValues will be stored as a full dense vector
-    for (size_t j = 0; j < nBasis; j++)
-      A(i+1,j+1) = splineValues.basisValues[j];
+    int iel = lrspline->getElementContaining(upar[i], vpar[i]);
+    lrspline->computeBasis(upar[i],vpar[i],splineValues, iel);
+    size_t k = 0;
+    for (auto& function : lrspline->getElement(iel)->support()) {
+      int j = function->getId();
+      A(i+1,j+1) = splineValues.basisValues[k++];
+    }
   }
 
   // Solve for all solution components - one right-hand-side for each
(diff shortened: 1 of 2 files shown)


src/ASM/LR/ASMu2Drecovery.C
```diff
@@ -414,8 +414,10 @@ LR::LRSplineSurface* ASMu2D::regularInterpolation (const RealArray& upar,
     return nullptr;
   }
 
-  DenseMatrix    A(nBasis,nBasis);
-  Matrix         B(points,true);
+  SparseMatrix   A(SparseMatrix::SUPERLU);
+  A.resize(nBasis, nBasis);
+  Matrix B2(points,true); // transpose to get one vector per field
+  StdVector B(B2.ptr(), B2.size());
```
kmokstad: Eller bare StdVector B(B2);
fordi StdVector har en std::vector<double> constructor og Matrix kan castes til utl::vector (som igjen er en std::vector)

src/ASM/LR/ASMu2Drecovery.C
```diff
@@ -426,7 +428,8 @@ LR::LRSplineSurface* ASMu2D::regularInterpolation (const RealArray& upar,
     // optimization note:
     // without an element id, splineValues will be stored as a full dense vector
     for (size_t j = 0; j < nBasis; j++)
-      A(i+1,j+1) = splineValues.basisValues[j];
+      if (utl::trunc(splineValues.basisValues[j]) != 0.0)
```
kmokstad: utl::trunc var vel i utganspunktet kun ment for bruk ifm. (debug) utskrift av flyttalls-størrelser, som skriver ut 0.0 hvis lavere enn en gitt toleranse (som kan settes via en variabel). Å bruke den slik for numerisk oppførsel er litt skummelt, hvis denne toleransen er satt noe slakk. Da vil resultatet kunne blir påvirka utilsektet. Bør vel i dette tilfellet heller finne en mer robust løsning for å avgjøre hvilke basis funksjoner som er logisk lik 0.0

VikingScientist: En bedre løsning her er å spesifisere hvilket element `iel` man er i, og da få ut en sparse vektor med basisfunksjonsverdier:

``` c++
int iel = lrspline->getElementContaining(upar[i],vpar[i]);
lrspline->computeBasis(upar[i],vpar[i],splineValues, iel);
int k = 0;
for(auto function : lrspline->getElement(iel)->support() ) {
  int j = function->getId();
  A(i+1, j+1) = splineValues.basisValues[k++];
}
```
PS: jeg har ikke testet koden over, så ikke klipp-lim ukritisk

VikingScientist:
Looks good. Change the function evaluation to use the sparse format and then we're ready to pull.

Note that the regular interpolation on LR-spline objects may create singular matrices A if one is not very careful in picking interpolation points. In particular: the Greville points are not a safe choice and may create singular A. This was the case long before this pull request, so nothing that will influence the merging, but I include it here as a heads up if anyone start using this function on a large scale. Might consider including this bit of information somewhere in the code. Either throw an error from the function when this happens or just put a warning in the documentation. If one is using the greville points as input, then an easy check would be to see if any input points coincide, i.e. there exist some pair (i,j) for which `upar[i]==upar[j]` and `vpar[i]==vpar[j]`.

src/ASM/LR/ASMu2Drecovery.C
akva2: -- original --
ASMu2D::projectSolutio|    542.10 77443.39 |    185.94 26562.60 |     7

Peak memory usage: 3.2GB

-- new, mine --
ASMu2D::integrate(I)  |     38.90  1852.45 |     37.72  1796.08 |    2

Peak memory usage: 220MB

-- new, kjetil --
ASMu2D::projectSolutio|      2.49   355.38 |      2.49   355.52 |     7

Peak memory usage: 220MB

-- new, kjetil, old liblr --
ASMu2D::projectSolutio|    152.47 21780.98 |    152.61 21802.05 |     7
Peak memory usage: 220MB

last result just to  show why i didn't do it like this prior to the optimization.

akva2:
updated. note that all of the (apparent) singular cases now go away in my tests (Lshape and interiorlayer with poisson) due to the sparse LU. round-off in dense lu for large systems has been biting us more often than necessary, masquerading as the greville point issue to non-investigative eyes.

src/ASM/LR/ASMu2Drecovery.C
akva2: overlooked this. changed.

PR #73, opened 2016-11-11
Fix LR dirichlet boundary conditions

akva2:
Nodes in edgeCurve were sorted, but nodes in getEdgeNodes were not.

PR #78, opened 2016-11-15
Fix mixed LR

akva2:
Some fixes for the mixed LR ASM module. I can now run 'mixed' Stokes fine through LR. More stuff will be amended to fix compatible stokes + LR, so hold on the merge (but feel free to review now)!

@kmokstad please have a look at the printing commit.

### DIFF
--- a/src/ASM/LR/ASMu2Dmx.C
+++ b/src/ASM/LR/ASMu2Dmx.C
@@ -445,6 +445,9 @@ bool ASMu2Dmx::integrate (Integrand& integrand, int lIndex,
       return false;
     }
 
+    if (integrand.getIntegrandType() & Integrand::ELEMENT_CORNERS)
+      this->getElementCorners(iel,fe.XC);
+
     // Get integration gauss points over this element
     this->getGaussPointParameters(gpar[t2-1],t2-1,nGP,geoEl,xg);
 
--- a/src/ASM/LR/ASMu2Dmx.C
+++ b/src/ASM/LR/ASMu2Dmx.C
@@ -96,11 +96,8 @@ void ASMu2Dmx::clear (bool retainGeometry)
 
 size_t ASMu2Dmx::getNoNodes (int basis) const
 {
-  if (basis > (int)nb.size() || basis < 0)
-    basis = 0;
-
-  if (basis == 0)
-    return std::accumulate(nb.begin(), nb.end(), 0);
+  if (basis > (int)nb.size() || basis < 1)
+    return this->ASMbase::getNoNodes(basis);
 
   return nb[basis-1];
 }
--- a/src/ASM/LR/ASMu2Dmx.C
+++ b/src/ASM/LR/ASMu2Dmx.C
@@ -237,6 +237,8 @@ bool ASMu2Dmx::generateFEMTopology ()
   std::cout <<"NEL = "<< nel <<" NNOD = "<< nnod << std::endl;
 #endif
 
+  geo = m_basis[geoBasis-1].get();
+
   return true;
 }
 
--- a/src/ASM/LR/ASMu2Dmx.C
+++ b/src/ASM/LR/ASMu2Dmx.C
@@ -522,13 +522,23 @@ bool ASMu2Dmx::evalSolution (Matrix& sField, const Vector& locSol,
 
   std::vector<Go::BasisPtsSf> splinex(m_basis.size());
 
+  std::vector<size_t> nc(nfx.size());
+  std::copy(nfx.begin(), nfx.end(), nc.begin());
+  // assume first basis only
+  if (locSol.size() < std::inner_product(nb.begin(), nb.end(), nfx.begin(), 0u)) {
+    std::fill(nc.begin(), nc.end(), 0);
+    nc[0] = nfx[0];
+  }
+
   // Evaluate the primary solution field at each point
   sField.resize(std::accumulate(nfx.begin(), nfx.end(), 0), nPoints);
   for (size_t i = 0; i < nPoints; i++)
   {
     size_t ofs=0;
     Vector Ztmp;
-    for (size_t j=0; j < m_basis.size(); ++j) {
+    for (size_t j=0; j <  m_basis.size(); ++j) {
+      if (nc[j] == 0)
+        continue;
       // Fetch element containing evaluation point.
       // Sadly, points are not always ordered in the same way as the elements.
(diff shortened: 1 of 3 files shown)


akva2:
there we go. compatible was easier than anticipated. runs fine as well now.

src/ASM/LR/ASMu2Dmx.C
```diff
@@ -668,3 +668,22 @@ bool ASMu2Dmx::refine (const LR::RefineData& prm,
...
+  size_t nbb = 0;
+  while (nbb + nb[b] < inod && b < nb.size())
+    nbb += nb[b++];
+  ++b;
+
+  const LR::Basisfunction* basis = this->getBasis(b)->getBasisfunction(inod-nbb-1);
+  if (!basis) {
+    std::cerr << "Asked to get coordinate for node " << inod
+              << ", but only have " << this->getBasis(b)->nBasisFunctions() << std::endl;
+    return Vec3();
```
kmokstad: Somewhat unclear error message, consider adding at the end: " nodes in basis "<< b+1

src/ASM/LR/ASMu2Dmx.C
```diff
@@ -668,3 +668,22 @@ bool ASMu2Dmx::refine (const LR::RefineData& prm,
...
+  }
+  return Vec3(&(*basis->cp()),nsd);
+
+}
```
kmokstad: Remove blank line 688

src/ASM/LR/ASMu2Dmx.C
akva2: e75dd4a

src/ASM/FiniteElement.C
```diff
@@ -42,7 +48,7 @@ std::ostream& FiniteElement::write (std::ostream& os) const
 std::ostream& MxFiniteElement::write (std::ostream& os) const
 {
   this->FiniteElement::write(os);
-  for (size_t b = 0; b+1 < Nx.size(); b++)
+  for (size_t b = 0; b < Nx.size(); b++)
   {
```
kmokstad: Not sure, but to me it looks like it should have been "for (size_t b = 1; b < Nx.size(); b++)"
Or else the first basis is printed twice, due to calling the parent class method?

src/ASM/FiniteElement.h
```diff
@@ -114,6 +114,8 @@ class MxFiniteElement : public FiniteElement
 protected:
   //! \brief Writes the finite element object to the given output stream.
   virtual std::ostream& write(std::ostream& os) const;
+  //! \brief Global Output stream operator.
+  friend std::ostream& operator<<(std::ostream& os, const MxFiniteElement& fe);
```
kmokstad: Do you really need to overload this operator? Since it only invokes FiniteElement::write which is virtual, the correct behaviour will be achieved without it when invoked with a MxFiniteElement object as argument value.

src/ASM/FiniteElement.C
```diff
@@ -20,6 +20,12 @@ std::ostream& operator<< (std::ostream& os, const FiniteElement& fe)
 }
 
 
+std::ostream& operator<< (std::ostream& os, const MxFiniteElement& fe)
+{
+  return fe.write(os);
+}
+
+
```
kmokstad: See comment below on the header file

src/ASM/FiniteElement.C
akva2: first basis is not in Nx. Nx is only the additional bases (see MxFiniteElement constructor). it's this way due to inheritance stuff (we want to look like a FiniteElement for the first basis).

src/ASM/FiniteElement.h
akva2: yes, i need the implementation. it prints as a FiniteElement otherwise. just try for yourself.

VikingScientist:
Do we have a basic mixed LR test to accompany this PR?

src/ASM/FiniteElement.C
kmokstad: OK, just disregard my previous comment, your suggested fix is correct. (I did not see the rest of the body, which you need to do in order to see that)

src/ASM/FiniteElement.h
akva2: Ja jeg var like overrasket selv, men det var det debuggeren fortalte meg.

akva2:
rebased as the printing was handled separately. added some fixes, we can now run adaptive stokes with taylor-hood, div-compatible and, with some other fixes that will come separately, stabilized formulations.

src/ASM/FiniteElement.h
kmokstad: The conclusion on this was that it is _not_ needed, @akva2 probably triggered a bug in the compiler, or something. I just pushed the one-liner fix (line 45) plus a test-program, such that this commit can now be removed from this PR.

akva2:
as for the tests, i will setup tests, but i cannot yet as this pr depends on https://github.com/OPM/IFEM/pull/73

src/ASM/LR/ASMu2Dmx.C
```diff
@@ -633,14 +646,22 @@ bool ASMu2Dmx::refine (const LR::RefineData& prm,
...
-                                        line->multiplicity_);
-      else
-        m_basis[j]->insert_const_v_edge(line->const_par_,
-                                        line->start_, line->stop_,
-                                        line->multiplicity_);
+      else {
+        int mult = 1;
+        if (ASMmxBase::Type == ASMmxBase::REDUCED_CONT_RAISE_BASIS1 ||
+            ASMmxBase::Type == ASMmxBase::REDUCED_CONT_RAISE_BASIS2) {
+          int p = m_basis[bas]->order(line->span_u_line_ ? 0 : 1)-1;
```
VikingScientist: Actually, if the line is spanning the u-direction, then the reduced continuity is *across* it, i.e. in the v-direction. You will have to switch the polynomial degree you are looking up here.

src/ASM/LR/ASMu2Dmx.C
```diff
@@ -633,14 +646,22 @@ bool ASMu2Dmx::refine (const LR::RefineData& prm,
...
+          int k = p - line->multiplicity_;
+          int q = m_basis[j]->order(line->span_u_line_ ? 0 : 1)-1;
```
VikingScientist: The same here

PR #81, opened 2016-11-18
Cosmetics: proper whitespace use in src/ASM/LR

VikingScientist:
Please don't make this PR linger because it *will* create conflicts if you make alterations to any of the LR-files without it.


akva2:
thanks!

PR #89, opened 2016-12-02
added: use BLAS in outer_product

akva2:
Not a lot BLAS can do better, but it can possibly achieve slightly better cache utilization through smarter traversals.


akva2:

PR #90, opened 2016-12-06
fixed: do not stick null ptrs in function map

akva2:
parsing may fail


kmokstad:
Merged manually, with some modifications (RealFunc* instead of auto)

PR #92, opened 2016-12-09
New class SIMmodel

kmokstad:
A new class SIMmodel is introduced containing those parts of SIMbase that dealt with model input and generation, including initial condition handling. This will require some minor modifications of those apps using initial condtions + some regression test updates due to change in console output.

Have so far only tested it for Elasticity and FractureElasticity, but please have a look.

PR #93, opened 2016-12-12
Skew-symmetric convection

timovanopstal:
add skew-symmetric form of convection to weak operators.


akva2:
this part is fine.

timovanopstal:
since @akva2 review: fixed bug in conservative convection (reflected in NavierStokes reg test) and decided to avoid using Advection for convective form as well: for consistent notation -- do you agree?

akva2:
sure, not a problem. i'd inline it when moving to blas in any case.

timovanopstal:
status: ready to rebase/test/merge after #92 and #94

timovanopstal:
jenkins build this with downstreams ifem-stokes=40 ifem-navierstokes=38 please

PR #98, opened 2016-12-21
Added: protected method SIMinput::parseTopologySet

kmokstad:
Used by the <refine> and <raisorder> tags. It also allows specifying a topology set defining the patch(es) to refine.

This is a redo of #71 but (in my opinion) in a more elegant way.
The functionality should be exactly the same.


akva2:
looks good.

PR #83, opened 2016-11-22
Subdivision (in the sense of domain decomposition)

timovanopstal:
Parallelization in IFEM is currently patch-based, this PR decouples parallelization from patch structure. In this way geometry generation is not driven by parallelization considerations.

PR #102, opened 2017-01-19
Subdivision for mixed bases

timovanopstal:
Extend the merged #83 to mixed bases.

Changes are more extensive than anticipated: I'll probably have to work in ASMmxBase::establishBases() too.


src/ASM/ASMs2D.h
```diff
@@ -115,6 +115,8 @@ class ASMs2D : public ASMstruct, public ASM2D
   //! \brief Returns the spline curve representing a boundary of this patch.
   //! \param[in] dir Parameter direction defining which boundary to return
   virtual Go::SplineCurve* getBoundary(int dir, int = 1);
+  //! \brief Returns the number of bases.
+  virtual size_t getNoBasis() const { return 1; }
```
akva2: nei. denne sitter i ASMbase.

src/ASM/ASMs2D.C
```diff
@@ -632,6 +632,19 @@ bool ASMs2D::connectBasis (int edge, ASMs2D& neighbor, int nedge, bool revers,
   for (int& it : masterNodes)
     it += master;
 
+  int n1, n2;
+  if (!neighbor.getSize(n1,n2,basis)) return false;
+  std::cout << "\tmaster coords:";
+  for (int i=1; i<=n1; ++i)
```
akva2: mest en fyi. #include "Vec3Oper.h" ogo du kan printe en vec3.

src/ASM/ASMbase.h
```diff
@@ -165,6 +165,8 @@ class ASMbase
   virtual unsigned char getNoFields(int b = 0) const { return b > 1 ? 0 : nf; }
   //! \brief Returns the number of Lagrange multipliers per node.
   unsigned char getNoLagPerNode() const { return nLag; }
+  //! \brief Returns the number of bases.
+  virtual size_t getNoBasis() const { return 1; }
```
akva2: we do not need another one! https://github.com/OPM/IFEM/blob/master/src/ASM/ASMbase.h#L235

PR #109, opened 2017-01-24
Added: Optional second argument to the solveStep method

kmokstad:
telling which one of the two sub-step solvers to invoke first.

Not sure if there is a better way of doing this. But the clue is that I want the option to solve S2 before S1 without swapping the template argument (because VTF export, etc. is connected to S1).


akva2:

PR #110, opened 2017-01-26
added: check multiplicities of end knots on edges

akva2:
we do not want direct evaluation unless they are interpolatory.
this is not always the case in subdivision models.

this needs equivalent treatment in 3D, but i have put that off until we have decided the color of the shed.


akva2:

PR #111, opened 2017-01-27
Some small fixes

akva2:
- Fix sanity check/ hack fix in parallel domain decomposition
- Fix 3D mixed ASM for uneven amount of elements in each direction


akva2:

PR #115, opened 2017-02-10
Added: Force scalar parsing of neumann condition if direction > nsd

TheBB:
Latest in the silly hacks series. Without this it's impossible to get scalar Neumann conditions on sims with more than one field. Useful e.g. in poroel for separate traction and flux conditions.

### DIFF
--- a/src/SIM/SIMinput.C
+++ b/src/SIM/SIMinput.C
@@ -982,7 +982,7 @@ bool SIMinput::setTracProperty (int code, Property::Type ptype,
 bool SIMinput::setNeumann (const std::string& prop, const std::string& type,
                            int direction, int code)
 {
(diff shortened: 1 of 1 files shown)


TheBB:
Maybe it's less dirty to have an xml attribute called parse_as or some such.

kmokstad:
VIl tro det er greit, direction > nsd brukes vel ikke til andre formål i neumann sammenheng.
Men ville kosta på meg en ekstra parantes for å gjøre logikken mer entydig, dvs.
`if ((direction == 0 && this->getNoFields() == 1) || direction > nsd)`

akva2:
please add paranthesis, then we can merge.

TheBB:
Updated!

akva2:
jenkins build this with downstream please

PR #116, opened 2017-02-14
add: manually change precision of pointdata

timovanopstal:


timovanopstal:

timovanopstal:

src/SIM/SIMoutput.C
```diff
@@ -79,6 +79,10 @@ bool SIMoutput::parseOutputTag (const TiXmlElement* elem)
   if (strcasecmp(elem->Value(),"resultpoints"))
     return this->SIMinput::parseOutputTag(elem);
 
+  int precision;
+  if (utl::getAttribute(elem,"precision",precision))
+    this->setResultPrecision(precision);
+
```
kmokstad: Why the temporary and extra method, since myResPrec is owned by this class you could just do:
`utl::getAttribute(elem,"precision",myResPrec);`

timovanopstal: makes sense, will change

PR #118, opened 2017-02-15
Quell warnings with clang

akva2:
With these fixes, clang now compiles IFEM-core and CFD applications warnings-less except a few issues which are taken care of in https://github.com/OPM/IFEM/pull/102

Replaces #117

### DIFF
--- a/src/ASM/ASMs2D.h
+++ b/src/ASM/ASMs2D.h
@@ -529,6 +529,7 @@ protected:
   //! \param[out] XC Coordinates of the element corners
   void getElementCorners(int i1, int i2, std::vector<Vec3>& XC) const;
 
+  using ASMbase::generateThreadGroups;
   //! \brief Generates element groups for multi-threading of interior integrals.
   //! \param[in] integrand Object with problem-specific data and methods
   //! \param[in] silence If \e true, suppress threading group outprint
--- a/Apps/Common/SIMSemi3D.h
+++ b/Apps/Common/SIMSemi3D.h
@@ -256,6 +256,7 @@ public:
     return true;
   }
 
+  using SIMadmin::parse;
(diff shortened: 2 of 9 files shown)


akva2:

timovanopstal:
also performing a local check just to be sure ;-) but it looks good

src/ASM/ASMs2D.h
```diff
@@ -529,6 +529,7 @@ class ASMs2D : public ASMstruct, public ASM2D
   //! \param[out] XC Coordinates of the element corners
   void getElementCorners(int i1, int i2, std::vector<Vec3>& XC) const;
 
+  using ASMbase::generateThreadGroups;
   //! \brief Generates element groups for multi-threading of interior integrals.
```
kmokstad: why is it only ASMs2D that require this using-statement? The other 2D classes (mx and Lag) also reimplements only one of the generateThreadGroups methods, so if this one gives warning and not the other two it does not make sense in my head.

akva2: ASMbase expose https://github.com/OPM/IFEM/blob/master/src/ASM/ASMbase.h#L375 and https://github.com/OPM/IFEM/blob/master/src/ASM/ASMbase.h#L377

ASMs2D only implements https://github.com/OPM/IFEM/blob/master/src/ASM/ASMs2D.h#L536

this hides https://github.com/OPM/IFEM/blob/master/src/ASM/ASMbase.h#L377 from its children, which is what the warning is all about.

there is no warning in the children, because they only see https://github.com/OPM/IFEM/blob/master/src/ASM/ASMs2D.h#L536 and they thus implement all overloads inherited from their children.

timovanopstal:
merged manually

PR #119, opened 2017-02-15
Added: Print analytic solution at the result points, when available

kmokstad:
Added: Utility method SIMbase::getReactionForces.
Use this in the apps instead of myEqSys->getReactions().


akva2:

akva2:
since this breaks no downstreams (i guess some elasticity apps can be adjust to use the new helper though), i'll merge.

PR #112, opened 2017-02-02
Added: Assembly of scalar quantities together with the equation system.

kmokstad:
Added: Debug print of extracted element solution vectors.
Changed: Default value of argument witRF in SIMbase::initSystem to false.
Changed: Removed three spaces in the summary print for scalar solutions.

No merge yet, companion to the staggered branch for OpenFrac

PR #120, opened 2017-02-15
Triangle patches

kmokstad:
This adds support for 2D patches with triangular Lagrange elements (currently linears only, but extendable).

Includes #112 also, so merge that one first.


akva2:
This works well for dirichlet problems, very nice. However, boundary integrals are currently broken.

kmokstad:
Which was as expected, as I did not test that part yet. Will look at it in between. Just leave this on open

kmokstad:
jenkins build this with downstreams ifem-elasticity=13 please

PR #121, opened 2017-02-16
functionality in utl:matrix3d

timovanopstal:
I have a temporary hack I'm using, but distilled this matrix3d functionality from it that might be useful to pull into master.

### DIFF
--- a/src/LinAlg/Test/TestMatrix.C
+++ b/src/LinAlg/Test/TestMatrix.C
@@ -14,8 +14,8 @@
 
 #include "gtest/gtest.h"
 #include <numeric>
-
-#include <numeric>
+#include <iomanip>
+#include <fstream>
 
 
 TEST(TestMatrix, AddBlock)
@@ -60,4 +60,22 @@ TEST(TestMatrix, AddRows)
   for (size_t j = 1; j <= 5; j++, fasit++)
     for (size_t i = 1; i <= 2; i++, fasit++)
       ASSERT_EQ(a(i,j), fasit);
+}
+
+
+TEST(TestMatrix3D, DumpRead)
+{
+  size_t n0(2), n1(3), n2(4);
+  utl::matrix3d<double> A(n0, n1, n2);
+  Real* p = A.ptr();
+  for (size_t i = 0; i<A.size(); ++i, ++p)
+    *p = 3.14159*i*i;
+
+  std::string fname = std::tmpnam(nullptr);
+  std::ofstream os(fname);
+  os << std::setprecision(16) << A;
+  std::ifstream is(fname, std::ios::in);
+  utl::matrix3d<double> B(is);
+  B -= A;
+  ASSERT_NEAR(B.norm2(), 0.0, 1.0e-13);
 }
--- a/src/LinAlg/matrix.h
+++ b/src/LinAlg/matrix.h
@@ -753,6 +753,32 @@ namespace utl //! General utility classes and functions.
       n[2] = n_3;
     }
 
+    //! \brief Constructor to read a matrix from a stream.
+    matrix3d(std::istream& is, std::streamsize max = 10)
+    {
+      // Read size
+      is.ignore(10, ':');
+      is.ignore(1);
+      for (size_t i=0; i<3; ++i)
+        is >> n[i];
+      is.ignore(1, '\n');
+
+      // Read contents
+      double num;
+      elem.resize(n[0]*n[1]*n[2]);
+      for (size_t k=0; k<n[2]; ++k) {
+        is.ignore(max, '\n');
+        for (size_t i=0; i<n[0]; ++i) {
+          is.ignore(max, ':');
+          for (size_t j=0; j<n[1]; ++j) {
+            is >> num;
+            elem[n[0]*n[1]*k + n[0]*j + i] = num;
+          }
+          is.ignore(1, '\n');
+        }
+      }
+    }
+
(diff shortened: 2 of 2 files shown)


src/LinAlg/Test/TestMatrix.C
```diff
@@ -61,3 +61,19 @@ TEST(TestMatrix, AddRows)
...
+
+TEST(TestMatrix3D, DumpRead)
+{
+  size_t n0(2), n1(3), n2(4);
+  utl::matrix3d<double> A(n0, n1, n2);
+  Real* p = A.ptr();
+  for (size_t i = 0; i<A.size(); ++i, ++p)
+    *p = 3.14159*i*i;
+
+  std::string fname = "temp.txt";
```
akva2: std::tmpnam()

src/LinAlg/matrix.h
```diff
@@ -989,6 +1027,22 @@ namespace utl //! General utility classes and functions.
   }
 
   template<> inline
+  matrix3d<float>& matrix3d<float>::add(const matrix3d<float>& X, float alfa)
```
akva2: also need the non-blas fallback implementations

src/LinAlg/matrix.h
```diff
@@ -753,6 +755,32 @@ namespace utl //! General utility classes and functions.
       n[2] = n_3;
     }
 
+    //! \brief Constructor fill matrix from istream.
+    matrix3d(std::string fname)
```
akva2: pass by const ref.

src/LinAlg/matrix.h
```diff
@@ -753,6 +755,32 @@ namespace utl //! General utility classes and functions.
...
+      double num;
+      elem.clear();
+      while (is >> num)
+        elem.push_back(num);
+      if (n[0]*n[1]*n[2] != elem.size())
+        std::cerr << "matrix3d: error reading from file, expected size "
+                  << n[0]*n[1]*n[2] << " but got " << elem.size() << std::endl;
+    }
+    //! \brief Dump contents to output stream.
+    void dump(std::string fname, size_t precision = 16)
```
akva2: pass by const ref.

timovanopstal:
thanks @akva2! amended the commit.

src/LinAlg/matrix.h
```diff
@@ -753,6 +755,32 @@ namespace utl //! General utility classes and functions.
...
+      std::ofstream os(fname);
+      os << std::setprecision(precision);
+      os << n[0] << std::endl << n[1] << std::endl << n[2] << std::endl;
+      for (size_t k=1; k<=n[2]; ++k)
+        for (size_t j=1; j<=n[1]; ++j)
+          for (size_t i=1; i<=n[0]; ++i)
+            os << this->operator()(i,j,k) << "\n";
+    }
+
     //! \brief Resize the matrix to dimension \f$n_1 \times n_2 \times n_3\f$.
```
kmokstad: Well, I would rather not have file operations inside low-level classes like this. Instead the constructor and dump methods above should have an std::istream& and std::ostream& argument instead, such that it could be decided outside if these are from/to files or other streams. Or even better, implement as operator<< and operator>> instead (like for the vector and matrix classes in this file).

Moreover, std::endl is the same as "\n" followed by a flush, so in line 777 it would be sufficient with three "\n" instead. No point in flushing except for at the end of the print.

timovanopstal: I understand your POV, I decided against supplying a stream so that I could call dump in one line of code. The << operator has already been implemented so an alternative would be to implement the corresponding >> instead.

timovanopstal:
This should incorporate all the comments (non-blas fallback, and passing streams instead of function names). One remark: I ended up printing the shape of the 3D tensor inside `operator<<` to avoid looping through the potentially big file twice.

timovanopstal:

src/LinAlg/matrix.h
```diff
@@ -1678,6 +1742,7 @@ namespace utl //! General utility classes and functions.
     if (A.empty())
       return s <<" (empty)"<< std::endl;
 
+    s <<"Shape: "<< A.dim(1) <<" "<< A.dim(2) <<" "<< A.dim(3);
```
kmokstad: Not better with "Dimension:" instead?

src/LinAlg/matrix.h
```diff
@@ -21,6 +21,8 @@
 
 #include <vector>
 #include <iostream>
+#include <iomanip>
+#include <fstream>
```
timovanopstal: agreed.

src/LinAlg/matrix.h
```diff
@@ -753,6 +755,32 @@ namespace utl //! General utility classes and functions.
       n[2] = n_3;
     }
 
+    //! \brief Constructor fill matrix from istream.
+    matrix3d(std::ifstream& is, std::streamsize max = 10)
```
timovanopstal: agreed.

src/LinAlg/matrix.h
timovanopstal: Works for me, changed.

timovanopstal:
shameless self-merge

timovanopstal:
hmm... I was absolutely sure I pressed "rebase and merge"

kmokstad:
Seems not, (urgh). Maybe the administrator can rebase if we ask nice and before branching off to far?
(I prefer to do the merging and pushing from console, I don't trust these GUI solutions too much..)

PR #124, opened 2017-03-07
changed: allow specifying number of components in vector field

akva2:
useful when writing vector fields derived from a scalar equation.

noticed while wanting to output the derived velocities in the darcy application as a vector field.


VikingScientist:

Issue #122, opened 2017-03-07
Feature request: discretization type=triangles


VikingScientist:
See last comment in PR #120

kmokstad:
PR #125 is now merged, so this is no longer any issue

PR #125, opened 2017-03-07
Improved support for triangles

kmokstad:
On request, here (the second commit) is a solution to #122.
The first commit is just a cosmetic thing (I think).


akva2:

PR #129, opened 2017-03-10
Added: Option to store also the diverged solution on VTF before exiting

kmokstad:
Useful when debugging sensitive models.


akva2:
fix commit message please (diverGed)

PR #131, opened 2017-03-13
Some small fixes in 3D LR ASM

akva2:
- avoid copying a matrix
- fix global integration point counter


akva2:

Issue #133, opened 2017-03-15
The README.md file lacks information on how to install HDF5 support


kmokstad:
Maybe as an optional package? Since the core works without it.

Now I (and other amateurs) have to guess that libhdf5-dev is the one you need. But when used with PETSc maybe another package is required? And wasn't it an issue also with hdf5 and petsc toghether that made it difficult before? Or maybe that was only with the older Linux.

Anyway, it would be nice if the README told what to do in both cases (with PETSc and without). I have not installed PETSC yet, so it looks like libhdf5-dev does it.

VikingScientist:
I guess this should be closed now? Neat trick for those that are not aware: If your commit message includes the text "fixes #X", (X is the issue track number) [or some variation of the keyword "fixes"](https://help.github.com/articles/closing-issues-via-commit-messages/), then github will automagically close the issue once that particular commit is merged into trunk.

kmokstad:
Yes. Good to know :-)

PR #135, opened 2017-03-16
Updated README.md on the dependent modules

kmokstad:
Added the commands to install HDF5 as of Issue #133.

Also removed a lot of other packages which are not needed to build the IFEM simulation module.
I guess they were used only when the modeling tools were based on GoTools as well? But now I understand the new Python module is GoTools free.

But please correct me if wrong...


akva2:
yeah, those were only needed to build the old geomodeler.

PR #136, opened 2017-03-20
fixed: extract the entire patch vector for fields

akva2:
if number of components differ, the fields classes now
takes care of extracting the required components from the
full patch vector.


akva2:
jenkins build this with downstreams ifem-advectiondiffusion=6 ifem-stokes=59 ifem-navierstokes=55 ifem-fsi=15 please

PR #141, opened 2017-04-22
Fixes for correct handling of nodal point loads in multi-patch models

kmokstad:
The first commit is more or less cosmetic, except for the propagation of the checkRHS flag.
The other two are changes needed to handle nodal loads correctly in multi-patch models.
External nodal loads are used in Elasticty apps only, so others should not be affected.


kmokstad:
jenkins build this with downstreams ifem-elasticity=20 please

kmokstad:
jenkins build this with downstreams ifem-elasticity=20 please

akva2:
looks fine.

PR #143, opened 2017-04-28
Preparation for coupled Boussinesq

akva2:
This adds some new weak/residual ops used in the coupled Boussinesq implementation.


akva2:
shameless self-merge. i'm the sole consumer of this code and never got any feedback on it earlier.

PR #147, opened 2017-05-05
Output refactor

kmokstad:
After the very nice restart refactor, I think this follows naturally. It will make sense to let SIMSolver also handle the result output as well as restart such that you don't need the static method in AppCommon.

Downstream changes are modest, and ready, and will be pushed to respective branches after general acceptance of this one. But maybe @akva2 have something to say on it first.

### DIFF
--- a/Apps/Common/SIMSolver.h
+++ b/Apps/Common/SIMSolver.h
@@ -72,11 +72,6 @@ public:
 
   //! \brief Advances the time step one step forward.
   bool advanceStep() { return tp.increment() && S1.advanceStep(tp); }
-  //! \brief Advances the time step \a n steps forward.
-  void fastForward(int n) { for (int i = 0; i < n; i++) this->advanceStep(); }
-
-  //! \brief Postprocesses the solution of current time step.
-  void postSolve(const TimeStep& t, bool rst = false) { S1.postSolve(t,rst); }
 
   //! \brief Solves the problem up to the final time.
   virtual int solveProblem(char* infile, const char* heading = nullptr,
--- a/Apps/Common/SIMSolver.h
+++ b/Apps/Common/SIMSolver.h
@@ -14,10 +14,11 @@
 #ifndef _SIM_SOLVER_H_
 #define _SIM_SOLVER_H_
 
+#include "IFEM.h"
 #include "SIMadmin.h"
 #include "TimeStep.h"
 #include "HDF5Writer.h"
-#include "IFEM.h"
+#include "XMLWriter.h"
 #include "tinyxml.h"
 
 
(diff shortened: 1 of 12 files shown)


akva2:
i dislike the smashing together of everything into huge monster interface just as much as i have ever done, but i won't protest this time either.

Apps/Common/SIMSolver.h
```diff
@@ -195,12 +196,36 @@ template<class T1> class SIMSolver : public SIMadmin
     return restartStep;
   }
 
+  //! \brief Handles application data output.
```
akva2: obviously this will be visible in downstreams, but who is now responsible for invoking this ? the protected nature of exporter implies the solver, but i do not see it doing so hence the question.

akva2: aha, i see now, it sets the internal pointer but still returns a reference to the world. that doesn't seem right.

kmokstad:
I won't call the SIMSolver interface a 'huge monster', that is to exaggerate beyond anything.
I just felt this was an improvement, because the template function in the AppCommon only was realized over SIMSolver anyway and not relevant for other types, then I don't see why this method can't be a member of SIMSolver instead, to make its use more obvious.

Anyway, I make one downstream PR now (for AdvectionDiffuction) such that you can see how this is used (it will be similar for the others, but I won't publish those if you still dislike this).

Apps/Common/SIMSolver.h
kmokstad: You're right there. No need to return the pointer as it is not used (and shouldn't be) outside. Just a leftover from the SIM:: function. Will take away.

akva2:
The solver class is now doing restart and data output setup. As my templates proved, this can be done complete separate, enforcing logic separation. Now everything is set in rigid structure where the only place to modify the exporter is a single function. I do not like unnecessary structures like this, flexibility is king, ease of use well below on my list of priorities. You asked. I answered. Now we move on.

kmokstad:
I think ease of use should have higher priority than full flexibility, if we want IFEM to be a package that others than just ourselves can use. And to keep restart and solver-output totally separate from the solver itself does not enhance flexibility in my eyes. One can't "restart" other things than a solver anyway, so why keeping them separate? Same with result output. The flexibility to not use restart and/or hdf-output in an application based on SIMSolver is still there, just don't invoke those methods.

akva2:
Your way. Move on.

kmokstad:
jenkins build this with downstreams ifem-advectiondiffusion=11 ifem-darcy=8 ifem-thermoelasticity=2 ifem-poisson=4 ifem-poroelasticity=17 ifem-stokes=66 ifem-navierstokes=65 ifem-fsi=18 ifem-openfrac=23 ifem-beamex=6 please

kmokstad:
jenkins build this with downstreams ifem-advectiondiffusion=11 ifem-darcy=8 ifem-thermoelasticity=2 ifem-poisson=4 ifem-poroelasticity=17 ifem-stokes=66 ifem-navierstokes=65 ifem-fsi=18 ifem-openfrac=23 ifem-beamex=6 please

kmokstad:
jenkins build this with downstreams ifem-advectiondiffusion=11 ifem-darcy=8 ifem-thermoelasticity=2 ifem-poisson=4 ifem-poroelasticity=17 ifem-stokes=66 ifem-navierstokes=65 ifem-fsi=18 ifem-openfrac=23 please

kmokstad:
jenkins build this with downstreams ifem-advectiondiffusion=11 ifem-darcy=8 ifem-thermoelasticity=2 ifem-poisson=4 ifem-poroelasticity=17 ifem-stokes=66 ifem-navierstokes=65 ifem-fsi=18 ifem-openfrac=23 please

PR #149, opened 2017-05-05
fixup wrong signature in SIMSemi3D

akva2:
oh override how i miss you

### DIFF
--- a/Apps/Common/SIMCoupled.h
+++ b/Apps/Common/SIMCoupled.h
@@ -53,7 +53,13 @@ public:
   }
 
   //! \brief Computes the solution for the current time step.
-  virtual bool solveStep(TimeStep& tp, bool firstS1 = true)
+  virtual bool solveStep(TimeStep& tp)
+  {
+    return this->solveStep(tp, true);
+  }
+
+  //! \brief Computes the solution for the current time step.
+  virtual bool solveStep(TimeStep& tp, bool firstS1)
   {
     if (firstS1)
       return S1.solveStep(tp) && S2.solveStep(tp);
--- a/Apps/Common/SIMSolverTS.h
+++ b/Apps/Common/SIMSolverTS.h
@@ -128,11 +128,10 @@ public:
             return 2;
       }
     }
-
-    return 0;
   }
 
 protected:
+  using SIMSolver<T1>::parse;
   //! \brief Parses a data section from an XML element.
   virtual bool parse(const TiXmlElement* elem)
   {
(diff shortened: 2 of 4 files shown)


Apps/Common/SIMSolverTS.h
```diff
@@ -128,11 +128,10 @@ template<class T1> class SIMSolverTS : public SIMSolver<T1>
             return 2;
       }
     }
-
-    return 0;
```
kmokstad: What? Remove the return value for solveProblem??? Wouldn't this only yield a compiler warning, or what do I not see here...

Apps/Common/SIMCoupled.h
```diff
@@ -53,7 +53,13 @@ template<class T1, class T2> class SIMCoupled
...
 
   //! \brief Computes the solution for the current time step.
-  virtual bool solveStep(TimeStep& tp, bool firstS1 = true)
+  virtual bool solveStep(TimeStep& tp)
+  {
+    return this->solveStep(tp, true);
+  }
+
+  //! \brief Computes the solution for the current time step.
+  virtual bool solveStep(TimeStep& tp, bool firstS1)
```
kmokstad: Sorry, but I don't see what was the problem here. The solveStep method is only here and in the sub-class SIMCoupledSI. Why can't they have the same signature? Other classes withe the solveStep method does not belong to this hierarchy, so that shouldn't matter There are many examples in the IFEM library with default values on arguments in virtual methods with no need to make a duplicate like in this case. So please explain better for a novice why this is wrong (so I don't do it again;-)

Apps/Common/SIMSolverTS.h
akva2: Because it is unreachable.

Apps/Common/SIMCoupled.h
akva2: Because solveStep(tp) is ambious, it can be resolved as either a literal solveStep(tp) or solveStep(tp,true).

Apps/Common/SIMSolverTS.h
kmokstad: ok see it now, because the loop never exits or breaks, only returns. Maybe not a good design..

Apps/Common/SIMCoupled.h
kmokstad: Guess you mean ambiguous, og ikke "ambiøs", as I read it first.. This was maybe another thing pointed out by the new compiler you are trying out? It is still not clear though, how can it be ambiguous when there is only one solveStep method in this hierarchy? solveStep(tp) and solveStep(tp,true) can only resolve to the same method. What other options do the compiler point out? That would be interesting to see (and would probably help my understanding of it).

akva2: The point is that solveStep(tp), could either mean the first method or the second method with its default param. When we then add the using statement to quell warnings, we get two overloads in children.

PR #152, opened 2017-05-11
Added: Interface for evaluating function derivatives

kmokstad:
Can be used for evaluating the residual of analytical solutions.
The default implementation returns zero, sub-classes have to implement the derivatives when needed.

The key ingredient is the intermediate template class SpatialFunction, which adds the interface for evaluating the derivatives. Notice also that RealFunc, VecFunc and [S]TensorFunc now are classes and not typedefs.

Only minor impact on a few of the apps (mainly some include-statements), will be added later when/if
this one is  merged. Not sure about that though, as not 100% happy with the design, and there might be compiler issues for all I know. Adding it here anyway such that those who care may have a look.


akva2:
i see no hook for expression functions? other than that it seems to be a straight forward extension.

kmokstad:
The Expression class only have the `Evaluate` method. I see no `Differentiate` method, which would require it to symbolically differentiate the expression the user provided (such a feature would be neat though, like maple, Mathematica, etc.). As it is now, the user would need to provide the expression for the differentiated expressions explicitly. Then the SpatialFunction::[d]deriv methods could be overloaded to evaluate these. That is a possible extension, but I didn't need that (yet), as I only use this for the NavierPlate solution so far (KirchhoffLove integrand).

akva2:
That would be what I had in mind. Some way for user to specify those derivatives. I understood that you had not needed it but it should be added before merge was all

kmokstad:
Sure, will do that

kmokstad:
jenkins build this with downstreams ifem-elasticity=23 ifem-stokes=67 please

src/Utility/ExprFunctions.C
```diff
@@ -186,6 +186,12 @@ EvalFunction::EvalFunction (const char* function)
   // Note, this will also catch things like tan(x), but...
   std::string expr(function);
   IAmConstant = expr.find_first_of('t') > expr.size();
+
+  // Need to do this, since they are fixed-size arrays, not vectors
+  for (auto& it : gradient)
```
akva2: you can use the init list instead, gradient{}, dgradient{} or you can use class initialization, gradient = {}, dgradient = {}.

src/Utility/ExprFunctions.C
```diff
@@ -223,24 +271,142 @@ Real EvalFunction::evaluate (const Vec3& X) const
...
   return result;
 }
 
 
 template<>
+Vec3 VecFuncExpr::deriv (const Vec3& X, int dir) const
+{
+  Vec3 result;
+
+  for (size_t i = 0; i < 3 && i < nsd; ++i)
```
akva2: && p[i] maybe.

src/Utility/ExprFunctions.C
kmokstad: I go for the first one : `gradient{}`

src/Utility/ExprFunctions.C
kmokstad: You mean `for (size_t i = 0; i < 3 && i < nsd && p[i]; ++i)` ?

Not sure that is necessary, because nsd is set to p.size() in the VexFuncExpr constructor (via the default setNoDim()) method, and should only contain non-nulls.

kmokstad:
jenkins build this with downstreams ifem-elasticity=23 ifem-stokes=67 please

PR #154, opened 2017-05-15
Added: An extra matrix3d method + test

kmokstad:
Another convenience method. Using this in the alternative Kirchhoff-Love integrand, where outer_product of columns of d2NdX2 are needed.


akva2:
looks fine.

PR #156, opened 2017-05-16
Fixed: 1D default geometry with nonzero X0

kmokstad:
The single-patch 1D model generator produce incorrect geometry if you specify X0.
This is a proposed fix to that.


akva2:
please update test suite. also there is a test with X0 there so i'm a bit confused.

kmokstad:
There. It was only the case with both X0 and L specified that failed. That case was not covered by the test. But now I added one. Not totally satisfied, as I had to remove some ".0" in the others, but..

akva2:
great, thx

PR #157, opened 2017-05-16
Neumann extension

kmokstad:
This (the last commit) is work in progress. See detailed explanation in that commit message.

Not 100% sure if this is the best way of doing this in a general way, but is the best I came up with now.
The drawback is that the initElementBou signature is changed, requiring all apps that override that method to be updated. So far only supported for 2D non-mixed tensor splines. Will complete for others (and downstream updates) if/when the general concept is accepted.

### DIFF
--- a/src/ASM/ASM1D.C
+++ b/src/ASM/ASM1D.C
@@ -15,6 +15,7 @@
 #include "ASMs1DC1.h"
 #include "ASMs1DLag.h"
 #include "ASMs1DSpec.h"
+#include "Vec3Oper.h"
 
 
 ASMbase* ASM1D::create (ASM::Discretization discretization, unsigned char nf)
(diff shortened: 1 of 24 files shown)


akva2:
Imo it would possibly be better with a separate overload, like mx vs single basis.

akva2:
Or maybe it can be carried by local integral? Do not have the overview so just tossing ideas around.

kmokstad:
IMO --> In my opinion?

Anyway, as the first fixup commit shows, I reconsidered it. Now the initElementBou signature is untouched. Instead, I added a new virtual method for passing the flag to the integrand. This method is then invoked before the integration loop over the patch boundary (so there should be no threading issues, etc.) Only those who really needs this are then bothered with it. Does the job at least.

akva2:
yes, imo = in my opinion. the proposed solution works fine for me.

kmokstad:

kmokstad:

akva2:
s/if each other/of each other in last commit message.

PR #158, opened 2017-05-19
added: L2-projected non-homogenous dirichlet boundary conditions for LR-b splines

akva2:
a bit copy-pasty but so is the 2d code i copy-pasted ;)

### DIFF
--- a/src/ASM/LR/ASMu3D.C
+++ b/src/ASM/LR/ASMu3D.C
@@ -27,6 +27,7 @@
 #include "CoordinateMapping.h"
 #include "GaussQuadrature.h"
 #include "ElementBlock.h"
+#include "MPC.h"
 #include "SplineUtils.h"
 #include "Utilities.h"
 #include "Profiler.h"
@@ -120,6 +121,7 @@ void ASMu3D::clear (bool retainGeometry)
 
   // Erase the FE data
   this->ASMbase::clear(retainGeometry);
+  this->dirich.clear();
 }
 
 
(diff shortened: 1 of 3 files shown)


src/ASM/LR/ASMu3D.C
```diff
@@ -374,40 +375,128 @@ void ASMu3D::closeFaces (int dir, int basis, int master)
...
+  int* c = de.corners;
+  for (const auto& it : corners.find(edge)->second)
+    *c++ = this->getCorner(it[0], it[1], it[2], basis);
+
+  int j = 0;
+  int bcode = abs(code);
+  for (auto b : edgeFunctions)
+  {
+    de.MLGN[j++] = b->getId();
+    int* cid = std::find(de.corners, de.corners+4, b->getId()+ofs);
```
kmokstad: Can't see why you need de.MLGN to be a `map<int,int>` here. Think you can achieve the same with an `IntVec` (as the MLGE member is) and just use `push_back` to build it here. Is the map-functionality used somewhere I did not see it? If not, consider revising. The same also in ASMu2D also btw (probably a copy-paste from there).

src/ASM/LR/ASMu3D.C
```diff
@@ -374,40 +375,128 @@ void ASMu3D::closeFaces (int dir, int basis, int master)
...
+    {
+      de.MLGE[i] = el->getId();
+    }
+    for (auto b : el->support())
+    {
+      de.MNPC[i].push_back(-1);
+      for (auto l2g : de.MLGN)
+        // can't do map::find, since this works on first
+        if (b->getId() == l2g.second)
+          de.MNPC[i].back() = l2g.first;
```
kmokstad: If agree on the above comment, lines 492-495 must be rewritten into
```
for (size_t j = 0; j < de.MLGN.size(); j++)
  if (b->getId() == de.MLGN[j])
    de.MNPC[i].back() = j;
```

src/ASM/LR/ASMu3D.C
```diff
@@ -2029,3 +2118,71 @@ void ASMu3D::generateThreadGroups (const Integrand& integrand, bool silence,
...
+    }
+    if (edgeControlpoints.empty() && edgeControlmatrix.empty())
+    {
+      std::cerr <<" *** ASMu3D::updateDirichlet: Projection failure."
+                << std::endl;
+      return false;
+    }
+
+    // Loop over the nodes of this boundary curve
+    for (nit = dirich[i].MLGN.begin(); nit != dirich[i].MLGN.end(); nit++)
```
kmokstad: And this loop becomes: `j = 0; for (int node : dirich[i].MLGN)`
and substitute `nit->second` by `node` and `nit->first`by `j` below (and increment j properly since can't do that in the range-based loop itself).

Totally up to you if you want to do it or not, since it touches some lines, but in my opinion it increases the readability.

src/ASM/LR/ASMu3D.h
```diff
@@ -16,6 +16,7 @@
 
 #include "ASMunstruct.h"
 #include "ASM3D.h"
+#include "LRSpline/LRSpline.h"
```
kmokstad: Do we actually need to include this header file here? Cannot see it is used here. Isn't it sufficient to forward-declare LRSplineVolume instead?

src/ASM/LR/ASMu3D.h
```diff
@@ -455,6 +502,8 @@ class ASMu3D : public ASMunstruct, public ASM3D
   const std::vector<Matrix>& bezierExtract; //!< Bezier extraction matrices
   std::vector<Matrix>      myBezierExtract; //!< Bezier extraction matrices
 
+  std::vector<DirichletFace> dirich; //<! Inhomogeneous Dirichlet boundary condition data
```
kmokstad: Wrong doxy-syntax

src/ASM/LR/ASMu3D.h
```diff
@@ -355,6 +382,26 @@ class ASMu3D : public ASMunstruct, public ASM3D
...
+  //! \brief Projects inhomogenuous (vector) dirichlet conditions by continuous L2-fit.
+  //! \param[in] edge low-level edge information needed to do integration
+  //! \param[in] values inhomogenuous function which is to be fitted
+  //! \param[out] result fitted value in terms of control-point values
+  //! \param[in] time time used in dynamic problems
+  bool faceL2projection (const DirichletFace& edge,
+                         const VecFunc& values,
+                         Real2DMat& result,
+                         double time) const;
+
```
kmokstad: Consider merging these two almost-identical methods into one, using the `FunctionBase` common interface for the function to project.

src/ASM/LR/ASMu3D.C
```diff
@@ -374,40 +375,128 @@ void ASMu3D::closeFaces (int dir, int basis, int master)
...
+  LR::LRSplineVolume* lr = this->getBasis(basis);
 
-  // enforce the boundary conditions
-  for(LR::Basisfunction* b  : thisEdge)
-    this->prescribe(b->getId()+1,dof,bcode);
+  // get all elements and functions on this edge
+  std::vector<LR::Basisfunction*> edgeFunctions;
+  std::vector<LR::Element*>       edgeElements;
+  lr->getEdgeFunctions(edgeFunctions,edge);
+  lr->getEdgeElements (edgeElements ,edge);
```
akva2: yes. i guess kjetil kept the terminology from the lr surfaces for compatiblity. renamed some of the vars for slightly better clarity.

src/ASM/LR/ASMu3D.C
akva2: agreed. it was inherited from the 2D code. i'll change it here and address 2D afterwards.

src/ASM/LR/ASMu3D.h
akva2: i can do it, but at the cost of writing a custom deleter function for the shared ptr. i think that cost more than it was worth. but if you prefer it i can do it.

src/ASM/LR/ASMu3D.h
akva2: yup. the base class wasn't available when i wrote the code.

src/ASM/LR/ASMu3D.h
akva2: fixed.

src/ASM/LR/ASMu3D.C
```diff
@@ -374,40 +375,127 @@ void ASMu3D::closeFaces (int dir, int basis, int master)
...
+
+  int* c = de.corners;
+  for (const auto& it : corners.find(face)->second)
+    *c++ = this->getCorner(it[0], it[1], it[2], basis);
+
+  int j = 0;
+  int bcode = abs(code);
+  for (auto b : faceFunctions)
+  {
+    de.MLGN[j++] = b->getId();
```
VikingScientist: This crashes when MLGN is an IntVec and not a `std::map<int,int>`

akva2: which is why we use push_back..

akva2:
jenkins build this please

akva2:
jenkins build this please

PR #159, opened 2017-05-20
Changed: Cleaned away some unused/defunct items from AdaptiveSIM

kmokstad:
Suggest this cleanup to reduce potential confusion:
* The initAdaptor and writeGlv methods no longer use the nNormProj argument, which means that the getNoNorms method is not needed either.
* The getNormPrefixes method is not needed. There is no point in passing the projection names as prefixes to the HDF5-exporter, because norms of the projected solutions are not exported to HDF5 as it is implemented now (only for FE solution and analytical, when existing). If we also want to have norms of projected solutions on HDF5, this has to be implemented in a different way (do the solutionNorms call outside the exporter method, and only pass the resulting arrays, like we do for the solutions themselves. This, eventually, has to be a separate PR, though, not that urgent).
* The setupProjections method (which only resizes the proj array) is removed. This task is moved to initAdaptor instead, which means that this method has to be invoked _before_ registering the projection arrays with the DataExporter. Notice that SIMSolverAdap still cannot be used if projections are wanted on the HDF5.export (since there is no hookup for doing this call in the SIMSolver template). initAdaptor is currently invoked by SIMSolverAdap::solveProblem, which is too late for achieving this. This is not a big problem though, as it is only the Elasticity app that exports projected fields for adaptive simulations to HDF5, and this app does not use SIMSolverAdap, yet.
* The other changes are cosmetics

### DIFF
--- a/Apps/Common/SIMSolverAdap.h
+++ b/Apps/Common/SIMSolverAdap.h
@@ -39,18 +39,15 @@ public:
   //! \brief Solves the problem up to the final time.
   virtual int solveProblem(char* infile, const char* heading, bool = false)
   {
-    if (SIMSolver<T1>::exporter)
-      SIMSolver<T1>::exporter->setNormPrefixes(aSim.getNormPrefixes());
-
-    aSim.setupProjections();
-    aSim.initAdaptor(0,2);
+    if (!aSim.initAdaptor())
+      return 1;
 
     this->printHeading(heading);
 
     for (int iStep = 1; aSim.adaptMesh(iStep); iStep++)
       if (!aSim.solveStep(infile,iStep))
         return 1;
-      else if (!aSim.writeGlv(infile,iStep,aSim.getNoNorms()))
+      else if (!aSim.writeGlv(infile,iStep))
         return 2;
       else if (SIMSolver<T1>::exporter)
         SIMSolver<T1>::exporter->dumpTimeLevel(nullptr,true);
--- a/src/SIM/AdaptiveSIM.C
+++ b/src/SIM/AdaptiveSIM.C
@@ -16,7 +16,6 @@
 #include "SIMenums.h"
 #include "ASMunstruct.h"
 #include "IntegrandBase.h"
-#include "SystemMatrix.h"
 #include "Utilities.h"
 #include "IFEM.h"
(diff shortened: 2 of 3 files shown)


kmokstad:
jenkins build this with downstreams ifem-elasticity=25 please

akva2:
Looks fine

VikingScientist:
I see that TRUE_BETA is grouped with the threshold options, but I don't know if that makes sense. I guess that re-using beta altogether is a bad idea. The fact is that we have three main ways of marking refinement, and ideally they should be tagged with different symbols (alpha, beta, gamma?)

**Alpha refinement (Threshold)**
Computes error on all functions or elements followed by
* Refine elements/functions with alpha (0-100) percent error of the MAX error
* Refine elements/functions with more than average error
* Refine elements/functions with alpha(100+) percent error of MIN error

**Beta refinement (default)**
Sorts all functions or elements followed by
* Refine the beta (0-100) percent of elements/functions that have the most error

**Gamma refinement (True_Beta)**
Sorts all functions or elements followed by
* Iteratively refine one-and-one function/element until the function space has grown by beta (0+) percent

Beta takes on wildly different values for different choice of refinement. To get (roughly) the same mesh with the different techniques above one would choose beta=20 (alpha-MAX error), beta=400 (alpha-MIN error), beta=15 (default), beta=100 (true beta). Also: true_beta is the only one which is not predictive, i.e. the refinement is outsourced to the LRSpline-library, while for the first methods, it is handled by IFEM.

VikingScientist:
The above comment was not meant as a critique of this PR, but rather as a critique of (my own) initial structure of the refinement options. 

The PR gets rid of legacy code and improves on the readability of the code. I absolutely loves this kind of commits :heart:

kmokstad:
I interpret the above comments that this commit is acceptable. I merged the old trueBeta flag into the enum variable, to make it more evident that these are not two independent options, but indicate five different ways of using the parameter beta. One could always think of more logic naming, but for now I only did this way to point out the relation to the removed variable. If you want to refactor this in more sensible way, that better fits the intended understanding, that's fine with me,

VikingScientist:
Yes, this commit is acceptable :smile:

PR #161, opened 2017-05-26
L2 refactor

kmokstad:
Includes an option (currently hardcoded) to enforce the end-point values in the L2-projection.
Not very well tested - only in 1D, so this is WIP.

### DIFF
--- a/src/ASM/ASMbase.h
+++ b/src/ASM/ASMbase.h
@@ -627,9 +627,11 @@ public:
   //! \param[out] sField Secondary solution field control point values
   //! \param[in] integrand Object with problem-specific data and methods
   //! \param[in] continuous If \e true, a continuous L2-projection is used
-  virtual bool globalL2projection(Matrix& sField,
-                                  const L2Integrand& integrand,
-                                  bool continuous = false) const;
+  //! \param[in] enforceEnds If \e true, enforce corner point value equality
+  bool globalL2projection(Matrix& sField,
+                          const L2Integrand& integrand,
+                          bool continuous = false,
+                          bool enforceEnds = false) const;
 
   //! \brief Projects the secondary solution using a continuous global L2-fit.
   //! \param[out] sField Secondary solution field control point values
--- a/src/ASM/GlbL2projector.C
+++ b/src/ASM/GlbL2projector.C
@@ -43,7 +43,7 @@ LinSolParams*      GlbL2::SolverParams = nullptr;
      out[1] = {2,2,2,3,3,3,5,5,5}
 */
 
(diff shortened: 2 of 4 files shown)


akva2:
missing: ASMu2Dmxrecovery.C.

worth considering: ASMLRSpline.C, edgeL2projection (and the equivalent faceL2projection in https://github.com/OPM/IFEM/pull/158) and the related duplication of code in ASMxyD implementations of updateDirichlet.

kmokstad:
I know, did not consider the mixed patches yet, until we know that this is actually how we will do it.
ASMbase::globalL2projection cannot be used for mixed, due to the different arrangement of the field variables.

The suggested refactor of the LR-L2-projection is also worth considering, but that is for projecting explicit functions for non-homogeneous boundary conditions, not the secondary solution as here.

akva2:
Yes. It could possibly be unified through fieldfunc, though there need to be substantial code simplicity to be had to go that way as there would be some overhead.

kmokstad:

kmokstad:

src/ASM/LR/ASMu2Dmx.C
```diff
@@ -745,20 +745,19 @@ void ASMu2Dmx::generateThreadGroups (const Integrand& integrand, bool silence,
...
   // TODO: Support for div-compatible
   int p1 = 0;
-  for (size_t i = 1; i <= m_basis.size(); ++i) {
-    if (this->getBasis(i)->order(0) > p1)
-      p1 = this->getBasis(i)->order(0), threadBasis = i;
+  for (size_t i = 1; i <= m_basis.size(); ++i)
+    if (this->getBasis(i)->order(0) > p1) {
+      threadBasis = this->getBasis(i);
+      p1 = threadBasis->order(0);
   }
```
akva2: wrong indent

kmokstad:
Only the last commit here is not merged yet. That one is put on hold as not sure yet if this is needed.

akva2:
it could be used to simplify constrainEdge/constrainFace and clarify the role of ASMbase::updateDirichlet (only handles actual nodal constraints).

akva2:
could you rebase please?

kmokstad:
Most of the refactoring mentioned above was already done long time ago (unifying the globalL2projection methods into a ASMbase method, etc.). Then only thing remaining here is the option to enforce the end/corner point values. Therefore rewriting this branch with that only on top + some small doxy fixes + a potiential bugfix in L2FuncIntegrand::evaluate()  (see second commit).

kmokstad:

PR #162, opened 2017-05-28
Added: Extract total rotation angles for result point print

kmokstad:
For problems with rotational DOFs, we want to print out the absolute rotation angles for the result points (as in FENRIS).


akva2:
tests? :)

kmokstad:
jenkins build this with downstreams ifem-beamex=7 please

PR #164, opened 2017-06-02
Added: Unit test that provokes the mixed/periodicity failure

kmokstad:
After the fixup commit 8b4e1a99cc3d4c8fc33b562c4950b3943351aed8 the only regression test using periodic boundary conditions and mixed dumps core. I think the logic behind the `addPeriodicity()` calls does not account for the existence of several bases. Here is a simple unit test that provokes the same error. This has to be re-thinked a bit, but until then that reg-test should be deactivated to avoid the fuss in the nightly builds.


akva2:
test disabled.

PR #165, opened 2017-06-05
Anasol projection

kmokstad:
This unifies the projection of scalar and vector functions into one set of methods, and this is now used in a general method for projection of the analytical secondary solution for verification purposes.

Also, the result point/line output is enhanced to also handle recovered solution fields.

### DIFF
--- a/src/SIM/SIMbase.C
+++ b/src/SIM/SIMbase.C
@@ -26,8 +26,7 @@
 #include "GlbNorm.h"
 #include "ElmNorm.h"
 #include "AnaSol.h"
-#include "Function.h"
-#include "Vec3.h"
+#include "TensorFunction.h"
 #include "Vec3Oper.h"
 #include "Utilities.h"
 #include "Profiler.h"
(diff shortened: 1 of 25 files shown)


kmokstad:

src/SIM/SIMbase.C
```diff
@@ -1625,34 +1625,60 @@ bool SIMbase::project (Matrix& ssol, const Vector& psol,
...
 
 
-bool SIMbase::project (Vector& values, const RealFunc* f,
-                       int basis, int iField, int nFields, double time) const
+bool SIMbase::project (Vector& values, const FunctionBase* f,
+                       int basis, int iField, int nFields,
+                       SIMoptions::ProjectionMethod pMethod, double time) const
 {
   bool ok = true;
+  values.resize(f->dim()*mySam->getNoNodes());
```
akva2: the addition of this resize call breaks the stokes and navier-stokes regression tests. the vector should be of total model size (mySam->getNoDOFs()), not f->dim*mySam->getNoNodes(). the vector already has the correct size set by the caller in existing code, if you need this for the l2 stuff you have to resize it up front or fix the size you set.

akva2:
LinEl test needs a version="2" to pass.

akva2:
having a common base for function is great. nice.

src/SIM/SIMbase.C
kmokstad: No, we can't use SAM::getNoDOFs() here, because the number of components per node in the secondary solutions that are projected, usually are different from the number of DOFs per node.
But I moved that resize to projectAnaSol instead to avoid the problem in Stokes/NavierStokes.

kmokstad:
`<CGL2 version="2/>` is now needed for problems with immersed boundaries, and nonlinear problems using internal integration point buffers in the L2-projections. For all others the default (version="1") is good. This is due to the changes in OPM/IFEM#161 though, not the new things here.

kmokstad:

akva2:
afaict none of this is a step back. even if it is not complete it's fine to merge as far as i am concerned.

kmokstad:
No merge yet, as this is still somewhat WIP.

kmokstad:
jenkins build this with downstreams ifem.elasticity=29 please

kmokstad:
jenkins build this with downstreams ifem.elasticity=29 please

kmokstad:
jenkins build this with downstreams ifem.elasticity=29 please

akva2:
jenkins build this with downstreams ifem-elasticity=29 please

PR #166, opened 2017-06-06
Static analyzer fixes / warning qulls

akva2:
This is a 'koldtbord' of commits addressing cppcheck warnings.

- I do not expect us to merge it in this form
- I fully expect some of the fixes to be rejected
- I hope that as many as possible are acceptable

Once we have gone over this and made our decision, I will start working on suppression. That way, we will eventually end up with a useful tool that will tell us when we introduce new issues.

Note I didn't manage to quite finish it today. Some 10 files missing still.


akva2:
Spoke with Trond, he told me you are squeezed on time. That is fine, this is a long-term thing. Easy question though; is it fine to mark single param constructors where we do not rely on implicit casts as explicit?

kmokstad:
Yes, that was one of the questions that struck me when I saw all this: Why do we now need to mark all (or most of) the constructors explicit, when it was not needed before? I guess that code analyzer recommend to do so, but I would like to see some human reasoning before doing that, and a clear rule for when it is needed, and when it is not necessary.

akva2:
it is never strictly needed, until you get burned. i have fixed 2-3 issues over the years that was caused by implicit casts hiding what was really going on. just to be clear, what we are talking about is situations like
```c++

class Something {
  Something(const int n) {..}
}

void some_method(const Something& foo, int bar) {... }

some_method(1, 2); -> some_method(Something(1), 2);
```
instead of a compilation error. these kinds of errors can happen when you change code around and whatnot. and sometimes, it won't lead to a complete crash, just behavior that appears to be sane but really isn't.

so it's more of a when not to. and the answer there is: when the implicit casts are too convenient to not utilize. like Vec3 -> Vector etc. otherwise the safe thing is to mark the constructors explicit. then the code above will not compile, you would have to explicitly do
```c++
some_method(Something(1), 2);
```
for it to compile.

VikingScientist:
I think these are great! Less warnings and doesn't break anything. I have no problems with any of these. Except of course that they need to be squashed together into logical topics so we don't spam the repository with too many commits.

VikingScientist:
Does anyone object to these changes? I like them, except for the shear amount of commits. I suggest squash & merge. The PR is already over 2 months old, and you can't leave PR's that touch 120 files lie unmerged for long or there is bound to be conflicts.

kmokstad:
I agree, a PR with 171 commits is too much. It must be squashed into more logic units before I will allow merge. Especially since each commit is quite tiny.

On the explicit constructors: That's fine, but is it needed for the copy-constructors as well? It looks as some of those are marked explicit as well (either here or in some of the apps). Does that make sense?

akva2:
i think i should merge the 171 commits as-is.

see what i did? same thing as you. i didn't read what was written prior.

akva2:
explicit cc's makes little sense. if i have done that, it's in error. will look out for them, but i take it the explicit is okay then, will pick those and squash into something suitable for merge....

kmokstad:
Is there anything left in this one that not already has been accounted for?

PR #167, opened 2017-06-08
Draft multipatch refine

akva2:
This is a draft for multipatch refinements.

I open early as it shows some layering/abstraction problems that needs addressing. ie i need to do ASMu2D stuff in SIMinput.


akva2:
@VikingScientist i am seeing issues in 3D and i'm a bit confused as i don't see what needs to be different there. please advice.

akva2:
updated: issues have also been triggered in 2D. problem identified -> back to the drawing board.

VikingScientist:
This will continue at PR#176

PR #169, opened 2017-06-12
External energy fix and SIMNodalConstraint simplification

kmokstad:
Only the first commit might be controversial, but I think the simplification is OK? Did not see any practical reason why not the apply method could not me a member of SIMNodalConstraint (but please bear with me if there are some template magic here I don't see). At least the unit tests still passes.

The other fix only affects models containing multi-point constraints (master-slave couplings) and calculates reaction forces, essentially only the C1-plate problems, as far as I know.

### DIFF
--- a/Apps/Common/SIMNodalConstraint.C
+++ b/Apps/Common/SIMNodalConstraint.C
@@ -326,63 +326,61 @@ protected:
 #endif
 
 
-template<>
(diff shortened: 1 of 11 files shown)


kmokstad:

akva2:
not controversial. at some point i needed it for specialization but design evolved. and it was simply not removed.

kmokstad:
jenkins build this with downstreams ifem-elasticity=28 ifem-poisson=8 please

PR #170, opened 2017-06-16
Fix CGL2 projection for ASMu2Dmx

akva2:
Copy-paste error I presume.


akva2:

PR #172, opened 2017-06-23
Collapse degenerated edges into a single node

kmokstad:
This adds the input-file tag `<collapse>` for 2D patches, which can be used when you have a polar parametrization where one of the edges collapse into a point. This then automatically assigns a common node number of all the collocated control points (replaces functionality formerly found in the getGNO utility). The extension to 3D (face to point and face to edge) is not there yet (but can be added).

I did this for the quarter-circular patch, which I ended up not using now after all, but...


akva2:
looks fine. i think it's good to add the others at the same time. also it seems appropriate with some tests.

PR #173, opened 2017-08-09
add storemesh with mixed

akva2:
respect &lt;store_eps_mesh/&gt; for mixed


akva2:
jenkins build this please

VikingScientist:
Any chance we might be able to squeeze in multipatch support in this PR while we're at it? The code would be pretty much copy-paste from https://github.com/VikingScientist/LRsplines/blob/master/Apps/drawLRmultipatch.cpp

If it proves tedious or if you don't have the time, then I see no reason to hold this PR back since there isn't anything wrong with it.

PR #175, opened 2017-08-09
Various fixes to support adaptive/mixed/projection

akva2:
This adds recovery functionality for mixed models, as well as support for multipatch refinement.


akva2:

src/ASM/LR/ASMu2Dmx.C
```diff
@@ -335,10 +337,18 @@ bool ASMu2Dmx::integrate (Integrand& integrand,
...
+              m_basis[b]->computeBasis(fe.u,fe.v,spline,els[b]-1);
+              SplineUtils::extractBasis(spline,fe.basis(b+1),dNxdu[b],d2Nxdu2[b]);
+            }
+          } else {
+            for (size_t b=0; b < m_basis.size(); ++b) {
+              Go::BasisDerivsSf spline;
+              m_basis[b]->computeBasis(fe.u, fe.v, spline, els[b]-1);
+              SplineUtils::extractBasis(spline,fe.basis(b+1),dNxdu[b]);
+            }
           }
```
kmokstad: Are the outer braces here for the if-else needed? Since only the for-statements in each block. Or do the code analyzer complain if left out.

src/ASM/LR/ASMu2Dmx.C
```diff
@@ -626,9 +626,89 @@ bool ASMu2Dmx::evalSolution (Matrix& sField, const Vector& locSol,
...
+    if (use2ndDer) {
+      if (!utl::Hessian(Hess,fe.hess(geoBasis),Jac,Xnod,
+                        d2Nxdu2[geoBasis-1],fe.grad(geoBasis),true))
+        continue;
+
+      for (size_t b = 0; b < m_basis.size(); ++b)
+        if ((int)b != geoBasis)
+          if (!utl::Hessian(Hess,fe.hess(b+1),Jac,Xnod,
+                d2Nxdu2[b],fe.grad(b+1),false))
+            continue;
```
kmokstad: This `continue` statement has no effect, so you can as well skip the innermost `if`also

src/ASM/LR/ASMu2Dmx.C
```diff
@@ -626,9 +626,89 @@ bool ASMu2Dmx::evalSolution (Matrix& sField, const Vector& locSol,
...
+    }
+
+    // Now evaluate the solution field
+    Vector   solPt;
```
kmokstad: only one space

src/ASM/LR/ASMu2Dmx.C
```diff
@@ -626,9 +626,89 @@ bool ASMu2Dmx::evalSolution (Matrix& sField, const Vector& locSol,
...
+    }
+
+    // Evaluate the basis functions at current parametric point
+    MxFiniteElement fe(elem_sizes);
+    std::vector<Matrix> dNxdu(m_basis.size());
+    Matrix Jac, Xnod;
+    std::vector<Matrix3D> d2Nxdu2(m_basis.size());
+    Matrix3D Hess;
+    if (use2ndDer)
+    {
```
kmokstad: Same question here on the braces as in the previous commit

src/SIM/SIMbase.C
```diff
@@ -1796,7 +1797,7 @@ bool SIMbase::setPatchMaterial (size_t patch)
 }
 
 
-bool SIMbase::addMADOF (unsigned char basis, unsigned char nndof)
+bool SIMbase::addMADOF (unsigned char basis, unsigned char nndof, bool otherbasis)
 {
```
kmokstad: You need a corresponding change in the header file SIMbase.h

src/SIM/SIMbase.h
```diff
@@ -600,7 +600,8 @@ class SIMbase : public SIMadmin, public SIMdependency
   //! \brief Adds a MADOF with an extraordinary number of DOFs on a given basis.
   //! \param[in] basis The basis to specify number of DOFs for
   //! \param[in] nndof Number of nodal DOFs on the given basis
-  bool addMADOF(unsigned char basis, unsigned char nndof);
+  //! \param[in] otherbasis If true, include other basis in madof
+  bool addMADOF(unsigned char basis, unsigned char nndof, bool otherbasis = true);
```
kmokstad: Ah, here it is. Please move this to the previous commit then

src/SIM/SIMbase.C
```diff
@@ -1285,7 +1285,7 @@ bool SIMbase::solutionNorms (const TimeDomain& time,
           if (ssol[k].empty())
             norm->getProjection(k).clear();
           else
-            pch->extractNodeVec(ssol[k],norm->getProjection(k),nCmp);
+            this->extractPatchSolution(ssol[k],norm->getProjection(k),lp-1,nCmp,1);
```
kmokstad: I'm not 100% sure about this change. Have a faint gut feeling there was something, sometime, when I had to use the extractNodeVec. But can't remember it now. Does it pass all the test? (They may not cover all aspects of norm evaluation though..)

src/ASM/IntegrandBase.C
```diff
@@ -325,7 +325,8 @@ bool NormBase::initElement (const std::vector<int>& MNPC,
                             const std::vector<size_t>& basis_sizes,
                             LocalIntegral& elmInt)
 {
-  return this->initProjection(MNPC,elmInt) &&
+  std::vector<int> MNPC1(MNPC.begin(), MNPC.begin()+elem_sizes.front());
+  return this->initProjection(MNPC1,elmInt) &&
          myProblem.initElement(MNPC,elem_sizes,basis_sizes,elmInt);
```
kmokstad: No need to copy the MNPC in the non-mixed case, so how about `if (MNPC.size() > elem_size.front())` make copy `else` as before

src/SIM/AdaptiveSIM.C
```diff
@@ -309,7 +336,11 @@ bool AdaptiveSIM::adaptMesh (int iStep)
   {
     IFEM::cout <<"\nRefining by increasing solution space by "<< beta
                <<" percent."<< std::endl;
-    prm.errors = eNorm.getRow(eRow);
+    std::vector<DblIdx> errors(model.getPatch(1)->getNoNodes(1));
+    remapErrors(errors, model.getPatch(1), eNorm.getRow(eRow));
```
kmokstad: But is this necessary in the non-mixed case then? How about the `remapErrors` method is virtual in ASMbase and with the above implementations in ASMu2Dmx and ASMu3Dmx, respectively, and one-to-one mapping in ASMbase?

src/ASM/LR/ASMu2Dmx.C
akva2: no, but my mental parser fails. my dislike for the insistence on not using braces for multi-line blocks is huge and i cannot follow it while coding because it slows me down a lot. and i sometime fails at applying the postprocessing step of removing them.

src/ASM/LR/ASMu2Dmx.C
akva2: a bit too copy-pasty, changed both to the intended return false.

src/SIM/SIMbase.C
akva2: yes, your long-term memory does not fail you. short term however.. ;) i did ask on skype about the breakage. in particular the local nodes.

src/ASM/IntegrandBase.C
akva2: this is a mixed overload. won't be here for non-mixed.

src/SIM/AdaptiveSIM.C
akva2: doh. this is why i want review ;) i've been living in a mixed-only world. will fix.

akva2:
all comments addressed. it still fails due to local nodes (at the very least), i will re-run jenkins to have fresh logs.

akva2:

VikingScientist:
I need this... could we make it merge?

akva2:
it's still causing issues with local nodes as we discussed in length on tuesday.

kmokstad:
This is awaiting my addressing of the issue. But since Trond wants me to get OpenFrac going as fast as possible, I'm not sure if I can look at it next week either, so I suggest @akva2 tries to isolate the only issue that was problematic in a separate PR and then you can push the rest here, such that the conflicts with #176 can be resolved, if critical.

akva2:
sadly this is unusable without the offending part. we wil just have to live with the branch for the time being. there are other issues anyways..

src/ASM/ASMbase.h
```diff
@@ -599,6 +599,17 @@ class ASMbase
...
+  //! \note The error value must be first and the index second, such that the
+  //! internally defined greater-than operator can be used when sorting the
+  //! error+index pairs in decreasing error order.
+  typedef std::pair<double,int> DblIdx;
+
+  //! \brief Remap element wise errors from geometry mesh to refinement mesh.
+  //! \param     errors The remapped errors
+  //! \param[in] patch The patch which holds the meshes
+  //! \param[in] origErr The element wise errors on the geometry mesh
+  virtual void remapErrors(std::vector<DblIdx>& errors, const Vector& origErr);
```
kmokstad: Does the default version need to do anything at all, i.e., will it ever be invoked for the non-LR sub-classes?
If the answer is "no", the empty method can be inlined with only anonymous arguments.
If yes, only remove the non-existing `patch` argument from doxy.

src/ASM/LR/ASMu2Dmx.h
```diff
@@ -194,6 +194,12 @@ class ASMu2Dmx : public ASMu2D, private ASMmxBase
                                 int basis, int thick = 1,
                                 int orient = 0, bool local = false) const;
 
+  //! \brief Remap element wise errors from geometry mesh to refinement mesh.
+  //! \param     errors The remapped errors
+  //! \param[in] patch The patch which holds the meshes
```
kmokstad: No argument `patch` here, remove from doxy.

src/ASM/LR/ASMu3Dmx.h
```diff
@@ -164,11 +164,16 @@ class ASMu3Dmx : public ASMu3D, private ASMmxBase
   virtual bool refine(const LR::RefineData& prm, Vectors& sol,
                       const char* fName = nullptr);
 
+  //! \brief Remap element wise errors from geometry mesh to refinement mesh.
+  //! \param     errors The remapped errors
+  //! \param[in] patch The patch which holds the meshes
```
kmokstad: Remove `patch` here also

src/SIM/AdaptiveSIM.C
```diff
@@ -15,10 +15,15 @@
 #include "SIMoutput.h"
 #include "SIMenums.h"
 #include "ASMunstruct.h"
```
kmokstad: Since you now include ASMu2D and ASMu3D below which both inherit ASMunstruct, there is no longer any need for this include statement.

src/SIM/AdaptiveSIM.C
```diff
@@ -15,10 +15,15 @@
...
+#include "ASMmxBase.h"
+#include "ASMu2D.h"
+#include "ASMu3D.h"
 #include "IntegrandBase.h"
 #include "Utilities.h"
 #include "IFEM.h"
 #include "tinyxml.h"
+#include "LRSpline/LRSplineSurface.h"
+#include "LRSpline/LRSplineVolume.h"
```
kmokstad: These two includes can now be removed also (not used).

src/ASM/ASMbase.h
akva2: This is changed in the hybrid between this and multipatch refinements so i will ignore this one. The answer is that yes, we transfer between data structures so it has to do things. Doxy is fixed in hybrid.

kmokstad:
If you pull #185 before this one, I think the problem on the added nodes associated with local coordinate systems and mixed is resolved (at least, a proposed solution). I tried this in my branch kmokstad/mix_project. However, the last commit that introduce the remapErrors function still cas\use some failures for some other tests. But I think that is unrelated.

akva2:
Great. The failures are likely expected , in particular if it is the stokes tests.

akva2:
this is now rebased, has the multipatch refinement integrated, all prior comments on either PRs have been addressed (i think).

akva2:
jenkins build this with downstreams ifem-poisson=9 ifem-elasticity=30 ifem-stokes=74 please

akva2:
note additional issues have been found which requires some restructuring, and that part is very WIP at the moment.

src/ASM/LR/ASMu2Dmx.C
```diff
@@ -353,6 +362,18 @@ bool ASMu2Dmx::integrate (Integrand& integrand,
...
 
+          // Compute Hessian of coordinate mapping and 2nd order derivatives
+          if (use2ndDer) {
+            if (!utl::Hessian(Hess,fe.hess(geoBasis),Jac,Xnod,
+                              d2Nxdu2[geoBasis-1],fe.grad(geoBasis),true))
+              ok = false;
+
+            for (size_t b = 0; b < m_basis.size() && ok; ++b)
+              if ((int)b != geoBasis)
+                ok &= utl::Hessian(Hess,fe.hess(b+1),Jac,Xnod,
```
kmokstad: Since the loop is terminated on ok=false, the &= operator is superfluous, just = will suffice. Alternatively, remove the && ok. (Actually the latter makes more sense since this Hessian call will never return false anyway).

src/ASM/ASMbase.h
```diff
@@ -183,6 +183,10 @@ class ASMbase
   //! \brief Returns the global node number for the given node.
   //! \param[in] inod 1-based node index local to current patch
   virtual int getNodeID(size_t inod, bool = false) const;
+  //! \brief Returns local 1-based index of the element with given global number.
+  //! \details If the given node number is not present, 0 is returned.
+  //! \param[in] globalNum Global node number
```
kmokstad: Error in doxy: Should be Global _element_ number

src/ASM/LR/ASMu2Dmx.C
```diff
@@ -908,3 +908,17 @@ void ASMu2Dmx::getBoundaryNodes (int lIndex, IntVec& nodes, int basis,
...
+
+void ASMu2Dmx::remapErrors(Vector& errors, const Vector& origErr) const
+{
+  const LR::LRSplineSurface* basis = this->getBasis(1);
+  const LR::LRSplineSurface* geo = this->getBasis(ASMmxBase::geoBasis);
+
+  for (const LR::Element* elm : basis->getAllElements()) {
+    int gEl = geo->getElementContaining((elm->umin()+elm->umax())/2.0,
+                                        (elm->vmin()+elm->vmax())/2.0) + 1;
+    for (const LR::Basisfunction* b: elm->support())
```
kmokstad: Cosmetics: space between "b:" for readability

src/ASM/LR/ASMu3Dmx.C
```diff
@@ -769,3 +769,18 @@ Vec3 ASMu3Dmx::getCoord (size_t inod) const
   }
   return Vec3(&(*basis->cp()),nsd);
 }
+
+
+void ASMu3Dmx::remapErrors(Vector& errors, const Vector& origErr) const
```
kmokstad: You don't really use the algebra operations on these argument, so it makes more sense to use RealArray (which is an alias for `std::vector<double>`) as argument type instead in this method

src/SIM/AdaptiveSIM.C
kmokstad: Still there?

src/SIM/AdaptiveSIM.C
kmokstad: Still there?

src/SIM/AdaptiveSIM.C
```diff
@@ -310,9 +318,17 @@ bool AdaptiveSIM::adaptMesh (int iStep)
...
+      return false;
+    }
     IFEM::cout <<"\nRefining by increasing solution space by "<< beta
                <<" percent."<< std::endl;
-    prm.errors = eNorm.getRow(eRow);
+    Vector errors(model.getPatch(1)->getNoNodes(1));
+    static_cast<ASMunstruct*>(model.getPatch(1))->remapErrors(errors, eNorm.getRow(eRow));
+    prm.errors.reserve(errors.size());
+    for (auto &it : errors)
+      prm.errors.push_back(it);
```
kmokstad: Looks unnecessary with the loop, why not just `prm.errors = errors`', or even better, pass prm.errors directly to remapError and avoid the copy all together.

src/SIM/SIMinput.C
```diff
@@ -1075,12 +1075,53 @@ bool SIMinput::refine (const LR::RefineData& prm,
...
+        int midx = this->getLocalPatchIndex(it.master);
+        ASMunstruct* pch = dynamic_cast<ASMunstruct*>(this->getPatch(midx));
+        ASMunstruct* spch = dynamic_cast<ASMunstruct*>(this->getPatch(sidx));
+        if (pch && spch)
+          extraRefine |= pch->matchNeighbour(spch, it.midx, it.sidx, it.orient);
+      }
+#ifdef HAVE_MPI
+      //for (const ASM::Interface& it : adm.dd.ghostConnections) // TODO
+#endif
+    }
```
kmokstad: More logic with `do { ... } while (extraRefine);` here

src/ASM/LR/ASMu2Dmx.C
akva2: done

src/ASM/LR/ASMu2Dmx.C
akva2: the && ok is there to avoid doing this on prior error status. removed the ok &=.

src/ASM/ASMbase.h
akva2: done

src/ASM/LR/ASMu3Dmx.C
akva2: changed.

src/SIM/AdaptiveSIM.C
akva2: sorry cleaned SIMinput missed this one. ASMuxD.h is removed, ASMunstruct is kept and the LR includes are now gone. the loop was a leftover from the old code using a different data structure and has been changed to remapping directly

src/SIM/SIMinput.C
akva2: i assume you meant logical. done.

akva2:
jenkins build this with downstreams ifem-poisson=9 ifem-elasticity=30 ifem-stokes=74 please

PR #177, opened 2017-08-10
Dump of FE model to Matlab format

kmokstad:
to facilitate comparison with Matlab solvers. So far only 2D linear quads.

The other commit is something else, an option scaling of the parameter domain when specifying non-uniform knots.

### DIFF
--- a/src/SIM/SIMoutput.C
+++ b/src/SIM/SIMoutput.C
@@ -201,8 +201,8 @@ bool SIMoutput::parse (char* keyWord, std::istream& is)
 
 void SIMoutput::preprocessResultPoints ()
 {
-  for (size_t i = 0; i < myPoints.size(); i++)
-    this->preprocessResPtGroup(myPoints[i].first,myPoints[i].second);
+  for (ResPtPair& rptp : myPoints)
+    this->preprocessResPtGroup(rptp.first,rptp.second);
 }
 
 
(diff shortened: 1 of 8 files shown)


src/SIM/SIMoutput.C
```diff
@@ -1110,7 +1110,82 @@ bool SIMoutput::writeGlvE (const Vector& vec, int iStep, int& nBlock,
...
+                                double scale) const
+{
+  // Write function definition
+  os <<"function [Node, Element";
+  for (const std::string& setname : sets) os <<", "<< setname;
+  os <<"]="<< name << std::endl;
+
+  // Write out nodes for the specified topology sets
+  ASMbase* pch;
+  TopologySet::const_iterator tit;
```
akva2: consider limiting scope as analyzers will complain.

src/SIM/SIMoutput.C
```diff
@@ -1110,7 +1110,82 @@ bool SIMoutput::writeGlvE (const Vector& vec, int iStep, int& nBlock,
...
+      else if (count%16 == 1)
+        os <<",...\n    "<< node;
+      else
+        os <<", "<< node;
+    os <<"];"<< std::endl;
+  }
+
+  // Write out all nodes
+  os <<"\n  Node=[1";
+  size_t nnod = mySam->getNoNodes();
```
akva2: should this consider mixed, or at least fail and not dump the entire model ?

kmokstad: It will not make sense to dump the whole model in the mixed case, only the first (geometry) basis, so maybe 'D' should be added

kmokstad: On second thought, the element dump will in any case be invalid in case of mixed here, so instead I added an error return if meshes with more that 4 element nodes are encountered, and only 2D models.

PR #176, opened 2017-08-09
Draft multipatch refine

VikingScientist:
Continuation of PR #167

PR #178, opened 2017-08-10
fix ASMu3D::evalSolution and ASMu3D::updateDirichlet for multi-patch

akva2:
also unify ASMu3D and ASMu2D code a little. note the first
also requires an updated libLRSpline


akva2:

VikingScientist:
Tested and this works :+1:

One could argue that we should bump the LRspline version so we can ensure that the fix is also present upstream, but PR #176 is going to require more extensive update, so I think it is good enough to wait for the version bump that is going to accompany that one.

PR #179, opened 2017-08-10
Bugfix in 3D LR recovery

VikingScientist:


akva2:
i think it would be wise to flip the fix. nothing else in asm interface speaks 0 indices. sure, it's internal but...

VikingScientist:
`getElementCoordinates()` and `getParametricVolume()` used right above this both takes 1-indexed input.

akva2:
yes, which was my point. or i don't follow.

akva2:
just in case my formulation was unclear: i think it's better to do the subtraction inside getGaussParameters instead of in the caller code to have it behave the same - 1-based indices going into the function.

VikingScientist:
oh... I was under the impression that the entire `iel`-variable should be 0-based, but in the `ASMu3D::assembleL2matrices` it needs to be used as both 0 and 1. But you area right, one might make it 1-indexed inside `ASMu3D::getGaussPointParameters`. I'll fix this tomorrow.

akva2:
It is 0 based when interacting with SAM. I think that was the system i found in the madness at some point.

VikingScientist:
jenkins build this please

VikingScientist:
jenkins build this please

akva2:

akva2:
all good, going in.

PR #180, opened 2017-08-13
Fixed: The code inside the #if SP_DEBUG > 2 did not compile.

kmokstad:
B is therefore changed into a single vector (of dimension nnod*ncomp)
instead of an array of vectors of length nnod. The equation solver
can handle this format, no need to loop over the RHS outside here.
Also some indentation fixes here and there.

### DIFF
--- a/src/ASM/LR/ASMu3Drecovery.C
+++ b/src/ASM/LR/ASMu3Drecovery.C
@@ -15,16 +15,14 @@
 #include "LRSpline/Element.h"
 
 #include "ASMu3D.h"
+#include "FiniteElement.h"
+#include "IntegrandBase.h"
 #include "CoordinateMapping.h"
 #include "GaussQuadrature.h"
 #include "SparseMatrix.h"
 #include "DenseMatrix.h"
-#include "FiniteElement.h"
 #include "SplineUtils.h"
-#include "Utilities.h"
 #include "Profiler.h"
-#include "IntegrandBase.h"
-#include "Vec3.h"
 #include <array>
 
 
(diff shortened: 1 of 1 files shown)


src/ASM/LR/ASMu3Drecovery.C
```diff
@@ -558,88 +553,80 @@ bool ASMu3D::faceL2projection (const DirichletFace& edge,
...
+        for (il = 0; il < edge.MNPC[ie].size(); il++) // local i-index
+          if ((ig = 1+edge.MNPC[ie][il]) > 0)         // global i-index
+          {
+            for (jl = 0; jl < edge.MNPC[ie].size(); jl++) // local j-index
+              if ((jg = 1+edge.MNPC[ie][jl]) > 0)         // global j-index
+                A(ig,jg) += fe.N[il]*fe.N[jl]*fe.detJxW;
+
+            RealArray val = values.getValue(X);
+            for (size_t k = 0; k < m; k++)
+              B(ig+(k-1)*n) += fe.N[il]*val[k]*fe.detJxW;
```
akva2: k*n, not (k-1)*n

akva2:
except the above, this seems fine. please pick https://github.com/akva2/IFEM/tree/edge_proj (and review) while you are at it.

src/ASM/LR/ASMu3Drecovery.C
kmokstad: Fixed.

kmokstad:
Didn't pick that branch yet, since not sure how many commits there are relevant. But I reviewed it briefly and added a comment. So if this one does not break any tests it is ready to be merged (I need it, just for compiling Debug mode).

akva2:
Then merge

PR #181, opened 2017-08-14
make single-parameter constructors as explicit

akva2:
to avoid unexpected implicit casts

### DIFF
--- a/3rdparty/expreval/except.h
+++ b/3rdparty/expreval/except.h
@@ -69,7 +69,7 @@ namespace ExprEval
     class NotFoundException : public Exception
     {
     public:
-        NotFoundException(const ::std::string &name);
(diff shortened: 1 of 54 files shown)


akva2:

akva2:
once more with better timing.

akva2:

kmokstad:
I see you already merged this, even before I just noticed it was opened a separate PR for it. But that's fine this one. The only thing I could point on is when the constructor is protected, such explicit declaration seems unnecessary, since it then will be visible for sub-classes only. Then less (or no) danger for the mentioned auto-casting issues? Similar for private nested classes. But of course it will not cause any harm either, so this is just a comment.

akva2:
I agree. Sadly cppcheck doesnt..

PR #182, opened 2017-08-14
New ASM class for (linear) Lagrange meshes from file

kmokstad:
The added class ASMs2DMatlab is a subclass of ASM2DLag and reads the mesh file dumped by the SIMoutput:dumpMatlabGrid method, which corresponds to the mesh-files from the S_Miehe matlab solver. All this for the convenience of comparing the two solvers on the same grid.


VikingScientist:
Rename to avoid calling it 'matlab'

src/ASM/ASMs2DMatlab.C
```diff
@@ -0,0 +1,211 @@
...
+    {
+      while (readLine(is,cline))
+      {
+        size_t id; Vec3 X;
+        std::stringstream(cline) >> id >> X;
+#if SP_DEBUG > 1
+        std::cout << id <<": "<< X << std::endl;
+#endif
+	this->setCoord(id,X);
+        if (cline.find_last_of(';') == cline.length()-1) break;
```
akva2: cline.back() == ';'

src/ASM/ASMs2DMatlab.C
```diff
@@ -0,0 +1,211 @@
...
+      std::string setname = cline.substr(0,cline.length()-1);
+      if (readLine(is,cline,']'))
+      {
+	size_t i = 0; // Remove the '...' continuation markers
+	while ((i = cline.find_first_of('.',i)) != std::string::npos)
+	  cline.erase(i,1);
+
+	int node;
+	IntVec& nodes = nodeSets[setname];
+        std::istringstream selem(cline);
```
akva2: indent

src/ASM/ASMs2DMatlab.C
```diff
@@ -0,0 +1,211 @@
...
+
+  return true;
+}
+
+
+bool ASMs2DMatlab::generateFEMTopology ()
+{
+  p1 = p2 = 2; // So far only linear elements supported
+
+  nx = sqrt(nnod);
```
akva2: nx = ny  = sqrt(nnod) ?

src/ASM/ASMs2DMatlab.C
```diff
@@ -0,0 +1,211 @@
...
+
+  return true;
+}
+
+
+int ASMs2DMatlab::getNodeSetIdx (const std::string& setName) const
+{
+  std::map<std::string,IntVec>::const_iterator it = nodeSets.begin();
+  for (int i = 1; it != nodeSets.end(); i++, ++it)
+    if (it->first == setName) return i;
```
akva2: strictly speaking i may be invalidated by additional inserts (not going to happen so just a knee-jerk).

src/ASM/ASMs2DMatlab.C
```diff
@@ -0,0 +1,211 @@
...
+}
+
+
+void ASMs2DMatlab::generateThreadGroups (const Integrand&, bool, bool)
+{
+  // TODO: Add some coloring scheme later
+  threadGroups[0].resize(1);
+  threadGroups[1].resize(0);
+  threadGroups[0][0].resize(nel);
+  std::iota(threadGroups[0][0].begin(),threadGroups[0][0].end(),0);
```
akva2: maybe we should consider sticking this in base ?

src/ASM/ASMbase.C
```diff
@@ -509,6 +509,20 @@ void ASMbase::constrainPatch (int dof, int code)
 }
 
 
+void ASMbase::constrainNodes (const IntVec& nodes, int dof, int code)
```
akva2: nice. we can probably use this to simplify some other methods at some point.

src/ASM/ASMs2DMatlab.C
kmokstad: fixed

src/ASM/ASMs2DMatlab.C
kmokstad: was some left-over tabs inserted by emacs, cleaned now

src/ASM/ASMs2DMatlab.C
kmokstad: Doesn't really matter, as they are not used anyway when the mesh is unstructured. Removed.

src/ASM/ASMs2DMatlab.C
kmokstad: True, but the assumption is that the nodeSets is defined only by the read method, and after that left untouched. Therefore the access method only returns a const reference to it.

src/ASM/ASMs2DMatlab.C
kmokstad: yes

kmokstad:
Renamed the new class to ASMu2DLag instead, as suggested, and added the matlab-syntax reading as a separate method. Then, it can be extended by other formats when needed.

kmokstad:

src/ASM/ASMu2DLag.C
```diff
@@ -195,24 +196,38 @@ bool ASMu2DLag::generateFEMTopology ()
...
 }
 
 
+IntVec& ASMu2DLag::getNodeSet (const std::string& setName)
+{
+  for (NodeSet& nset : nodeSets)
+    if (nset.first == setName)
+      return nset.second;
+
+  nodeSets.push_back(std::make_pair(setName,IntVec()));;
```
akva2: maybe std::unordered_map is more convenient ?

PR #183, opened 2017-08-21
fixed: generate topology and topologysets even if the model is pre-existing

akva2:
If not the connections are lost on refinementstep > 1.


akva2:

kmokstad:
This looks OK, don't see any issues here. But the commit message needs some explanation: Is it so that the connections are not lost in the first refinement step, but in the second and beyond? That sounds wierd.

akva2:
just how you count, first step is prior to refinement. pushed with a more descriptive commit message.

PR #184, opened 2017-08-21
Fixed: Partly undo of commit 89faa82c6f972e01029f2ec4e11f1f0e4865bdb2

kmokstad:
Do only one iteration if the first iteration already satisfies the
convergence tolerance when refNopt=ALL. No check on subiter&FIRST here.

This only removes some unnecessary iterations in cases with sub-iterations, and may affect tests on apps using that feature (but hopefully only to the better).


akva2:

akva2:
no objections from me, all tests pass.

PR #185, opened 2017-08-27
Changed: The bool argument of SIMbase::getNoNodes is removed.

kmokstad:
Its default value (false) was not used in practice and it was
thus only a potential source of confusion.
Fixed: Also include the 'X'-nodes (if any) when counting the nodes
associated with the first basis. That means, we assume that only
the first basis can have Dirichlet boundary conditions defined in
local coordinate axes.

If you pull this one before #175, the unresolved problem mentioned there regarding the local coordinate system nodes should be fixed, I think. At least, this is a proposed solution to it.

### DIFF
--- a/Apps/Common/SIMSemi3D.h
+++ b/Apps/Common/SIMSemi3D.h
@@ -114,7 +114,7 @@ public:
   {
     planeNodes.resize(this->getNoPlanes());
     for (size_t i=0;i<m_planes.size();++i)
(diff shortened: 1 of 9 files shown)


kmokstad:
jenkins build this with downstreams ifem-stokes=73 ifem-navierstokes=71 ifem-fsi=22 please

akva2:
all good. can i merge ?

PR #188, opened 2017-08-31
Fixed: Copy the norm prefixes to avoid valgrind complaint

kmokstad:
This resolves the memory issue http://afem:8000/job/IFEM-memtest/1010/

The problem is that AdaptiveSIM keeps an array of C-pointers to the norm prefices to use, but when the mesh is refined and the input-file is parsed again for the new mesh, these pointers are invalidated (at least when run through valgrind). Now I make a copy of those strings instead and the problem vanishes. Maybe not the most elegant, but works.


akva2:
if it does the job. personally i simply always use std::string and not char* to avoid these things ;)

akva2:
merging. fed up with failure emails ;)

PR #190, opened 2017-09-07
changed: use bezier extraction in evaluation of secondary solution.

akva2:
Faster, and works around issues in the computeBasis method for some models.

@VikingScientist

### DIFF
--- a/src/ASM/LR/ASMu3D.C
+++ b/src/ASM/LR/ASMu3D.C
@@ -1963,6 +1963,14 @@ bool ASMu3D::evalSolution (Matrix& sField, const IntegrandBase& integrand,
   Matrix   dNdu, Jac, Xnod;
   Matrix3D d2Ndu2, Hess;
 
+  const int p1 = lrspline->order(0);
+  const int p2 = lrspline->order(1);
+  const int p3 = lrspline->order(2);
+  Go::BsplineBasis basis1 = getBezierBasis(p1);
+  Go::BsplineBasis basis2 = getBezierBasis(p2);
+  Go::BsplineBasis basis3 = getBezierBasis(p3);
+  Matrix B(p1*p2*p3, 4); // Bezier evaluation points and derivatives
+
   // Evaluate the secondary solution field at each point
   for (size_t i = 0; i < nPoints; i++)
   {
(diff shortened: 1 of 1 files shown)


src/ASM/LR/ASMu3D.C
```diff
@@ -1982,16 +1990,41 @@ bool ASMu3D::evalSolution (Matrix& sField, const IntegrandBase& integrand,
...
+    basis1.computeBasisValues(gpar[0][i], u, 1);
+    basis2.computeBasisValues(gpar[1][i], v, 1);
+    basis3.computeBasisValues(gpar[2][i], w, 1);
+
+    size_t ib=1; // basis function iterator
+    for(int k=0; k<p3; k++) {
+      for(int j=0; j<p2; j++) {
+        for(int ii=0; ii<p1; ii++, ib++) {
+          B(ib,1) = u[2*ii  ]*v[2*j  ]*w[2*k  ];
+          B(ib,2) = u[2*ii+1]*v[2*j  ]*w[2*k  ]*2.0/dv;
```
VikingScientist: I guess this should be divided by du, and not dv

src/ASM/LR/ASMu3D.C
```diff
@@ -1982,16 +1990,41 @@ bool ASMu3D::evalSolution (Matrix& sField, const IntegrandBase& integrand,
...
+    double u[2*p1];
+    double v[2*p2];
+    double w[2*p3];
+    basis1.computeBasisValues(gpar[0][i], u, 1);
+    basis2.computeBasisValues(gpar[1][i], v, 1);
+    basis3.computeBasisValues(gpar[2][i], w, 1);
+
+    size_t ib=1; // basis function iterator
+    for(int k=0; k<p3; k++) {
+      for(int j=0; j<p2; j++) {
```
VikingScientist: I would like these to be `jj` and `kk` just for consistency.

VikingScientist:
Tested this and in the case of `du==dv`, then it works as advertised with speedup and with no crash during secondary solution evaluation anymore. Will merge after fixup.

src/ASM/LR/ASMu3D.C
akva2: nice spotting, bad fingers.

src/ASM/LR/ASMu3D.C
akva2: done.

VikingScientist:
Jenkins build this please

akva2:
and now you learned the hard way that you have to test things with downstreams.

PR #194, opened 2017-09-12
Changed: Undo of 02d84c07e2dc79d3adeec067a47c5fe8d878f3ce

kmokstad:
It feels unnecessary and confusing with two versions of solveStep
and the need for the using statement when one of them is overridden?

I think the reason for the May 5th commit on this is that it was forgotten up update the overridden solveStep method in the one and only downstream application that did override it (SIMKEpsilonBase).
I believe it is much better(?) to update that signature also such that we don't need to overload this method, and consequently can skip the annoying need for using-statements.

### DIFF
--- a/Apps/Common/SIMCoupled.h
+++ b/Apps/Common/SIMCoupled.h
@@ -53,13 +53,7 @@ public:
   }
 
   //! \brief Computes the solution for the current time step.
-  virtual bool solveStep(TimeStep& tp)
-  {
-    return this->solveStep(tp, true);
-  }
-
-  //! \brief Computes the solution for the current time step.
-  virtual bool solveStep(TimeStep& tp, bool firstS1)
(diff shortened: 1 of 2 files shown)


kmokstad:
jenkins build this with downstreams ifem-navierstokes=72 ifem-openfrac=26 please

akva2:
this was a hotfix to restore ci after inconsiderate changes in openfrac. if you have found a better way it's all good.

kmokstad:
That's good. But since the check failed I'll wait to pull it in until you have fixed that unrelated issue (I suspect it is the mentioned LR-thing and nothing to do with this).

akva2:

akva2:
jenkins build this with downstreams ifem-navierstokes=72 ifem-openfrac=26 please

PR #195, opened 2017-09-18
Nodeset fix

kmokstad:
It seems like the unordered_map class does not work as I anticipated. It do sorts on the key which makes the unit test on Matlab input fail (first commit). In the second commit I replaced by a vector of pairs instead which works better.

Is this a bug in STL or is it supposed to work like this?


kmokstad:

PR #196, opened 2017-09-18
Added: Dump of solution matrix/vectors of several steps to the same file

kmokstad:
Convenient if you want to compare a time history in IFEM with a matlab implementation


kmokstad:

PR #198, opened 2017-09-22
changed: print CFL for negative maxCFL value

akva2:
this allows to print but not apply time stepping control


akva2:
jenkins build this with downstreams ifem-navierstokes=73 please

PR #199, opened 2017-09-26
Lu solver stuff

akva2:
This fixes some mess related to usage of PETSc LU solver and holds a minimal effort UMFPACK addition.

1) sudo apt-get install libsuitesparse-dev
2) configure ifem with -DIFEM_USE_UMFPACK=1
3) give -umfpack command line parameter or use &lt;linearsolver class="umfpack"/&gt;

@kmokstad tested working for the fracture case. the umfpack effort is minimal, missing a few things like condition number, factorization re-use, multiple RHS and cholesky solver (for SPD matrices) but it should be enough that you can move on with testing.

the documentation in /usr/include/suitesparse/umfpack_(solve|numeric|symbolic).h is fairly okay, so consult that.


kmokstad:
This is OK, but everything (except the first commit) need to be squashed though.
And there should perhaps be added a supported package statement in IFEM::Init, with version info, if relevant.

akva2:
sanity check

akva2:

PR #204, opened 2017-10-04
fixed: lower-dimension connections is not an error

akva2:
they are necessary in parallel, and depending on how the model is partitioned, they may be process-local. we should just ignore them.

### DIFF
--- a/src/SIM/SIM2D.C
+++ b/src/SIM/SIM2D.C
@@ -82,7 +82,7 @@ bool SIM2D::addConnection (int master, int slave, int mIdx,
   int lslave = this->getLocalPatchIndex(slave);
   if (lmaster > 0 && lslave > 0)
   {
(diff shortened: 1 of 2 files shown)


src/SIM/SIM2D.C
```diff
@@ -82,7 +82,7 @@ bool SIM2D::addConnection (int master, int slave, int mIdx,
   int lslave = this->getLocalPatchIndex(slave);
   if (lmaster > 0 && lslave > 0)
   {
-    if (dim < 1) return false;
+    if (dim == 0) return true; // ignored in serial
```
kmokstad: Does negative dim have some interpretation for 2D? If not, maybe change to `if (dim < 1);` to make it similar as the 3D case

akva2: changed.

PR #205, opened 2017-10-04
add support for sub-grid mixed elements

akva2:
this is two field formulation with C^{p+1}/p and C^p/p-1 bases, but where the first basis has
an additional subdivision of each element.

this is an inf-sup stable basis for Stokes.

this required rewriting the MNPC establishment for tensor splines, previous implementation assumed the same element structure for all grids.

### DIFF
--- a/src/ASM/ASMmxBase.C
+++ b/src/ASM/ASMmxBase.C
@@ -229,6 +229,66 @@ ASMmxBase::SurfaceVec ASMmxBase::establishBases(Go::SplineSurface* surf,
                                                                   u0,vg,XYZ1,ndim,
                                                                   false,XYZ1));
     geoBasis = 3;
+  } else if (type == SUBGRID) {
+    // basis1 should be one degree higher than basis2 and C^p-1 continuous
+    int ndim = surf->dimension();
+    result[1].reset(new Go::SplineSurface(*surf));
+    Go::SplineSurface tmp(*surf);
+    for (size_t i = 0; i < 2; ++i) {
+      RealArray extraKnots;
+      RealArray::const_iterator uit = tmp.basis(i).begin();
+      double ucurr, uprev = *(uit++);
+      while (uit != tmp.basis(i).end())
+      {
+        ucurr = *(uit++);
+        if (ucurr > uprev)
+          extraKnots.push_back(ucurr*0.5 + uprev*0.5);
+        uprev = ucurr;
+      }
+      if (i == 0)
+        tmp.insertKnot_u(extraKnots);
+      else
+        tmp.insertKnot_v(extraKnots);
+    }
+    Go::BsplineBasis b1 = tmp.basis(0).extendedBasis(tmp.order_u()+1);
+    Go::BsplineBasis b2 = tmp.basis(1).extendedBasis(tmp.order_v()+1);
+
+    // Compute parameter values of the Greville points
+    size_t i;
+    RealArray ug(b1.numCoefs()), vg(b2.numCoefs());
+    for (i = 0; i < ug.size(); i++)
+      ug[i] = b1.grevilleParameter(i);
+    for (i = 0; i < vg.size(); i++)
+      vg[i] = b2.grevilleParameter(i);
+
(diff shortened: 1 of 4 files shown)


akva2:

src/ASM/ASMmxBase.C
```diff
@@ -229,6 +229,70 @@ ASMmxBase::SurfaceVec ASMmxBase::establishBases(Go::SplineSurface* surf,
...
+    Go::SplineSurface tmp(*surf);
+    for (size_t i = 0; i < 2; ++i) {
+      RealArray extraKnots;
+      RealArray::const_iterator uit = tmp.basis(i).begin();
+      double ucurr, uprev = *(uit++);
+      while (uit != tmp.basis(i).end())
+      {
+        ucurr = *(uit++);
+        if (ucurr > uprev)
+          for (int i = 0; i < 1; i++)
```
VikingScientist: This is the smallest for-loop I ever laid eyes on. Either take some "number-of-subdivisions-per-element" parameter in, or consider removing it

src/ASM/ASMmxBase.C
```diff
@@ -229,6 +229,70 @@ ASMmxBase::SurfaceVec ASMmxBase::establishBases(Go::SplineSurface* surf,
...
+      // new rational coefs are (x/w', y/w', w')
+      // apparently gotools will rescale coeffs on surface creation.
+      result[0].reset(new Go::SplineSurface(surf3->basis(0), surf3->basis(1), surf3->coefs_begin(), ndim, true));
+    } else {
+      RealArray XYZ(ndim*ug.size()*vg.size());
+      // Evaluate the spline surface at all points
+      tmp.gridEvaluator(XYZ,ug,vg);
+
+      // Project the coordinates onto the new basis (the 2nd XYZ is dummy here)
+      result[0].reset(Go::SurfaceInterpolator::regularInterpolation(b1,b2,
```
VikingScientist: What you are doing is a pure knot insertion scheme. Would it not be shorter to simply clone `*surf` followed by knot insertion on this object. This should compute all the controlpoints to their proper value without interpolation schemes

src/ASM/ASMmxBase.C
```diff
@@ -352,6 +416,71 @@ ASMmxBase::VolumeVec ASMmxBase::establishBases(Go::SplineVolume* svol,
...
+    } else {
+      RealArray XYZ(ndim*ug.size()*vg.size()*wg.size());
+      // Evaluate the spline surface at all points
+      tmp.gridEvaluator(ug,vg,wg,XYZ);
+
+      // Project the coordinates onto the new basis (the 2nd XYZ is dummy here)
+      result[0].reset(Go::VolumeInterpolator::regularInterpolation(b1,b2,b3,
+                                                                   ug,vg,wg,
+                                                                   XYZ,ndim,
+                                                                   false,XYZ));
```
VikingScientist: Again: could not the interpolation be swapped out for knot insertion?

src/ASM/ASMmxBase.C
akva2: copy-paste from uniformRefine - quick solutions for quick living.. i can do better.

src/ASM/ASMmxBase.C
akva2: probably. i just copied the other code and made adjustments, will look into it.

VikingScientist:
Trying this out for the first time and my immediate reaction is that it was strange that GLview files do not put the pressure solution on the pressure mesh. In fact: the pressure mesh is nowhere to be found in the output files. How challenging would it be to make this happen? Could one fake it by making "parts" the pressure mesh, like LRsplines uses parts for element boundaries and mesh for element-interior sampling points.

VikingScientist:
Also: remember to update the wiki once this is in

src/ASM/ASMmxBase.C
akva2: as discussed on skype i do order raising along with the knot insertion so cannot comply.

PR #209, opened 2017-10-11
added: reuse of symbolic factorization for umfpack solver

akva2:
No idea why I thought this wasn't doable. Need more glasses!


kmokstad:
Looks OK, so if it works, no reason to delay merging it

VikingScientist:
More glasses for @akva2 :cocktail: :beer:

PR #214, opened 2017-10-22
Option to specify msgLevel on the input file

kmokstad:
To facilitate suppression of threading group info and VTF output which can be substatial for big multi-patch models. But it is still nice to be able to get such output without compiling debug


kmokstad:

kmokstad:

PR #216, opened 2017-10-26
stuff required to use fields as anasol functions

akva2:
Not intended for merge in this form, just as a basis for discussion.

This is my current solution for using fields from hdf5 as anasol's for comparison purposes.
Searching based on coordinates is broken in GoTools, and missing for LR splines. I thus opted for this other approach, where we include parameter values in the Vec3 passed for evaluation.
Likewise, we have to include parameter for tesselation points for VTF output.

The code is not clean, and this should be multiple commits etc. But I want comments on the approach, whether it is too invasive etc.

There will be a companion PR in IFEM-Poisson. In particular, I do not like the requirement to pass the entire anasol into the norms, but as the norm integrand is instantiated before the patch loop, it was required to grab the current function instance for the patch. Perhaps we need to bake the multiple fields into a common (Vec)FieldFunction instead. In that case we need to tell multiple functions rather than just the anasol class which patch is being processed...

```
<anasol type="fields" file="something">
  <primary>u</primary>
  <secondary>q_x|q_y|q_z</secondary>
</anasol>
```


akva2:
jenkins build this with downstreams ifem-poisson=18 please

src/ASM/LR/LRSplineFields2D.C
```diff
@@ -85,8 +85,16 @@ bool LRSplineFields2D::valueFE (const FiniteElement& fe, Vector& vals) const
 
 
 bool LRSplineFields2D::valueCoor (const Vec3& x, Vector& vals) const
-{
+{  if (x.params[0]) {
```
kmokstad: Newline between the first brace and the "if" please

src/Utility/Vec3.h
```diff
@@ -31,6 +31,7 @@ class Vec3
   Real& x; //!< Reference to X-component
   Real& y; //!< Reference to Y-component
   Real& z; //!< Reference to Z-component
+  const Real* params[3] = {};
```
kmokstad: Don't see why we need three pointers. Should be sufficient with just `Real* params;` here which then points to an array with parameter values. Or alternatively (maybe also prefarable), consider a Vec3 (or Vec4 maybe( sub-class which have the u, v and w parameters as additional members.

akva2: as i said this was just quickly tossed together, and in the fe class it was three vars so i just made three pointers so i could point directly (using the ability of them being null to flag searching required). i take it you do not think the approach is too invasive then?

kmokstad:
I have now looked at this in some more detail, and some suggested changes on the branch kmokstad/anasol_fields. It consists of three commits with the following main points:

* Changed into using `Vec4` instead of `Vec3` in the `valueCoor` and `gradCoor` methods. Implemented dummy versions of these methods in the base classes `Field`and `Fields` to avoid the need for dummies in all sub-classes without support for coordinate-evaluation. There where also some errors in the LRSplineFields[23]D methods I think, where the indices of some `Xnod` and `Vnod`arrays appeared to be swapped. These methods can not have been used yet (any unit tests)?

* The next commit is just a rebased version of this PR.

* The third and last commit is the proposed changes that are to be squashed with the previous commit, when/if approved: Here I have moved the new `params` member of Vec3 to Vec4 instead, and it is now a single `const Real*`pointer instead of an array of three, which appears confusing to me. All the field `valueCoor` methods as well as `FieldFunction(s)` and `SIMoutput` are updated accordingly. I also added proper initialization of the Vec4-parameters in the `ASMs2D` and ` ASMu2D` integrate methods (other ASM-classes needs to be done as well).

I haven't tested it (only compiled), but if you think this can work you can now go ahead and implement the rest with this framework.

akva2:
picked your changes. had to do some small fixes

- add assignment operator for vec3/4 from std::vector<Real> if not the params are lost in the X = Xnod*N calls in the integration loops since we go through the Vec3 constructor.
- patches apparently are 0 indexed.
- add patch selection for anasol in boundary integration for anasol derived neumann conditions.

these sit in a commit at end, intended to be squashed. i can now run things in poisson with minimal changes (have to inject the whole anasol and not just the function as previously discussed).

src/Utility/FieldFunctions.h
```diff
@@ -28,7 +29,7 @@ class ASMbase;
 class FieldFunction : public RealFunc
 {
   Field* field; //!< Pointer to the scalar field to be evaluated.
```
kmokstad: Cosmetic; no period '.' in one-liner member documentation

src/Utility/FieldFunctions.h
```diff
@@ -48,4 +53,36 @@ class FieldFunction : public RealFunc
...
 };
 
+
+/*!
+  \brief A vector-valued spatial function, defined trough a vector field.
+*/
+
+class VecFieldFunction : public VecFunc
+{
+  Fields* field; //!< Pointer to the vector field to be evaluated.
```
kmokstad: Here too, no period.

src/Utility/AnaSol.h
```diff
@@ -39,7 +39,7 @@ class AnaSol
   STensorFunc* stressSol; //!< Secondary solution field (stress field)
 
   size_t nPatches = 0; //!< Number of patches for field-type functions.
```
kmokstad: No period

src/Utility/Vec3.h
```diff
@@ -216,6 +219,9 @@ class Vec4 : public Vec3
   { x = X.x; y = X.y; z = X.z; t = X.t; idx = X.idx; u = X.u; return *this; }
   //! \brief Overloaded assignment operator.
   Vec4& operator=(Real val) { x = y = z = val; return *this; }
+  //! \brief Overloaded assignment operator.
+  Vec4& operator=(const std::vector<Real>& X)
+  { this->Vec3::operator=(X); return *this; }
```
kmokstad: If we need these overloads, shouldn't it also be a `t = X.size() > 3 ? X[3] : 0.0; idx = -1; ` in the latter to be consistent with the `Vec4(const std::vector<Real>& X)` constructor?

Alternatively, one could use `Vec4::assign(const Vec3&)` instead of this assign operator, which does not blank out the `u` pointer when the argument is not a `Vec4`.

src/Utility/AnaSol.h
```diff
@@ -125,7 +126,7 @@ class AnaSol
       return scalSecSol.size() <= idx ? nullptr : scalSecSol[idx];
 
     return scalSecSol.size() <= nPatches*idx+patch ?
-                                nullptr : scalSecSol[nPatches*idx+patch];
+                                nullptr : scalSecSol[nPatches*idx + patch];
```
kmokstad: The body of getScalarSol and getScalarSecSol should now be placed in the C-file instead.

src/Utility/Vec3.h
akva2: then i need to change all the X = Xnod * N; in integration methods.

akva2:
updated. patching handled inside function, implemented all function types and completed anasol parsing. made everything use .assign() and dropped the added operators.

src/ASM/ASMs2Dmx.C
```diff
@@ -638,7 +638,7 @@ bool ASMs2Dmx::integrate (Integrand& integrand,
               utl::getGmat(Jac,dXidu,fe.G);
 
             // Cartesian coordinates of current integration point
-            X = Xnod * fe.basis(geoBasis);
+            X.assign(Xnod * fe.basis(geoBasis));
```
kmokstad: There is no point in switching to using the assign method until you actually assign the X.u parameters.
So maybe delay touching this file until mixed is fully supported.

src/ASM/ASMs3D.C
```diff
@@ -1737,7 +1737,8 @@ bool ASMs3D::integrate (Integrand& integrand,
       Matrix   dNdu, Xnod, Jac;
       Matrix3D d2Ndu2, Hess;
       double   dXidu[3];
-      Vec4     X;
+      double   param[3] = { 0.0, 0.0, 0.0 };
```
kmokstad: In 3D there is no need to actually initialize the three param values, since all of them will be assigned anyway before use.

src/Utility/AnaSol.C
```diff
@@ -107,8 +113,62 @@ template <class T> void parseDerivatives (T* func, const std::string& variables,
...
+  if (vecSecSol) {
+    TensorFieldFunction* ff = dynamic_cast<TensorFieldFunction*>(vecSecSol);
+    if (ff)
+      ff->initPatch(pIdx);
+  }
+  if (stressSol) {
+    STensorFieldFunction* ff = dynamic_cast<STensorFieldFunction*>(stressSol);
+    if (ff)
+      stressSol->initPatch(pIdx);
+  }
```
kmokstad: Wouldn't it be nicer to have a `virtual void initPatch(size_t) {}` in the `FunctionBase` class instead, such that we don't need to to all this dynamic_cast-ing?

src/Utility/AnaSol.C
```diff
@@ -107,8 +113,62 @@ template <class T> void parseDerivatives (T* func, const std::string& variables,
...
   : vecSol(nullptr), vecSecSol(nullptr), stressSol(nullptr)
 {
-  std::string variables;
+  const char* type = elem->Attribute("type");
+  if (type && !strcasecmp(type,"fields")) {
+    parseFieldFunctions(elem, scalarSol);
+    return;
+  }
+
+  parseExpressionFunctions(elem, scalarSol);
```
kmokstad: Nicer with following?
```
if (type && !strcasecmp(type,"fields"))
  this->parseFieldFunctions(elem,scalarSol);
else
  this->parseExpressionFunctions(elem,scalarSol);
```

src/ASM/ASMs2D.C
```diff
@@ -1907,7 +1909,7 @@ bool ASMs2D::integrate (Integrand& integrand,
 #endif
 
           // Cartesian coordinates of current integration point
-          X = Xnod * fe.N;
+          X.assign(Xnod * fe.N);
```
kmokstad: In this method, where the integration point parameters are provided through an array, the only change needed is to add `X.u = elmPts[ip].data()` (like I did for the equivalent ASMu2D method, but forgot for ASMs2D), and we can skip the param temporary and the assign-method.

src/ASM/ASMs3Dmx.C
```diff
@@ -489,7 +489,8 @@ bool ASMs3Dmx::integrate (Integrand& integrand,
       Matrix3D Hess;
       double dXidu[3];
       Matrix Xnod, Jac;
-      Vec4   X;
+      double   param[3] = { 0.0, 0.0, 0.0 };
+      Vec4   X(param);
```
kmokstad: Ah, so it seems you did implement it for mixed 3D. Then I guess the param-stuff was just forgotten for ASMs2Dmx?

src/ASM/LR/ASMu2D.C
```diff
@@ -1270,7 +1271,8 @@ bool ASMu2D::integrate (Integrand& integrand,
 #endif
 
         // Cartesian coordinates of current integration point
-        X = Xnod * fe.N;
+        X.assign(Xnod * fe.N);
+        X.u = elmPts[ip].data();
```
kmokstad: Don't think you need to switch to using the assign-method here (and in lines 1220 and 1115) since I here use `X.u = (pointer)` instead of going via a double[3] temporary.

src/ASM/ASMs3D.C
```diff
@@ -2113,7 +2115,7 @@ bool ASMs3D::integrate (Integrand& integrand,
 #endif
 
           // Cartesian coordinates of current integration point
-          X = Xnod * fe.N;
+          X.assign(Xnod * fe.N);
```
kmokstad: Here too, can use `X.u = itgPts[iel][ip].data()` instead of using param temporary and changing to assign.

src/Utility/FieldFunctions.h
```diff
@@ -27,25 +29,161 @@ class ASMbase;
...
+  virtual ~VecFieldFuncBase();
+
+  std::vector<Fields*> field; //!< Pointer to the vector field to be evaluated
+  std::vector<ASMbase*> patch; //!< Pointer to the patch on which the field is defined
+  size_t pidx; //!< Current patch index
+
+public:
+  //! \brief Set currently active patch.
+  void initPatch(size_t pIdx) { if (pIdx < field.size()) pidx = pIdx; }
+
```
kmokstad: Delete empty line 93

akva2:
jenkins build this please

akva2:
testing

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

src/ASM/ASMs2D.C
akva2: done.

src/ASM/ASMs2Dmx.C
akva2: done.

src/ASM/ASMs3D.C
akva2: done.

src/ASM/ASMs3D.C
akva2: done.

src/ASM/ASMs3Dmx.C
akva2: done.

src/ASM/LR/ASMu2D.C
akva2: done, but 1115 seems a necessity to me.

src/Utility/AnaSol.C
akva2: definitely. refactor artifact..

src/Utility/AnaSol.C
akva2: sure. didn't know which solution you preferred.

src/Utility/FieldFunctions.h
akva2: done.

akva2:
all requested changed done. i will unit tests on the train later today.

src/ASM/LR/ASMu2D.C
kmokstad: yes, that was in another integrate method, my mistake

akva2:
tests added. the only exception is 3D LR - other issues there hinders me from creating hdf5 files with the data. i will revisit once i have fixed those, but i kindly ask that it is not considered a show-stopper.

akva2:

akva2:

akva2:

akva2:
last tests have been added..

akva2:

PR #219, opened 2017-11-09
Minor issues to clear the ground for dual fields and reaction force calculation for static linear elasticity

kmokstad:
OPM/IFEM-Elasticity#39 depends on this one. Apart from that, nothing here should be controversial.

### DIFF
--- a/src/SIM/SIMbase.C
+++ b/src/SIM/SIMbase.C
@@ -449,7 +449,7 @@ bool SIMbase::initSystem (int mType, size_t nMats, size_t nVec, size_t nScl,
   // Workaround SuperLU bug for tiny systems
   if (mType == SystemMatrix::SPARSE && this->getNoElms(true) < 3)
   {
-    std::cerr <<" ** System too small for SuperLU, falling back to Dense."
+    std::cerr <<"  ** System too small for SuperLU, falling back to Dense."
               << std::endl;
     mType = SystemMatrix::DENSE;
   }
@@ -462,9 +462,7 @@ bool SIMbase::initSystem (int mType, size_t nMats, size_t nVec, size_t nScl,
 
 bool SIMbase::setAssociatedRHS (size_t iMat, size_t iVec)
 {
-  if (!myEqSys) return false;
-
-  return myEqSys->setAssociatedVector(iMat,iVec);
+  return myEqSys ? myEqSys->setAssociatedVector(iMat,iVec) : false;
 }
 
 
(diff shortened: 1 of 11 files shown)


kmokstad:
jenkins build this with downstreams ifem-elasticity=39 please

src/SIM/SIMbase.C
```diff
@@ -1028,11 +1034,11 @@ bool SIMbase::solveSystem (Vector& solution, int printSol, double* rCond,
...
                                  const char* compName)
 {
-  solution.resize(myEqSys->getNoRHS());
+  solution.resize(this->getNoRHS());
   for (size_t i = 0; i < solution.size(); i++)
     if (!this->solveSystem(solution[i],printSol,nullptr,compName,i==0,i))
       return false;
-    else
+    else if (solution.size() > 2)
       printSol = 0; // Print summary only for the first solution
```
akva2: comment seems wrong.

akva2:
and i don't really understand the first fix, but i trust it's doing something.

akva2:
you will get a failure in elasticity after the LR merge. those pr's updating the regtests was stale it seems (same thing happens in poisson..)

kmokstad:
I know, probably I screwed up something when rebasing Kjetil's PRon the LR things. Thought it should be equivalent code, but apparently not :(

Regarding the first commit, it is only (afaik) affects the console printout for problems with multiple material regions. It then printed ca code value for the material that did not exist (typically 10000), and it caused be some confusion and lost time debugging ghosts. This fixes such that the value printed is the value stored in the Property container.

akva2:
personally i think it's just as likely the PR's were stale. i will check.

akva2:
yup. stale pr's, you didn't screw up anything.

akva2:
since you had some other things in the queue that affects those tests, do the update in elasticity with that or the downstream of this ?

kmokstad:
I know that the exact_p[12] elasticity fails, due to the inclusion of the external energy from the reaction forces (+ the adaptive failures, which then is not due to me (phew)). But I just wanted to see if there were other downstream issues, therefore this Jenkins job. I will fix the exact reg-test when I'm sure the new values are the correct ones (not on the top of priority just now though).

akva2:
then i will fix the disc-stair one - i need trustworthy CI.

src/SIM/SIMbase.C
kmokstad: Indeed, now it reads "first two solutions" (the second one being the upcoming dual solution)

kmokstad:
jenkins build this with downstreams ifem-poisson#20 ifem-elasticity=41 ifem-darcy#10 please

kmokstad:
jenkins build this with downstreams ifem-poisson=20 ifem-elasticity=41 ifem-darcy=10 please

kmokstad:
jenkins build this with downstreams ifem-poisson=20 ifem-elasticity=41 ifem-darcy=10 please

PR #224, opened 2017-11-15
changed: get rid of the hardcoded maximum number of norms

akva2:
we have nearly 40 norms in boussinesq..

### DIFF
--- a/src/SIM/SIMoutput.C
+++ b/src/SIM/SIMoutput.C
@@ -1016,8 +1016,7 @@ bool SIMoutput::writeGlvN (const Matrix& norms, int iStep, int& nBlock,
   NormBase* norm = myProblem->getNormIntegrand(mySol);
 
   Matrix field;
-  std::array<IntVec,30> sID;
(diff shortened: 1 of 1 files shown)


src/SIM/SIMoutput.C
```diff
@@ -1016,8 +1016,8 @@ bool SIMoutput::writeGlvN (const Matrix& norms, int iStep, int& nBlock,
   NormBase* norm = myProblem->getNormIntegrand(mySol);
 
   Matrix field;
-  std::array<IntVec,30> sID;
-  const size_t maxN = sID.size();
+  std::vector<IntVec> sID;
```
kmokstad: A more elegant solution would be just
`std::vector<IntVec> sID(norms.rows());`
which always will be big enough since `norms.rows()` will be equal to `field.rows()` if you see what extractElmRes does. Then most of the other changes (except deleting `maxN`) can be removed.

akva2:
jenkins build this with downstreams ifem-openfrac=19 please

PR #228, opened 2017-11-21
added: close/open file in hasGeometries

akva2:
probably the reason for the issues you saw @kmokstad


kmokstad:
Now included in #227. Have verified that this resolves the issue mentioned there.

PR #231, opened 2017-12-08
fixed: truebeta refinement was broken

akva2:
need to remap to elements, not basis functions

### DIFF
--- a/src/ASM/ASMunstruct.h
+++ b/src/ASM/ASMunstruct.h
@@ -168,7 +168,9 @@ public:
   //! \brief Remaps element-wise errors from geometry mesh to refinement mesh.
   //! \param[out] errors The remapped errors
   //! \param[in] orig The element-wise errors on the geometry mesh
-  virtual void remapErrors(RealArray& errors, const RealArray& orig) const = 0;
+  //! \param[in] elemErrors If true, map to elements instead of basis functions
+  virtual void remapErrors(RealArray& errors, const RealArray& orig,
+                           bool elemErrors = false) const = 0;
 
   //! \brief Match neighbours after refinement in multipatch models.
   //! \param neigh Neigbouring patch
--- a/src/ASM/LR/ASMu2D.C
+++ b/src/ASM/LR/ASMu2D.C
@@ -2180,9 +2180,15 @@ bool ASMu2D::matchNeighbour (ASMunstruct* neigh, int midx, int sidx, int orient)
 }
 
 
-void ASMu2D::remapErrors (RealArray& errors, const RealArray& origErr) const
+void ASMu2D::remapErrors (RealArray& errors,
+                          const RealArray& origErr, bool elemErrors) const
 {
   const LR::LRSplineSurface* basis = this->getBasis(1);
+
+  if (elemErrors) {
+    errors = origErr;
+    return;
+  }
 
   for (const LR::Element* elm : basis->getAllElements())
     for (const LR::Basisfunction* b : elm->support())
--- a/src/ASM/LR/ASMu3D.C
+++ b/src/ASM/LR/ASMu3D.C
@@ -2250,9 +2250,15 @@ bool ASMu3D::matchNeighbour(ASMunstruct* neigh, int midx, int sidx, int orient)
 }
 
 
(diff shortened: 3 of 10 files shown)


akva2:
jenkins build this with downstreams ifem-stokes=79 ifem-openfrac=19 please

src/ASM/ASMunstruct.h
```diff
@@ -168,7 +168,9 @@ class ASMunstruct : public ASMbase
   //! \brief Remaps element-wise errors from geometry mesh to refinement mesh.
   //! \param[out] errors The remapped errors
   //! \param[in] orig The element-wise errors on the geometry mesh
-  virtual void remapErrors(RealArray& errors, const RealArray& orig) const = 0;
+  //! \param[in] elemErrors If true, map to elements instead of basis functions
+  virtual void remapErrors(RealArray& errors, const RealArray& orig,
+                           bool elemErrors) const = 0;
```
kmokstad: Maybe with default value = false such that the method might be invoked as before without change in behaviour

src/ASM/LR/ASMu2D.C
```diff
@@ -2180,13 +2180,17 @@ bool ASMu2D::matchNeighbour (ASMunstruct* neigh, int midx, int sidx, int orient)
...
 
-  for (const LR::Element* elm : basis->getAllElements())
-    for (const LR::Basisfunction* b : elm->support())
-      errors[b->getId()] += origErr[elm->getId()];
+  if (elemErrors)
+    errors = origErr;
+  else
+    for (const LR::Element* elm : basis->getAllElements())
+      for (const LR::Basisfunction* b : elm->support())
+        errors[b->getId()] += origErr[elm->getId()];
```
kmokstad: Better with
```
if (elemErrors) {
  errors = origErr;
  return;
}
```
before line 2186 and the rest can be unchanged (less diff)

src/ASM/LR/ASMu3D.C
```diff
@@ -2250,13 +2250,17 @@ bool ASMu3D::matchNeighbour(ASMunstruct* neigh, int midx, int sidx, int orient)
...
 
-  for (const LR::Element* elm : basis->getAllElements())
-    for (const LR::Basisfunction* b : elm->support())
-      errors[b->getId()] += origErr[elm->getId()];
+  if (elemErrors)
+    errors = origErr;
+  else
+    for (const LR::Element* elm : basis->getAllElements())
+      for (const LR::Basisfunction* b : elm->support())
+        errors[b->getId()] += origErr[elm->getId()];
```
kmokstad: same comment here as for the ASMu2D method

src/SIM/AdaptiveSIM.C
```diff
@@ -364,7 +365,7 @@ bool AdaptiveSIM::adaptMesh (int iStep)
 
       // remap from geometry basis to refinement basis
       Vector locErr(patch->getNoNodes(rBasis));
-      static_cast<ASMunstruct*>(patch)->remapErrors(locErr, locNorm);
+      static_cast<ASMunstruct*>(patch)->remapErrors(locErr, locNorm, false);
```
kmokstad: not needed, if the elemeErrors argument has default value = false, as suggested

akva2:
all suggestions taken, merging.

src/ASM/LR/ASMu2D.C
```diff
@@ -2180,10 +2180,16 @@ bool ASMu2D::matchNeighbour (ASMunstruct* neigh, int midx, int sidx, int orient)
...
+void ASMu2D::remapErrors (RealArray& errors,
+                          const RealArray& origErr, bool elemErrors) const
 {
   const LR::LRSplineSurface* basis = this->getBasis(1);
 
+  if (elemErrors) {
+    errors = origErr;
+    return;
+  }
+
```
kmokstad: I wrote _before_ line 2186 (now the basis variable is unused of elemErrors is true, but never mind. Some other time (also for the ASUm3D)

akva2: woops. missed that bit. you are ofc right.

PR #234, opened 2017-12-12
Changed: Suppress result point output for big groups (lines)

kmokstad:
Triviality. Just tired of the long output of the result lines after each mesh refinement..

### DIFF
--- a/src/SIM/SIMoutput.C
+++ b/src/SIM/SIMoutput.C
@@ -216,25 +216,32 @@ void SIMoutput::preprocessResPtGroup (std::string& ptFile, ResPointVec& points)
     else if ((p->inod = pch->evalPoint(p->u,p->u,p->X)) < 0)
       p = points.erase(p);
     else
+      (p++)->npar = pch->getNoParamDim();
+  }
+
+  if (points.size() < 5 || msgLevel > 1)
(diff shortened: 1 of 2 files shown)


akva2:
jenkins build this with downstreams ifem-openfrac=19 please

src/SIM/SIMoutput.C
```diff
@@ -216,8 +216,12 @@ void SIMoutput::preprocessResPtGroup (std::string& ptFile, ResPointVec& points)
     else if ((p->inod = pch->evalPoint(p->u,p->u,p->X)) < 0)
       p = points.erase(p);
     else
+      (p++)->npar = pch->getNoParamDim();
+  }
+
+  if (points.size() < 5 || msgLevel > 1)
+    for (ResPointVec::const_iterator p = points.begin(); p != points.end(); ++p)
```
akva2: range based for while at it ? unless there is some iterator usage hiding in lines not shown..

kmokstad: Not really, except for then have to change all `->` into `.` Can do that of course

kmokstad:
jenkins build this with downstreams ifem-openfrac=19 please

PR #237, opened 2018-01-02
added: unit tests for gaussian quadratures

akva2:
It seems I do not understand the triangle quadrature


akva2:
removed the tests for the triangle quadrature as i do not yet understand the precision of it etc.

akva2:
jenkins build this please

PR #239, opened 2018-01-04
Trace operator for matrix and matrix3D

kmokstad:
Just some convenience methods. Also (first commit) an extra argument for the SIMbase::project method such that it can be used when we only want a specified component.

### DIFF
--- a/src/LinAlg/matrix.h
+++ b/src/LinAlg/matrix.h
@@ -555,6 +555,9 @@ namespace utl //! General utility classes and functions.
       return *this;
     }
 
+    //! \brief Return the trace of the matrix (sum of its diagonal elements).
+    T trace() const { return this->elem.sum(0,nrow+1); }
+
 #define THIS(i,j) this->operator()(i,j)
 
     //! \brief Compute the determinant of a square matrix.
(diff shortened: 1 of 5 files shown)


kmokstad:
jenkins build this with downstreams ifem-poisson=23 please

src/LinAlg/matrix.h
```diff
@@ -916,6 +919,12 @@ namespace utl //! General utility classes and functions.
       return col;
     }
 
+    //! \brief Return the trace of the \a i1'th sub-matrix.
+    T trace(size_t i1) const
+    {
+      return this->elem.sum(i1-1,this->n[0]*(this->n[1]+1));
```
akva2: a index check is possibly appropriate.

kmokstad:
jenkins build this with downstreams ifem-poisson=23 please

src/LinAlg/matrix.h
kmokstad: Yes, indeed. Added.

kmokstad:
jenkins build this with downstreams ifem-poisson=23 please

PR #240, opened 2018-01-05
fixed: ASMbase needs the maximum number of components

akva2:
if not, we cannot apply boundary conditions for simulators where
number of fields > nsd. fixes boussinesq application.


akva2:
3D doesn't have the n_s/n_f param separation. the other constructors are cc's, and/or one where you can manually set n_f, and LR has been fixed earlier. there are other issues in 3D but that's unrelated to this.

PR #241, opened 2018-01-09
fixed: this should not be virtual

akva2:
quells warnings about deleting object with virtual members and
non-virtual destructor


akva2:
jenkins build this please

PR #243, opened 2018-01-11
fixed: iterate over basis functions instead of getting by ID

akva2:
the data structure is O(n) for access (and O(1) insert). 

@kmokstad should fix observed inefficiencies in openfrac + hdf5 export.


VikingScientist:

PR #244, opened 2018-01-13
Profiling of DataExporter::dumpTimeLevel + 3D fix

kmokstad:
You did not give me a chance to comment on #243 before it was merged, as I did not check sintef-email yesterday. Therefore a new PR here with some amendments. Also check if the same time thief is hidden elsewhere.


akva2:
yes kjetil was a bit quicker on the trigger than anticipated. looks all good to me. there is no additional usage where we loop over all function afaict, so the rest of the calls to getBasisfunction are necessary.

kmokstad:

PR #245, opened 2018-01-19
added: specify parameter domain in model generator

akva2:


src/SIM/ModelGenerator.C
```diff
@@ -21,6 +21,32 @@
...
+    print |= utl::getAttribute(geo,min.c_str(),pmin);
+    print |= utl::getAttribute(geo,max.c_str(),pmax);
+    if (print)
+      IFEM::cout <<"\n\t" << min[0] << "["<< pmin << ","<< pmax <<"]";
+    std::stringstream str;
+    str << pmin << " " << pmin << " " << pmax << " " << pmax;
+    g2.append("\n2 2\n");
+    g2.append(str.str());
+  }
+  g2.append("\n");
```
kmokstad: Alternative rewrite (for compactness, but not tested)
```
  char* min = const_cast<char*>("umin");
  char* max = const_cast<char*>("umax");
  std::stringstream str;
  for (size_t i = 0; i < num; ++i, ++min[0], ++max[0]) {
    double pmin = 0.0, pmax = 1.0;
    if (utl::getAttribute(geo,min,pmin) | utl::getAttribute(geo,max,pmax))
      IFEM::cout <<"\n\t"<< min[0] <<"["<< pmin <<","<< pmax <<"]";
    str <<"\n2 2\n"<< pmin <<" "<< pmin <<" "<< pmax <<" "<< pmax;
  }
  g2.append(str.str());
  g2.append("\n");
```

akva2: took most of it, but i cannot do those char* shenanigans. cdata is in read only memory, it instantly segfaults.

PR #246, opened 2018-01-31
Some small fixes

akva2:
- allow parallel test execution
- quell some clang warnings


akva2:
i can do the long rationalization in real life, but it basically comes down to http://en.cppreference.com/w/cpp/language/list_initialization. the first { is the initializer list, then second { is the values for the array.

akva2:
jenkins build this please

PR #248, opened 2018-02-05
Doxy issues and don't use utl::vector when not needed

kmokstad:
Pull this one before #246


akva2:
jenkins build this please

PR #249, opened 2018-02-05
fixed: make sure we have a row to store anasol component in

akva2:
This is necessary with Chorin. Due to the split nature of things, we do not have a pressure in the primary field when storing the VTF (it's written by the pressure correction SIM).

This causes problems, as the anasol still holds the pressure (necessary for norm calculations) and this sanity check comes to the rescue. Should only add sanity in any case so i think it's an acceptable solution.


akva2:
it's a bit tricky to improve resizing which is why it was done as-is i believe. the assumption is that anasol holds as many components as the primary solution and that the components are named thereafter. which means that the only mismatches we can really handle is truncated anasol or too many components in anasol. the former is already handled, this handles the latter.

PR #250, opened 2018-02-08
Move patch-only dependent code from SIMinput to ASM-scope

kmokstad:
Some refactoring to make SIMinput cleaner.

There are two issues with this one that need some attention before this can be merged.

1. In ASMu2D::InterfaceChecker::Intersection (which I now changed into a std::pair, was done before the doxy arrived), I cannot see that the continuity flag (now the first member) is ever set or used. It is referred in the Stokes-app (the JumpCheckeru2D class), but where is it defined?

2. The second commit (SIM3D::clonePatches was missing) causes the 3D FSI tests to fail, due to the results are slightly off. Need to check whether it is the new or the old results that are the "correct" ones, and either update the tests for find the source of the error.

### DIFF
--- a/src/ASM/ASMunstruct.h
+++ b/src/ASM/ASMunstruct.h
@@ -33,13 +33,12 @@ namespace LR //! Utilities for LR-splines.
     options[0] is the beta percentage of elements to refine,
     options[1] is the knotline multiplicity (default 1),
     options[2] is the refinement scheme (default 0),
-    (FULLSPAN=0, MINSPAN=1, ISOTROPIC ELEMENTS=2, ISOTROPIC FUNCTIONS=3),
-    options[3] is the symmetry, i.e., always refine a multiple of this value,
-    options[4] is nonzero if testing for linear independence at all iterations,
-    options[5] is the maximum number of T-joints allowed in the model,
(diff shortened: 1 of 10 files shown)


kmokstad:
OK, as I said, it was changed into a pair because the previous struct was undocumented, and to minimize the needed doxy..., but since that has been changed now I can revert that part. But still it is not good with an uninitialized member (the continuity)

akva2:
agree. initialize it to 0.

kmokstad:
jenkins build this with downstreams ifem-fsi=26 please

kmokstad:
Regarding the second issue, I think the new response is the correct one. The difference was due to that the FSI solver thought the chorin sub-solvers were sharing the grids, but they were actually not. The result was then that only the velocity prediction solver got its grid updated when moving the structure, whereas the correction solvers worked on the initial grids all the way.

kmokstad:
Det ble failure her. Men ser ikke noe annet galt enn de doxy-warningene?

akva2:
ja, bare merge, hickup i doxygen scanneren (ser på det nå).

akva2:
jenkins build this with downstreams ifem-fsi=26 please

PR #251, opened 2018-02-08
added: store continuity of mesh line intersections in ASMu2D

akva2:
As pointed out in #250 this was unimplemented.


akva2:
jenkins build this please

PR #257, opened 2018-04-21
Nonlinear shell support

kmokstad:
This adds stuff needed when doing nonlinear Kirchhoff-Love shells.

### DIFF
--- a/src/ASM/ASMs2D.C
+++ b/src/ASM/ASMs2D.C
@@ -1649,6 +1649,15 @@ bool ASMs2D::integrate (Integrand& integrand,
           break;
         }
 
+        if (integrand.getIntegrandType() & Integrand::UPDATED_NODES)
+          if (!time.first || time.it > 0)
+            if (!this->deformedConfig(Xnod,A->vec))
+            {
+              A->destruct();
+              ok = false;
+              break;
+            }
+
         if (xr)
         {
           // --- Selective reduced integration loop ----------------------------
@@ -1883,6 +1892,15 @@ bool ASMs2D::integrate (Integrand& integrand,
           ok = false;
           break;
         }
+
+        if (integrand.getIntegrandType() & Integrand::UPDATED_NODES)
+          if (!time.first || time.it > 0)
+            if (!this->deformedConfig(Xnod,A->vec))
+            {
+              A->destruct();
+              ok = false;
+              break;
+            }
 
 
         // --- Integration loop over all quadrature points in this element -----
@@ -2630,6 +2648,12 @@ bool ASMs2D::evalSolution (Matrix& sField, const IntegrandBase& integrand,
   // Fetch nodal (control point) coordinates
   Matrix Xnod, Xtmp;
   this->getNodalCoordinates(Xnod);
+  if (integrand.getIntegrandType() & Integrand::UPDATED_NODES)
+  {
+    Vectors& eV = const_cast<IntegrandBase&>(integrand).getSolutions();
+    if (!this->deformedConfig(Xnod,eV,true))
+      return false;
+  }
 
   FiniteElement fe(p1*p2,firstIp);
   fe.p = p1 - 1;
--- a/src/ASM/Integrand.h
+++ b/src/ASM/Integrand.h
@@ -146,14 +146,15 @@ public:
     STANDARD           = 0, //!< Default integrand type (first derivatives only)
     NO_DERIVATIVES     = 1, //!< Integrand don't want any derivatives
     SECOND_DERIVATIVES = 2, //!< Integrand wants second derivatives
-    AVERAGE            = 4, //!< Integrand wants basis function averages
-    ELEMENT_CORNERS    = 8, //!< Integrand wants element corner coordinates
-    ELEMENT_CENTER    = 16, //!< Integrand wants element center coordinates
-    G_MATRIX          = 32, //!< Integrand wants the G matrix
-    NODAL_ROTATIONS   = 64, //!< Integrand wants nodal rotation tensors
-    XO_ELEMENTS      = 128, //!< Integrand is defined on extraordinary elements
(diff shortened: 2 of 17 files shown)


kmokstad:
The conflict in the CMakeLists.txt file is not resolved, since the new version does not work with my setup.
It is very unfortunate to remove the support for the old file structure (without the IFEM-prefices). What is the rationale behind this change? I thought it worked before also with either structure.
And, the spacing in the new file is totally off (yes, there should have been a PR before it was pushed).

akva2:
Wooops. No rationale, it was not intended to be pushed as is. I was stressing with poster and stuff and didnt pay attention to what sat in my master branch it seems.

kmokstad:
jenkins build this with downstreams ifem-elasticity=45 please

kmokstad:
jenkins build this with downstreams ifem-elasticity=45 please

src/LinAlg/Test/TestMatrix.C
```diff
@@ -122,3 +122,22 @@ TEST(TestMatrix3D, DumpRead)
...
+  utl::matrix3d<double> B(5,4,3), C, D;
+
+  std::iota(a.begin(),a.end(),1.0);
+  std::iota(A.begin(),A.end(),1.0);
+  std::iota(B.begin(),B.end(),1.0);
+
+  C.multiply(A,B);
+  ASSERT_TRUE(D.multiplyMat(a,B));
+
+  std::vector<double>::iterator c = C.begin();
```
akva2: const_iterator

src/LinAlg/matrix.h
```diff
@@ -971,6 +995,23 @@ namespace utl //! General utility classes and functions.
...
 
+    //! \brief Check dimension compatibility for matrix-matrix multiplication,
+    //! when the matrix A is represented by a one-dimensional vector.
+    bool compatible(const std::vector<T>& A, const matrix3d<T>& B,
+                    size_t& M, size_t& N, size_t& K)
+    {
+      N = B.n[1]*B.n[2];
+      K = B.n[0];
+      M = K > 0 ? A.size() / K : 0;
+      if (M*K == A.size() && !A.empty()) return true;
```
akva2: reverse if

src/ASM/ASMs2D.C
```diff
@@ -2630,6 +2648,12 @@ bool ASMs2D::evalSolution (Matrix& sField, const IntegrandBase& integrand,
   // Fetch nodal (control point) coordinates
   Matrix Xnod, Xtmp;
   this->getNodalCoordinates(Xnod);
+  if (integrand.getIntegrandType() & Integrand::UPDATED_NODES)
+  {
+    Vectors& eV = const_cast<IntegrandBase&>(integrand).getSolutions();
```
akva2: isn't the const_cast a bit .. icky ?

src/ASM/Integrand.h
```diff
@@ -153,7 +153,8 @@ class Integrand
     NODAL_ROTATIONS   = 64, //!< Integrand wants nodal rotation tensors
     XO_ELEMENTS      = 128, //!< Integrand is defined on extraordinary elements
     INTERFACE_TERMS  = 256, //!< Integrand has element interface terms
-    NORMAL_DERIVS    = 512  //!< Integrand p'th order normal derivatives
+    NORMAL_DERIVS    = 512, //!< Integrand p'th order normal derivatives
+    UPDATED_NODES    = 1024 //!< Integrand wants updated nodal coordinates
```
akva2: for later, but we should change this to '1 << n' at some point imo.

src/SIM/SIMoutput.C
```diff
@@ -517,45 +517,42 @@ int SIMoutput::writeGlvS1 (const Vector& psol, int iStep, int& nBlock,
...
+  std::vector<IntVec> sID;
+  sID.reserve(haveXsol ? nf+nf : nf);
+
   Matrix field;
   Vector lovec;
 
-  size_t i, j, k;
   int geomID = myGeomID;
-  for (i = 0; i < myModel.size(); i++)
+  for (ASMbase* pch : myModel)
```
akva2: can this be const ?

src/LinAlg/Test/TestMatrix.C
kmokstad: fixed

src/LinAlg/matrix.h
kmokstad: not 100% sure, but I assume you here mean I shall check A.empty() first? Can change it.
Don't think it matters though, except avoiding one multiplication in the case A _is_ empty.
And btw, all the other compatible methods evaluates the if statement in this order, then they should be changed also, for consistency.

src/ASM/Integrand.h
kmokstad: Instead of the enums? Not sure if I got it, but I kind of like the readability of the enum names instead of a 'cryptical' integer value, or is it possible to use << expressions in the enum-declaration above?

src/SIM/SIMoutput.C
kmokstad: I guess it can, since only const member functions are invoked for the pointed-to object.
Changed.

src/ASM/ASMs2D.C
kmokstad: it is indeed, but I did not see any better solution, since the integrand argument is const

kmokstad: The eV array is updated by the ASMbase::deformedConfig method, so a const overload of getSolutions will not work. Think therefore we leave it like it is, at least for now

src/ASM/Integrand.h
kmokstad: Changed to 1<<n, for n=2,3,4,... seems to work fine

kmokstad:
jenkins build this with downstreams ifem-elasticity=45 please

PR #258, opened 2018-04-22
Added: virtual method for app-specific load/time step heading

kmokstad:
I override the new method to print out the prescribed crack pressure in OpenFrac, for instance.

In the other commit I tried to account for the two different file structures that might be common under Apps. Guess it makes sense to support the IFEM-... structure, as these are the names used in the repositories. The README.md changes is only to restore the line wrapping. Having everything in one big line makes it hard to spot changes in later revisions using diff.


Apps/CMakeLists.txt
```diff
@@ -21,35 +21,64 @@ set(EXECUTABLE_OUTPUT_PATH ${CMAKE_BINARY_DIR}/bin)
...
-          add_subdirectory(IFEM-ThermoElasticity)
+    add_subdirectory(IFEM-ThermoElasticity)
   endif()
   if(EXISTS ${PROJECT_SOURCE_DIR}/IFEM-PoroElasticity)
     add_subdirectory(IFEM-PoroElasticity)
   endif()
   if(EXISTS ${PROJECT_SOURCE_DIR}/IFEM-OpenFrac)
     add_subdirectory(IFEM-OpenFrac)
   endif()
+elseif(EXISTS ${PROJECT_SOURCE_DIR}/Elasticity)
```
akva2: this can be deduplicated by setting some var elasticity_root var first.

README.md
```diff
@@ -1,8 +1,20 @@
...
-IFEM is an object-oriented toolbox for implementing isogeometric finite element solvers for linear and nonlinear partial differential equations. The main toolbox together with structural mechanics applications was developed through the ICADA project at SINTEF Digital in collaboration with NTNU. The important extension to fluid mechanics and coupled problems was done during the NOWITECH and FSI-WT project. The purpose of this code is to serve as a common base for isogeometric PDE-simulators, using splines and NURBS as basis functions in the finite element formulations. The toolbox contains methods for doing linear and non-linear, stationary and dynamic time-domain analyses, as well as eigenvalue analyses. IFEM is parallelized using the Petsc library and enables adaptive refinement using LR B-splines based on residual or recovery methods for a posteriori error estimation.
+
+IFEM is an object-oriented toolbox for implementing isogeometric finite element
+solvers for linear and nonlinear partial differential equations.
+The main toolbox together with structural mechanics applications was developed
+through the ICADA project at SINTEF Digital in collaboration with NTNU.20a1ff8e
```
akva2: hash at end.

Apps/CMakeLists.txt
kmokstad: Tried something like that in the last fixup commit. Not sure if this is the best way but think it works?

kmokstad:

Apps/CMakeLists.txt
```diff
@@ -37,72 +40,90 @@ elseif(EXISTS ${PROJECT_SOURCE_DIR}/Darcy)
...
-    add_subdirectory(Elasticity/FiniteDeformation/Nonlinear)
-  endif()
-  if(EXISTS ${PROJECT_SOURCE_DIR}/ThermoElasticity)
-    add_subdirectory(ThermoElasticity)
-  endif()
-  if(EXISTS ${PROJECT_SOURCE_DIR}/PoroElasticity)
-    add_subdirectory(PoroElasticity)
-  endif()
-  if(EXISTS ${PROJECT_SOURCE_DIR}/FractureDynamics)
+  elsif(EXISTS ${PROJECT_SOURCE_DIR}/FractureDynamics)
```
akva2: elseif

PR #259, opened 2018-04-29
Improved support for time-dependent discrete loads

kmokstad:
This includes support for (time)derivatives of ScalarFunc objects and accounting to time-variation when calculating the external energy contributions. This is needed when we are going to do nonlinear shell and beam problems with varying point loads.

This partly supersedes one commit in #257 which will be rebased after this one.


kmokstad:
jenkins build this with downstreams ifem-elasticity=46 please

src/Utility/ExprFunctions.C
```diff
@@ -104,50 +106,59 @@ EvalFunc::EvalFunc (const char* function, const char* x)
...
+  for (ExprEval::Expression* it : expr)
     delete it;
-  for (auto& it : f)
+  for (ExprEval::FunctionList* it : f)
     delete it;
-  for (auto& it : v)
+  for (ExprEval::ValueList* it : v)
     delete it;
-  arg.clear();
+  if (gradient)
```
akva2: delete is null safe.

src/Utility/ExprFunctions.C
```diff
@@ -205,26 +226,26 @@ EvalFunction::EvalFunction (const char* function) : gradient{}, dgradient{}
...
+  for (ExprEval::Expression* it : expr)
     delete it;
-  for (auto& it : f)
+  for (ExprEval::FunctionList* it : f)
     delete it;
-  for (auto& it : v)
+  for (ExprEval::ValueList* it : v)
     delete it;
-  for (auto& it : gradient)
+  for (EvalFunction* it : gradient)
```
akva2: clear these as well i guess.

src/Utility/ExprFunctions.h
```diff
@@ -42,23 +42,36 @@ class EvalFunc : public ScalarFunc
...
+
+  //! \brief Returns the first-derivative of the function.
+  virtual Real deriv(Real x) const;
 
 protected:
   //! \brief Non-implemented copy constructor to disallow copying.
   EvalFunc(const EvalFunc&);
-  //! \brief Non-implemented assigment operator to disallow copying.
+  //! \brief Non-implemented assignment operator to disallow copying.
   EvalFunc& operator=(const EvalFunc&);
```
akva2: c++11 has a useful construct - = delete. yields more understandable error messages should this be attempted to be used.

src/Utility/ExprFunctions.C
kmokstad: OK, removed the if ()

src/Utility/ExprFunctions.C
kmokstad: not sure about that. gradient (and dgradient) are std::array's and not vectors. Then clear is not defined. But do we need to nullify the array members (pointers) or is that automatically done through the delete operator.

src/Utility/ExprFunctions.h
kmokstad: changed.

kmokstad:
jenkins build this with downstreams ifem-elasticity=46 please

src/Utility/ExprFunctions.C
akva2: delete does not nullify, std::fill or for() then.

akva2:
jenkins does not use the Apps/ CMakeLists.txt, so errors there will slip.

src/Utility/ExprFunctions.C
kmokstad: std::array has a fill member, so used that instead.

kmokstad:
jenkins build this with downstreams ifem-elasticity=46 please

Issue #261, opened 2018-05-16
Account for <collapse> when setting up element blocks for multi-threading


kmokstad:
The `<collapse>` keyword is used to collapse an edge into a point or a face into a line when you have a polar parametrization (quarter hemisphere, for example). All control points along that edge/face are then assigned the same node number such that they share the same nodal DOFs at that location. But, depending on the orientation of your parametrization, this may ruin the element blocks for multi-threading, such that blocks that are assembled in parallel may try to update the same entry in the global matrix. This will then yield random and incorrect results locally.

The solution would be to check for whether `<collapse>` has been used, and then orient the blocks accordingly, such that the collapsed edge/face is not spitted into several blocks.

In the meantime, the workaround is to run without multi-threading, using OMP_NUM_THREADS=1 (command)

kmokstad:
Now fixed through #264.

PR #262, opened 2018-05-16
changed: disable multithreading when encountering the collapse keyword

akva2:
since multiple nodes are collapsed to one, the mesh coloring
algorithm fails.

ref https://github.com/OPM/IFEM/issues/261


akva2:
Better to have it always working without user intervention in the mean time imo.

PR #265, opened 2018-06-26
move reference data from external files into source files

akva2:


akva2:
jenkins build this please

PR #270, opened 2018-07-16
LR find closest node

kmokstad:
The first two commits are the same as #267.

The other two to facilitate adaptive analysis of a plate with a center point load.
Since we are not guarantied that the refined mesh will include the control points of the starting mesh, we need to find the closest control point instead to apply the load at (approximation that will converge as the mesh is refined under the load, hopefully).

Besides, there was a bug in the threadGroup generation in that the old groups were not nullified when refining the mesh (why has this not been ans issue before)?

Finally (last commit), introduce option to specify the precision in norm and component output, also when running adaptively.

### DIFF
--- a/Apps/Common/MeshUtils.C
+++ b/Apps/Common/MeshUtils.C
@@ -27,28 +27,25 @@ static bool compute(Vector& result, const SIMbase& model,
 {
   result.resize(model.getNoElms());
 
-  for (int l = 0; l < model.getNoPatches(); l++) {
-    Vector locvec;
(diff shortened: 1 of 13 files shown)


kmokstad:
jenkins build this with downstreams ifem-elasticity=48 ifem-stokes=93 ifem-navierstokes=84 ifem-openfrac=32 please

akva2:
hmm, i guess i haven't noticed the thread group stuff since it's not used with mixed and i have only worked with mixed problems lately.

PR #274, opened 2018-07-24
Fixed VTF output for mixed lagrange

kmokstad:
The important  thing here is that getNoNodes returned rubbish if invoked with basis < 0 (which is the case in the VTF-export). Correct behavior is to return the same as for basis=0.
This fix is needed for sintefmath/IFEM-FiniteDeformation#11 to pass.

The rest here is more or less cosmetics, adding `elem_size` as member in ASMmxBase instead of local variable that has to be recalculated each time it is needed.


kmokstad:
jenkins build this please with downstreams ifem-finitedeformation=11 please

src/ASM/ASMmxBase.h
```diff
@@ -95,7 +93,8 @@ class ASMmxBase
   std::vector<int> MADOF; //!< Matrix of accumulated DOFs for this patch
 
 protected:
-  std::vector<size_t>        nb;  //!< Number of basis functions in each basis
+  std::vector<size_t>  elem_size; //!< Number of elements in each basis
```
akva2: wrong doxy.

PR #276, opened 2018-08-03
fixed: we need to evaluate nb[b] before incrementing b

akva2:
whoopsie is obvious, but i still want a jenkins sanity check before commit..


akva2:

PR #277, opened 2018-08-03
added: support for the dörfel marking strategy

akva2:
mark elements/functions which contributes beta% to the total error

### DIFF
--- a/src/SIM/AdaptiveSIM.C
+++ b/src/SIM/AdaptiveSIM.C
@@ -116,6 +116,8 @@ bool AdaptiveSIM::parse (const TiXmlElement* elem)
         threshold = MINIMUM;
       else if (type.compare("truebeta") == 0)
         threshold = TRUE_BETA;
+      else if (type.compare("dorfel") == 0)
+        threshold = DORFEL;
       IFEM::cout <<"\tRefinement percentage: "<< beta <<" type="<< threshold
                  << std::endl;
     }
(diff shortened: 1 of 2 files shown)


src/SIM/AdaptiveSIM.C
```diff
@@ -418,6 +420,17 @@ bool AdaptiveSIM::adaptMesh (int iStep, std::streamsize outPrec)
...
     break;
+  case DORFEL:
+  {
+    for (const DblIdx& error : errors)
+      sumErr += error.first;
+    double total = 0.0;
+    size_t i = 0;
+    while (total < sumErr*beta/100.0)
+      total += errors[i++].first;
+    limit = errors[i].first;
```
kmokstad: Cosmetic alternative to lines 428-431:
```
sumErr *= beta/100.0;
for (const DblIdx& error : errors)
  if (total < sumErr)
    total += error.first;
  else
  {
    limit = error.first;
    break;
  }
```
and we don't need the brace-pair after the case-statement, if the `total`variable is declared outside together with `limit`and `sumErr`.

akva2: i'll do the rewrite, but if i make total non-local i will change name. it's too easily confused with all sorts of things outside the local context.

kmokstad: sure, looks good :-) But maybe also initialize limit to zero (or maybe `errors.back().first`then is the correct value) before the loop, just to cover the unlikely event of a very big beta value such that curErr never gets above sumErr

akva2: done.

akva2:
test added in https://github.com/OPM/IFEM-Poisson/pull/28

merging.

PR #281, opened 2018-08-16
Efficient norm integration for LR

### LINKED ISSUE #280: Multi-threaded norm evaluation can be made more efficient
The norm evaluations uses the same element groups for multi-threaded assembly as the matrix assembly, where effort is done to establish groups with no overlapping nodes that can be assembled in parallel. However, especially for unstructured LR-spline grids the element groups tend to be highly unbalanced and thereby often causing more overhead than we gain by the multiple threads.

However the norm evaluation only assemble into non-overlapping element arrays, and can therefore be done i parallel and in random error with no danger of conflicting updates. It would therefore be preferable to avoid the groups made for matrix assembly here, and instead distribute the elements evenly over the threads. This is especially important when the norm-integration involve analytical solutions with complex evaluations, like series summation. The norm-integration is then the main bottleneck of the computation as of now.

kmokstad:
This fixes #280 

Still need for the mixed class, and maybe also tensor splines?

### DIFF
--- a/src/ASM/ASMs2D.C
+++ b/src/ASM/ASMs2D.C
@@ -1554,14 +1554,17 @@ bool ASMs2D::integrate (Integrand& integrand,
   const int n1 = surf->numCoefs_u();
   const int nel1 = n1 - p1 + 1;
 
+  ThreadGroups oneGroup;
+  if (glInt.threadSafe()) oneGroup.oneStripe(nel);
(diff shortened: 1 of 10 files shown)


kmokstad:

kmokstad:
Tensor splines now also included (small fix).
But not mixed, add that later in a separate PR when/if needed.

PR #285, opened 2018-08-28
added: support for K-refinement cycles

akva2:
This requires extending PETSc support to allow simulators as matrix-vector products and preconditioners.


akva2:
jenkins build this with downstreams ifem-poisson=29 please

PR #290, opened 2018-09-26
Fixed: Determinant of second-order symmetric tensors

TheBB:

PR #291, opened 2018-09-26
 Fixed: Determinant of 2D symmetric tensor

kmokstad:
Rebased  version of #290 including extended unit test and bugfix in assignment operator


kmokstad:

PR #292, opened 2018-09-28
fixed: add virtual destructors to classes with virtual members

akva2:


akva2:
jenkins build this please

akva2:
No difference as such in this case. Was introduced as syntax sugar in c++11, i prefer it but i dont care either way. It is used to force the generation if a default constructor even if another taking params is defined in certain cases, see https://en.cppreference.com/w/cpp/language/default_constructor

kmokstad:
OK, but here it was applied to the destructor, _not_  the constructor, and as fa as I know, there is not such thing as destructors with params..

akva2:
yes, but the syntax is also allowed for destructors. i like it cause it's explicit that these are generated for us. anyways, i have changed to {}. strictly speaking to have the equivalent i'd have to do virtual ~foo() noexcept {} but as we dont really do exceptions..

akva2:
jenkins build this please

PR #294, opened 2018-10-01
changed: make standard adaptive formulations output first mesh to eps

akva2:


src/ASM/LR/ASMLRSpline.C
```diff
@@ -143,15 +143,40 @@ bool ASMunstruct::refine (const LR::RefineData& prm,
 {
   PROFILE2("ASMunstruct::refine()");
 
+  auto&& storeMesh = [this, fName]()
```
kmokstad: Lambda functions is a nice concept, but in this case I think a protected member function would be a more natural choice?

akva2: Probably, i just did the same as was done for mixed.

akva2:
changed to priv member.

akva2:
jenkins build this please

src/ASM/ASMunstruct.h
```diff
@@ -227,6 +227,11 @@ class ASMunstruct : public ASMbase
 
   static int gEl;  //!< Global element counter
   static int gNod; //!< Global node counter
+
+private:
+  //! \brief Store basis to eps files.
+  //! \param fName Prefix for file names
```
kmokstad: `\param[in]`

src/ASM/LR/ASMLRSpline.C
```diff
@@ -143,15 +143,19 @@ bool ASMunstruct::refine (const LR::RefineData& prm,
 {
   PROFILE2("ASMunstruct::refine()");
 
+
```
kmokstad: Remove extra blank line

PR #295, opened 2018-10-01
Support output of condition numbers in newmark solver

TheBB:
This adds an option for `<printCond />` in NewmarkSIM with the same implementation that currently exists in NonlinSIM.


akva2:
jenkins build this please

PR #298, opened 2018-10-08
Added: virtual method ASMbase::getParameterDomain

kmokstad:
This method can be used both in #161 and by SIMElasticity in when checking that the parameterdomain is valid for the Image material stuff.

Also the stb_image.h header file is moved to the 3rdparty folder, where all such code that we use in our projects should be placed, to keep it tidy.

If you agree..


kmokstad:
jenkins build this with downstreams ifem-elasticity=62 please

PR #305, opened 2018-10-30
fixed: due to RAII usage in libcereal archive must go out of scope

akva2:
if not, the stream is not finished when we read out the data

### DIFF
--- a/src/SIM/SIMsolution.C
+++ b/src/SIM/SIMsolution.C
@@ -46,9 +46,11 @@ bool SIMsolution::saveSolution (SerializeMap& data,
 {
 #ifdef HAS_CEREAL
   std::ostringstream str;
-  cereal::BinaryOutputArchive archive(str);
-  for (const Vector& sol : this->getSolutions())
-    archive(sol);
+  {
+    cereal::BinaryOutputArchive archive(str);
+    for (const Vector& sol : this->getSolutions())
+      archive(sol);
+  }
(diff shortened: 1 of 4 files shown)


src/SIM/SIMsolution.C
```diff
@@ -46,9 +46,11 @@ bool SIMsolution::saveSolution (SerializeMap& data,
...
 #ifdef HAS_CEREAL
   std::ostringstream str;
-  cereal::BinaryOutputArchive archive(str);
-  for (const Vector& sol : this->getSolutions())
-    archive(sol);
+  {
+    cereal::BinaryOutputArchive archive(str);
+    for (const Vector& sol : this->getSolutions())
+      archive(sol);
+  }
```
kmokstad: It does not have an explicit `close`method one can use instead?

akva2: unfortunately no.

akva2: the alternative is new / delete.

akva2:
jenkins build this with downstreams ifem-advectiondiffusion=24 ifem-stokes=105 ifem-boussinesq=4 please

akva2:
jenkins build this with downstreams ifem-advectiondiffusion=24 ifem-stokes=105 ifem-boussinesq=4 please

akva2:
okay to merge ?

PR #306, opened 2018-10-31
Changed: Check app return code in regtest.sh

TheBB:
There seems to be at least three regtests in apps (OpenFrac specifically) that produce the correct output but crash after that point.


TheBB:

TheBB:
Okay, if I understand this correctly, no tests except those in https://github.com/OPM/IFEM-OpenFrac/pull/42 failed due to this change.

PR #310, opened 2018-11-07
fixed: properly serialize timestep class

akva2:


akva2:
jenkins build this with downstreams ifem-openfrac=45 please

akva2:
indeed. simpler fix in here now - the problem was stepIt was not assigned the proper value.

src/SIM/TimeStep.C
```diff
@@ -371,6 +371,7 @@ bool TimeStep::deSerialize (const std::map<std::string,std::string>& data)
     std::stringstream str(it->second);
     cereal::BinaryInputArchive ar(str);
     doSerializeOps(ar,*this);
+    stepIt = mySteps.end();
```
kmokstad: Agree, the clue is to reset stepIt. But this will be correct only if the restart time is within the last `<step>` domain (if more than one). Let say you have
```
  <timestepping>
    <step start="0.0" end="1.0">0.1</step>
    <step start="1.0" end="2.0">0.05</step>
    <step start="2.0" end="3.0">0.001</step>
  </timestepping>
```
and then you do a restart at t=1.5. Will it then be right?
Guess in general you need to search for the correct `stepIt` based on the specified restart time.

PR #317, opened 2018-11-24
Support for third derivatives in 1D and 2D

kmokstad:
Fixed/cleaned/rebased version of #314.
Also had to include the first commit (from last year actually), introducing the NORMS enum.
This will most likely cause some downstream failures until this is accounted for in the other apps.


src/ASM/ASMs2D.C
```diff
@@ -1519,6 +1519,7 @@ bool ASMs2D::integrate (Integrand& integrand,
   PROFILE2("ASMs2D::integrate(I)");
 
   bool use2ndDer = integrand.getIntegrandType() & Integrand::SECOND_DERIVATIVES;
```
akva2: you removed the or here. this mean you won't get hessians with only third derivative flag enabled. i for one would expect second derivatives as well when requesting third.

kmokstad:
The last fixup commit enables the correct Hessian2 calculations for 1D and 2D when `nsd`=`ndim`.
This fixes the 1D case at least (see the Elasticity/Linear/SSbeamPointLoad test).

But the 2D version is still buggy. Think you need the check the `d3Ndu3` --> `d3NdX3` transformation once more. The cross terms `d3NdX3(n,2,1,2)` and `d3NdX3(n,2,2,1)` appear all zero even when the corresponding terms in d3Ndu3 are non-zero and `Ji` is a diagonal matrix. That can't be right?

src/ASM/ASMs2D.C
kmokstad: That's correct. Think it is consistent that the integrand sets both flags if both 2nd and 3rd derivatives are needed, and in the unlikely event that only 3rd (but not 2nd) that can be achieved as well by only setting the 3rd derivative flag

akva2:
It is not consistent. Setting second derivs yields first and second. Not necessarily disagreeing but consistency is not a valid argument.

src/ASM/ASMs2D.C
kmokstad: Providing first derivatives is the default (enum value STANDARD), as that is always needed for any integrand (can switch of that too using NO_DERIVATIVES, which is used in the ElasticBar integrand where all matrices are set up explicitly, but that's a rare exception). Setting either the 2nd or 3rd flag, will provide those derivatives _in addition_ to the 1st. So in that context it is the logic behavior, in my opinion.

akva2: logic != consistent. anyways..

akva2:
uhm, if Ji is diagonal, we will get zero in the cross derivatives, since they involve dxi/dy and deta/dx. i sent a pr with the documents, i can't find any error, checked another time.

kmokstad:
Well, at least something is wrong. Try `LinEl -2DKL NavierPoint_p2.xinp` from the last commit in OPM/IFEM-Elasticity#69 where I've added three more result points. This problem is symmetric about both the vertical and horizontal middle axes and I expected the shear forces (4th and 5th component in the sol2 printout) to reflect that, i.e. qx(0.25,0.5) = qy(0.5,0.25), qy(0.5,0.25) = qx(0.25,0.5), and qx(0.25,0.25) = qy(0.25,0.25). But it turns out qx becomes zero in all those points.

src/ASM/ASMs2D.C
kmokstad: logic, consistent, whatever.. This convention is OK then?

akva2:
running that input file, i do not get to that block at all, it enters the
```
 else if (Ji.cols() <= 2)
{
     // Special treatment for one-parametric elements in multi-dimension space
     // as well as two-parametric elements in 3D space (shells)
     d3NdX3 = d3Ndu3;
     return true;
}
```

src/ASM/ASMs2D.C
akva2: yeah sure, it's fine, just, consistency was why i did as i did was my point.

kmokstad:
That's because the if-statement is wrong. It was fixed in the last commit in this PR. You'll need that one to test it.

akva2:
uhm, i'm up to date in this pr. i was not in the upstream though.

akva2:
it should, so something is indeed off, but i can't see what :(

akva2:
wait, looked again. that's exactly what my expressions will yield. so i'm back to thinking it's okay.

akva2:
found, pr'd to your ifem branch.

kmokstad:
Included the LR version now, it runs, but the adaptive test does not yield symmetric shear force results as the tensor splines version do. So this needs to be investigate further. For the final mesh (PointLoadedPlate-adap.xinp):
```
  Point #1:     sol1 =  1.266489e-01
                sol2 =  4.738063e-01  4.738063e-01  2.431959e-06 -6.109655e+00 -6.089919e+00
  Point #2:     sol1 =  7.795977e-02
                sol2 =  9.959507e-02  5.907097e-02  5.683958e-06  3.040253e-01 -1.051361e-01
  Point #3:     sol1 =  7.796022e-02
                sol2 =  5.876334e-02  9.991856e-02 -4.690785e-06  3.154557e-01  7.053218e-01
  Point #4:     sol1 =  5.205996e-02
                sol2 =  4.440930e-02  4.418585e-02 -4.308239e-02  3.239277e-01  3.203651e-01
```
Bending moments (first three in sol2) are symmetrix, but not component 4 and 5 (shear forces)

kmokstad:
jenkins build this with downstreams ifem-stokes=107 please

kmokstad:
jenkins build this with downstreams ifem-elasticity=69 ifem-stokes=107 please

akva2:
stokes pr has been merged so build will fail due to that.

kmokstad:
jenkins build this with downstreams ifem-elasticity=69 please

akva2:
@TheBB

kmokstad:
It seems we might have an external user of the plate/shell application soon, and then this feature (shear force calculation) is highly relevant. I threfore want to conclude this one now. The only unresolved is the apparent randomness of the vtfls output, making it tricky to create a reg-test for the shear force output.

akva2:
jenkins build this with downstreams ifem-elasticity=69 please

akva2:
jenkins build this with downstreams ifem-elasticity=69 please

PR #319, opened 2018-11-29
Added: support for LR nurbs in 2D

akva2:


kmokstad:
Updated the Elasticity PR to skip shear force calculation, then I can run adaptively with LRNurbs asn is back to the same seg-fault during vtf export as explained above

src/ASM/LR/ASMu2Dnurbs.h
```diff
@@ -67,6 +67,9 @@ class ASMu2Dnurbs : public ASMu2D
   //! \brief Evaluate basis functions and two derivatives in a point.
   virtual void computeBasis(double u, double v,
                             Go::BasisDerivsSf3& bas, int iel) const;
+
+private:
+  bool isNurbs;
```
kmokstad: I think it would make more sense with a variable `noNurbs` which default is false, since that is the exception.
And the doxy is missing.

src/ASM/LR/ASMu2Dnurbs.C
```diff
@@ -76,7 +78,13 @@ bool ASMu2Dnurbs::read (std::istream& is)
   } else { // probably a SplineSurface, so we'll read that and convert
     tensorspline = new Go::SplineSurface();
     is >> *tensorspline;
-    lrspline.reset(createFromTensor(tensorspline));
+    if (tensorspline->rational())
+      lrspline.reset(createFromTensor(tensorspline));
+    else {
+      lrspline.reset(new LR::LRSplineSurface(tensorspline));
+      std::cout << "LR-nurbs requested but input is a spline" << std::endl;
+      isNurbs = false;
```
kmokstad: Guess we need to set the isNurbs/noNurbs flag also when reading an lr-file.
I presume that file format does not account for LR nurbs yet?

src/ASM/LR/ASMu2Dnurbs.C
```diff
@@ -0,0 +1,360 @@
...
+  geo = lrspline.get();
+  return true;
+}
+
+
+bool ASMu2Dnurbs::raiseOrder (int ru, int rv)
+{
+  if (!this->ASMu2D::raiseOrder(ru,rv))
+    return false;
+
```
kmokstad: Her mangler det vel forøvrig en `if (noNurbs) return true;`

kmokstad:
This seems to work now. However, the added third-derivatives cannot be tested on the shell problem (ScordelisPoint), because shear force calculations are so far implemented for plate problems only.
So you can test it on, e.g., `PointLoadedPlate-adap.xinp`, for which you get only NaN out. I've traced this back to the element node coordinates being calculated. Maybe there is an issue with the number of space dimensions being 2 here instead of 3?

kmokstad:
No, the NaN came because that plate problem of course was defined with splines (no NURBS), using the default geometry feature, and a fall-back to the parent class method for noNurbs was missing in the getElementCoordinates method.

Added a NURBS version of this model now in OPM/IFEM-Elasticity#79, which I expect to give identical results as the noNurbs version. And it does, except for the shear forces, i.e., the third derivatives.

src/ASM/LR/ASMu2Dnurbs.C
```diff
@@ -0,0 +1,503 @@
...
+  bas.prepareDerivs(u, v, 0, -1, tmp.basisValues.size());
+  for (size_t i = 0; i < tmp.basisValues.size(); ++i) {
+    double w = lrspline->getBasisfunction(MNPC[iel][i])->cp(wi);
+    bas.basisValues[i] = tmp.basisValues[i]*w/W;
+    double H1 = (tmp.basisDerivs_u[i]*W-tmp.basisValues[i]*Wderxi);
+    bas.basisDerivs_u[i] = H1*w/(W*W);
+    double H2 = (tmp.basisDerivs_v[i]*W-tmp.basisValues[i]*Wdereta);
+    bas.basisDerivs_v[i] = H2*w/(W*W);
+    double dH1dx = tmp.basisDerivs_uu[i]*W-tmp.basisValues[i]*Wder2xi;
+    double dH2dx = tmp.basisDerivs_vv[i]*W-tmp.basisValues[i]*Wder2eta;
```
kmokstad: This can't be right. `dH2dx` is here the same as `dH2dy` in l433.
Shouldn't it be equal to `dH1dy`instead (symmetry)?

kmokstad:
jenkins build this with downstreams ifem-elasticity=79 please

akva2:
jenkins build this with downstreams ifem-elasticity=79 please

kmokstad:
Still some OpenMP issues for Poisson, it looks like. The LinEl failure I could not reproduce locally.

akva2:
jenkins build this with downstreams ifem-elasticity=79 please

akva2:
jenkins build this with downstreams ifem-elasticity=79 please

src/ASM/LR/ASMu2D.C
```diff
@@ -2510,17 +2572,14 @@ bool ASMu2D::refine (const LR::RefineData& prm,
...
-void ASMu2D::generateBezierBasis()
+void ASMu2D::generateBezierBasis ()
 {
-  const int p1 = geo->order(0);
-  const int p2 = geo->order(1);
-  bezier_u = this->getBezierBasis(p1);
-  bezier_v = this->getBezierBasis(p2);
-
+  bezier_u = getBezierBasis(geo->order(0));
+  bezier_v = getBezierBasis(geo->order(1));
```
kmokstad: Since this method was introduced in the first commit of this PR, the cosmetic changes above (l2515-2519) should be squashed into that commit instead, to reduce the size of this commit

src/ASM/LR/ASMu2Dnurbs.C
```diff
@@ -0,0 +1,339 @@
...
+  if (noNurbs)
+    return this->ASMu2D::evaluateBasis(iel,fe,derivs);
+
+  const LR::Element* el = lrspline->getElement(iel);
+  if (!el) return false;
+
+  fe.xi  = 2.0*(fe.u - el->umin()) / (el->umax() - el->umin()) - 1.0;
+  fe.eta = 2.0*(fe.v - el->vmin()) / (el->vmax() - el->vmin()) - 1.0;
+  RealArray Nu = bezier_u.computeBasisValues(fe.xi, derivs);
+  RealArray Nv = bezier_v.computeBasisValues(fe.eta,derivs);
```
kmokstad: In the corresponding ASMu2D-method you put a `#pragma omp critical` on these two statements. But I guess the same would apply here also? Or are they no longer needed. Should be consequent at least.

akva2:
jenkins build this with downstreams ifem-elasticity=79 please

PR #320, opened 2018-11-30
changed: simplify some CompatibleOperators

akva2:
use add and outer_product instead of hand-rolled loops


akva2:

PR #321, opened 2018-11-30
Fixes for some issue encountered while setting up adaptive poroelastic fracture

akva2:


src/ASM/LR/ASMLRSpline.C
```diff
@@ -158,16 +158,18 @@ bool ASMunstruct::refine (const LR::RefineData& prm,
...
   for (size_t j = 0; j < sol.size(); j++)
-    if (!(nf[j] = LR::extendControlPoints(geo,sol[j],this->getNoFields(1))))
+    if (!sol[j].empty() && !(nf[j] = LR::extendControlPoints(geo,sol[j],this->getNoFields(1))))
       return false;
 
   if (!this->doRefine(prm,geo))
     return false;
 
   nnod = geo->nBasisFunctions();
   for (int i = sol.size()-1; i >= 0; i--) {
```
kmokstad: Canot see the reason for the backward loop here. Looks weird.

akva2: it's a LIFO stack, ie, we can only remove the last component(s) for a node.

akva2:

PR #322, opened 2018-12-09
 Fixed: Pass global (not local) element number to the SAM assembly methods

### LINKED ISSUE #301: ASMs?D::findElementContaining does not account for zero knot-span element
ASMs2D::findElementContaining returns wrong element index if the specified parameters matches a knot with multiplicity 2 or higher. This implies we cannot run the Kirchhoff plate problem with point load, where the regularity is reduced to better fit the  sharpness of the solution under the point load.

kmokstad:
This fixes #301 because the assembly methods require the global element numbers, where the findElementContining methods returns local patch-wise element indices (including the 0-elements).

The other commit here is some LR cleanups from April.


kmokstad:

PR #323, opened 2018-12-13
Separate projection basis for 1D

kmokstad:
Revised part of #316 concerning 1D models.

The input is here handled in a slightly different way. Instead of an extra attribute on the refine/raiseorder tags, a separate `<projection>`-tag within the `<geometry>`scope is introduced, where commands that should apply to the projection basis are specified.

PR #325, opened 2019-01-10
Separate projection basis for 2D

kmokstad:
In addition to #323, this includes the equivalent for 2D as well. Can do 3D when that is needed, after this one has been approved and merged.

Notice that the adaptive refinement of the projection basis does not work as it is done here (by @akva2). It seems that way to many edges are inserted. I leave that to the LR experts to figure out. Therefore the WIP label.


kmokstad:
jenkins build this with downstreams ifem-elasticity=75 please

kmokstad:
The refinement of the projection basis now seems to work. It only remains to test the projection-basis-from-file before we can merge this one

kmokstad:
jenkins build this with downstreams ifem-elasticity=75 please

PR #326, opened 2019-01-15
Boundary conditions in local axis directions for 1D

kmokstad:
This (last commit) adds the support for boundary conditions in skew coordinate systems for 1D patches.
The local axes are calculated with the tanget direction as the local X-axis, the the binormal vector (that is, the normal vector of the plane of curvature) as the local Y-axis. If the curve is straight, the local Y-direction is chosen arbitrarily.

There is a potential issue when it comes to fixing rotations for C1-patches, where the local directions are are imposed to the second control point. What is the tangent (and binormal) directions here? Currently, we use the same as for the end point, but this might not be strictly correct unless the two points are very close to each other, or the curvature is very small.

### DIFF
--- a/src/ASM/ASM1D.h
+++ b/src/ASM/ASM1D.h
@@ -72,7 +72,6 @@ public:
   //! \param[in] xi Parameter value along the curve
   //! \param[in] dof Which DOFs to constrain at the node
   //! \param[in] code Inhomogeneous dirichlet condition code
-  //! \param[in] basis Basis to constrain node for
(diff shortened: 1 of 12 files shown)


kmokstad:
Not tested yet, so WIP

kmokstad:
jenkins build this with downstreams ifem-elasticity=74 ifem-beamex=18 please

akva2:
jenkins build this with downstreams ifem-elasticity=74 ifem-beamex=18 please

PR #329, opened 2019-01-23
Fixed: Removed LR::-prefixing of the basis argument in LR-scoped methods

kmokstad:
Only to avoid doxygen complaints.


kmokstad:

PR #334, opened 2019-02-04
Fixed mixed adaptive simulations on meshes with reduced continuity

akva2:
Previously we established a full-continuity basis for projection in these cases. We now keep the reduced continuity lines in the projection basis.

There is one caveat; we cannot automate this procedure when geometry is linear. In that case, a separate projection basis has to be specified through the <projection> keyword in the files.
I think this is a better compromise, the alternative is to flip the mixed establishment completely; have user give the highest order basis, and then establish the pressure basis through lowering order instead. This would require updating every single mixed setup already existing.

### DIFF
--- a/src/ASM/ASMmxBase.C
+++ b/src/ASM/ASMmxBase.C
@@ -16,6 +16,7 @@
 #include "GoTools/geometry/SurfaceInterpolator.h"
 #include "GoTools/trivariate/SplineVolume.h"
 #include "GoTools/trivariate/VolumeInterpolator.h"
+#include "SplineUtils.h"
 #include <numeric>
 
 char ASMmxBase::geoBasis             = 2;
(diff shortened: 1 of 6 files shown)


akva2:
jenkins build this with downstreams ifem-stokes=110 please

akva2:
jenkins build this with downstreams ifem-stokes=110 please

src/ASM/ASMmxBase.C
```diff
@@ -417,40 +427,55 @@ Go::SplineSurface* ASMmxBase::raiseBasis (Go::SplineSurface* surf)
...
+    std::vector<int> mult;
+    svol->basis(idx).knotMultiplicities(mult);
+    std::vector<int> cont(knots.size());
+    cont.front() = cont.back() = -1;
+    int order = svol->order(idx);
+    for (size_t i = 1; i < knots.size()-1; ++i)
+      if (mult[i] != 1)
+        cont[i] = order-mult[i]-1;
+      else
+        cont[i] = order-1;
```
kmokstad: same here

src/SIM/SIM2D.C
```diff
@@ -279,6 +279,11 @@ bool SIM2D::parseGeometryTag (const TiXmlElement* elem)
       IFEM::cout <<"\tCollapsed edge P"<< patch <<" E"<< edge << std::endl;
       pch->collapseEdge(edge);
     }
+
+#ifdef USE_OPENMP
+      // Cannot do multi-threaded assembly with collapsed edges
+      omp_set_num_threads(1);
```
kmokstad: The indentation here looks wrong. Anyway, a better way of fixing this would be to alter the stripe direction of be aligned with the collapse edge (thought we did that last year when doing the hemisphere problem, but can't find it)

akva2:
jenkins build this with downstreams ifem-stokes=110 please

PR #335, opened 2019-02-07
Some cleaning of the SIMxD classes

kmokstad:
* In the first commit, the beam-specific features are removed (they are now placed in the SIMElasticBar class instead), since this has relevance for elastic beam applications only.
* In the second commit, the periodicity input handling is moved to SIMinput and a virtual method in ASMbase is introduced to avoid casting the spline patch object when applying the periodicity conditions.


kmokstad:
jenkins build this with downstreams ifem-elasticity=78 ifem-fsi=37 please

PR #338, opened 2019-03-01
Support for spline patch definition in the xinp directly instead of via file

kmokstad:
A convenience for small models that are not easily covered by the model-generator.
Did this last fall, then almost forgot it. But think it is nice-to-have functionality.
The others here are minor things.


kmokstad:
jenkins build this with downstreams ifem-elasticity=81 please

kmokstad:
Not sure what is meant my "orthogonal changes" here but if referring to the first commit, maybe `git show -b ba085` is better as space-only differences then are ignored.

PR #340, opened 2019-03-06
Some multipatch issues

kmokstad:
This fixes some issues when running with ignored patches in a big multi-patch model (a debugging tool).
Also, there was an index error when writing projected results to VTF such that they we incorrect for multi-patch models.


kmokstad:

kmokstad:

PR #341, opened 2019-03-07
added: singleton entry point for stb image

akva2:
the header can only be used once. we need this wrapper
class with multiple texture material types.


akva2:
jenkins build this please

PR #348, opened 2019-03-26
Fixed: Use InsertAfterChild instead of LinkEndChild in XMLInputBase

kmokstad:
when the included file contains for than one tag, to ensure they are inserted into the right place.
Added: Debug print of the expanded XML-file content.


kmokstad:

kmokstad:

PR #349, opened 2019-04-02
Support for homogeneous Dirichlet for LR-C1

kmokstad:
With this we can model clamped Kirchhoff plates/shells also with LR-splines (i.e., adaptivity).
In-homogeneous conditions (prescribed rotation) and local axes are not supported, maybe later.
Also some general cleaning of the Dirichlet-handling for LR as a prerequisite,  and a fix on the number of Gauss points on boundaries when auto-selected and different order in the parameter directions.


kmokstad:
jenkins build this with downstreams ifem-elasticity=86 please

kmokstad:
The IGA master students would like to test some clamped problems..

PR #350, opened 2019-04-07
Added: Optionally collapse degenerated faces into an edge or node.

kmokstad:
Changed: Removed the basis argument in the constrainNode method
in ASMs3D to be compliant with the ASMs2D method (they are not
in use for mixed problems). Also some minor doxy corrections.
Added: Unit tests for the collapse methods for ASMs2D and ASMs3D.

This might be needed for Arcon's testing.


kmokstad:
jenkins build this with downstreams ifem-elasticity=87 please

PR #351, opened 2019-04-09
added: ASMbase method to obtain element-element connectivities

akva2:
Needed for graph partitioning.

### DIFF
--- a/src/ASM/ASMbase.h
+++ b/src/ASM/ASMbase.h
@@ -312,6 +312,9 @@ public:
   virtual bool getParameterDomain(Real2DMat& u,
                                   IntVec* corners = nullptr) const = 0;
 
+  using NeighArray = std::vector<std::vector<int>>; //!< Convenience alias
+  //! \brief Obtain element neighbours.
+  virtual void getNeighbours(NeighArray& neighs) const = 0;
 
   // Various preprocessing methods
   // =============================
--- a/src/ASM/ASMs1D.C
+++ b/src/ASM/ASMs1D.C
@@ -1863,3 +1863,14 @@ Fields* ASMs1D::getProjectedFields (const Vector& coefs, size_t) const
             << std::endl;
   return nullptr;
 }
+
+
+void ASMs1D::getNeighbours (NeighArray& neigh) const
+{
+  for (int i = 0; i < static_cast<int>(nel); ++i) {
+    if (i-1 > -1)
+      neigh[i].push_back(i-1);
+    if (i+1 < static_cast<int>(nel))
+      neigh[i].push_back(i+1);
+  }
+}
--- a/src/ASM/ASMs2D.C
+++ b/src/ASM/ASMs2D.C
@@ -3068,3 +3068,25 @@ size_t ASMs2D::getNoProjectionNodes () const
 {
   return proj->numCoefs_u() * proj->numCoefs_v();
 }
+
+
+void ASMs2D::getNeighbours (NeighArray& neigh) const
+{
+  int N1, N2, dummy;
+  this->getNoStructElms(N1, N2, dummy);
+  for (int i = 0; i < static_cast<int>(this->getNoElms(true)); ++i) {
+    int idx = this->getElmID(i+1)-1;
+    if (idx == -1)
+      continue;
+
+    neigh[idx].resize(4,-1);
+    if (i % N1 > 0)
+      neigh[idx][0] = idx-1;
+    if (i % N1 != N1-1)
+      neigh[idx][1] = idx+1;
+    if (i >= N1)
+      neigh[idx][2] = idx-N1;
+    if (i < N1*(N2-1))
+      neigh[idx][3] = idx+N1;
+  }
+}
--- a/src/ASM/ASMs3D.C
+++ b/src/ASM/ASMs3D.C
@@ -3309,3 +3309,29 @@ bool ASMs3D::getFaceSize (int& n1, int& n2, int basis, int face) const
 
   return true;
 }
+
+
+void ASMs3D::getNeighbours (NeighArray& neigh) const
+{
+  int N1, N2, N3;
+  this->getNoStructElms(N1, N2, N3);
+  for (int i = 0; i < static_cast<int>(this->getNoElms(false)); ++i) {
+    int idx = this->getElmID(i+1)-1;
+    if (idx == -1)
+      continue;
+
+    neigh[idx].resize(6,-1);
+    if (i % N1 > 0)
+      neigh[idx][0] = idx-1;
+    if (i % N1 != N1-1)
+      neigh[idx][1] = idx+1;
+    if (i % (N1*N2) >= N1)
+      neigh[idx][2] = idx-N1;
+    if (i % (N1*N2) < N1*(N2-1))
+      neigh[idx][3] = idx+N1;
+    if (i / (N1*N2) > 0)
+      neigh[idx][4] = idx - (N1*N2);
+    if (i / (N1*N2) < N3-1)
+      neigh[idx][5] = idx + (N1*N2);
+  }
+}
--- a/src/ASM/LR/ASMu2D.C
+++ b/src/ASM/LR/ASMu2D.C
@@ -2631,3 +2631,30 @@ void ASMu2D::computeBasis (double u, double v, Go::BasisDerivsSf3& bas,
 
   lrspline->computeBasis(u,v,bas,iel);
 }
+
+
+void ASMu2D::getNeighbours (NeighArray& neighs) const
+{
+  std::vector<std::set<int>> neigh(this->getNoElms(true));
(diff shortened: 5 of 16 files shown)


akva2:
jenkins build this please

src/ASM/ASMbase.h
```diff
@@ -312,6 +312,9 @@ class ASMbase
   virtual bool getParameterDomain(Real2DMat& u,
                                   IntVec* corners = nullptr) const = 0;
 
+  using NeighArray = std::vector<std::vector<int>>; //!< Convenience alias
```
kmokstad: `IntMat` has been typedef'ed to the same thing in the top of the file. So you might consider
`using NeighArray = IntMat;`
if you want to use this name instead?

src/ASM/ASMs1D.C
```diff
@@ -1863,3 +1863,14 @@ Fields* ASMs1D::getProjectedFields (const Vector& coefs, size_t) const
...
 }
+
+
+void ASMs1D::getNeighbours (NeighArray& neigh) const
+{
+  for (int i = 0; i < static_cast<int>(nel); ++i) {
+    if (i-1 > -1)
+      neigh[i].push_back(i-1);
+    if (i+1 < static_cast<int>(nel))
+      neigh[i].push_back(i+1);
```
kmokstad: For 2D and 3D you first resize and fill with -1, but here the first and last element will have only 1 element in its neighbor list. Can that be a problem? At least inconsistent.

And here the local patch-wise element index is used as `neigh` index, whereas for 2D and 3D the global one is used (from `getElmID`). So in the unlikely event of a parallel 1D multi-patch simulation this will fail, won't it?

src/ASM/ASMs2D.C
```diff
@@ -3068,3 +3068,25 @@ size_t ASMs2D::getNoProjectionNodes () const
 {
   return proj->numCoefs_u() * proj->numCoefs_v();
 }
+
+
+void ASMs2D::getNeighbours (NeighArray& neigh) const
+{
+  int N1, N2, dummy;
+  this->getNoStructElms(N1, N2, dummy);
+  for (int i = 0; i < static_cast<int>(this->getNoElms(true)); ++i) {
```
kmokstad: I think maybe a double loop over parameter directions (like in `genereteFEMTopology`) is preferable. Then this can be done without all the if- and mod-ery below. At least in 3D that will create more readable code.

src/ASM/ASMs2D.C
```diff
@@ -3068,3 +3068,25 @@ size_t ASMs2D::getNoProjectionNodes () const
...
+    int idx = this->getElmID(i+1)-1;
+    if (idx == -1)
+      continue;
+
+    neigh[idx].resize(4,-1);
```
kmokstad: So the big assumption here is that `neigh` has been resized outside to big enough?

src/ASM/ASMs2D.C
```diff
@@ -3068,3 +3068,25 @@ size_t ASMs2D::getNoProjectionNodes () const
...
+    if (i % N1 > 0)
+      neigh[idx][0] = idx-1;
```
kmokstad: I guess this will always work today, but you here implicitly assume that the global element numbers (in `MLGE`) have the same sequence as the patch-wise ordering). But if that is changed this will fail. It would be more consistent with
```
   neigh[idx][0] = this->getElmID(i);
   neigh[idx][1] = this->getElmID(i+2);
   neigh[idx][2] = this->getElmID(i-N1+1);
   neigh[idx][3] = this->getElmID(i+N1+1);
```
or use MLGE lookup directly instead of using getElmID.

src/ASM/ASMs3D.C
```diff
@@ -3309,3 +3309,29 @@ bool ASMs3D::getFaceSize (int& n1, int& n2, int basis, int face) const
...
+      neigh[idx][1] = idx+1;
+    if (i % (N1*N2) >= N1)
+      neigh[idx][2] = idx-N1;
+    if (i % (N1*N2) < N1*(N2-1))
+      neigh[idx][3] = idx+N1;
+    if (i / (N1*N2) > 0)
+      neigh[idx][4] = idx - (N1*N2);
+    if (i / (N1*N2) < N3-1)
+      neigh[idx][5] = idx + (N1*N2);
+  }
```
kmokstad: All my comments for the 2D method apply here as well.

src/ASM/LR/ASMu2D.C
```diff
@@ -2631,3 +2631,30 @@ void ASMu2D::computeBasis (double u, double v, Go::BasisDerivsSf3& bas,
...
+    double umid = (m->umin() + m->umax()) / 2.0;
+    double vmid = (m->vmin() + m->vmax()) / 2.0;
+    std::vector<RealArray> ua = {{m->umin() - epsilon, vmid},
+                                 {m->umax() + epsilon, vmid},
+                                 {umid, m->vmin() - epsilon},
+                                 {umid, m->vmax() + epsilon}};
+    for (const RealArray& u : ua) {
+      int el1 = lr->getElementContaining(u);
+      if (el1 > -1)
+        neigh[m->getId()].insert(el1);
```
kmokstad: Need here also global element indices, i.e., `.insert(this->getElmID(el1+1))` ?

src/ASM/LR/ASMu2D.C
```diff
@@ -2631,3 +2631,30 @@ void ASMu2D::computeBasis (double u, double v, Go::BasisDerivsSf3& bas,
...
+    }
+  }
+
+  for (size_t i = 0; i < neigh.size(); ++i) {
+    int idx = this->getElmID(i);
```
kmokstad: Hup, `getElmID` takes 1-based indices as argument, so `i+1` here (or replace my `MLGE[i]`, the getElmID method is more intended for external use, here and elsewhere in this commit).

src/ASM/LR/ASMu3D.C
```diff
@@ -2260,3 +2260,33 @@ void ASMu3D::extendRefinementDomain (IntSet& refineIndices,
...
+    std::vector<RealArray> ua = {{m->umin() - epsilon, vmid, wmid},
+                                 {m->umax() + epsilon, vmid, wmid},
+                                 {umid, m->vmin() - epsilon, wmid},
+                                 {umid, m->vmax() + epsilon, wmid},
+                                 {umid, vmid, m->wmin() - epsilon},
+                                 {umid, vmid, m->wmax() + epsilon}};
+    for (const RealArray& u : ua) {
+      int el1 = lr->getElementContaining(u);
+      if (el1 > -1)
+        neigh[m->getId()].insert(el1);
```
kmokstad: Again, `getElmID` is needed (or `MLGE[el1]`)

src/ASM/LR/ASMu3D.C
```diff
@@ -2260,3 +2260,33 @@ void ASMu3D::extendRefinementDomain (IntSet& refineIndices,
...
+    }
+  }
+
+  for (size_t i = 0; i < neigh.size(); ++i) {
+    int idx = this->getElmID(i);
```
kmokstad: Same here, as in ASMu2D.C:2656

src/ASM/LR/Test/TestASMu2D.C
```diff
@@ -255,3 +255,36 @@ TEST(TestASMu2D, Connect)
...
   ASSERT_TRUE(sim2.createFEMmodel());
 }
+
+
+class ASMuSquare : public ASMu2D
+{
+public:
+  ASMuSquare()
+  {
+    std::stringstream geo("200 1 0 0\n2 0\n2 2\n0 0 1 1\n2 2\n0 0 1 1\n0 0\n1 0\n0 1\n1 1\n");
```
kmokstad: Only the last `\n` is actually needed, I think

src/ASM/ASMbase.h
akva2: I have changed to use IntMat.

src/ASM/ASMs1D.C
akva2: now consistent with the others.

src/ASM/ASMs2D.C
akva2: i don't immediately see how a double loop would help.

src/ASM/ASMs2D.C
akva2: yes. this is called from the SIM and the SIM call make sure of that.

src/ASM/ASMs2D.C
akva2: indeed.

src/ASM/LR/ASMu2D.C
akva2: indeed.

akva2:
continued in #352

PR #354, opened 2019-04-23
Fixed: Undo the fix in e69c100b17d0e5cd59d3d4cb7efe3fa1302754c7

### LINKED ISSUE #347: Adaptive multipatch failure
The attached model [Frame.zip](https://github.com/OPM/IFEM/files/3005267/Frame.zip) fails in the 6th adaptive step with the error message:
```
	Connecting P11 E3 to P8 E4 reversed? 0
 *** ASMu2D::connectBasis: Non-matching edges, sizes 9 and 6
 *** SIM2D::parse: Error establishing connection.
  Parsing <topologysets>
 *** SIMadmin::read: Failure occured while parsing "geometry".
```
Run with `LinEl -adap Frame-AMR-b02-p2.xinp`to trigger, with #346 and OPM/IFEM-Elasticity#85. Apparently there is some control point mismatches over the patch interfaces although I thought that was supposed to be automatically handled by now?

kmokstad:
Fixes #347 which apparently was caused by this workaround.

Should probably revisit this later for a proper fix, but as this keeps popping up randomly for the multi-patch VCP test, I think we should merge this first. So far only one downstream regression is affected.

### DIFF
--- a/src/ASM/LR/ASMu2D.C
+++ b/src/ASM/LR/ASMu2D.C
@@ -383,7 +383,7 @@ void ASMu2D::extendRefinementDomain (IntSet& refineIndices,
       for (int edgeNode : bndry1[edge])
         if (edgeNode-1 == j)
         {
-          IntVec secondary = this->getOverlappingNodes(j);
+          IntVec secondary = this->getOverlappingNodes(j, edge/2+1);
           refineIndices.insert(secondary.begin(),secondary.end());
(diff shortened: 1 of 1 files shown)


kmokstad:
jenkins build this if downstreams ifem-poisson=32 please

kmokstad:
jenkins build this with downstreams ifem-poisson=32 please

PR #356, opened 2019-05-07
Fix periodicities on refined models and clamping for LRnurbs

kmokstad:
This fixes two issues we discovered with the IGA shell master students last week:

* Unable to apply clamped BCs on LRnurbs patches
* Specification of periodic boundaries was broken if model is refined (or order-elevated

The latter was a side-effect of the cleaning in #335 that was not caught by any regression test. Therefore added one.


kmokstad:

kmokstad:

kmokstad:

kmokstad:

PR #359, opened 2019-05-17
Fixed: Account for NURBS weights in ASMu2D::getNodalCoordinates also

kmokstad:
This is a leftover from #319. Without this, the L2-projection does not work for non-linear shell analysis as the NURBS weights are not accounted for when calculating the updated control point positions.


kmokstad:

PR #360, opened 2019-05-21
added: support output of explicit function to VTF

akva2:
Useful for inspection purposes

### DIFF
--- a/src/SIM/SIMoutput.C
+++ b/src/SIM/SIMoutput.C
@@ -21,6 +21,7 @@
 #include "Vec3Oper.h"
 #include "VTF.h"
 #include "ElementBlock.h"
+#include "Functions.h"
 #include "Utilities.h"
 #include "IFEM.h"
 #include "tinyxml.h"
@@ -40,6 +41,8 @@ SIMoutput::SIMoutput (IntegrandBase* itg) : SIMinput(itg)
 SIMoutput::~SIMoutput ()
 {
   if (myVtf) delete myVtf;
+  for (auto& func : myAddScalars)
+    delete func.second;
 }
 
 
@@ -83,7 +86,20 @@ bool SIMoutput::parseOutputTag (const TiXmlElement* elem)
 {
   IFEM::cout <<"  Parsing <"<< elem->Value() <<">"<< std::endl;
 
-  if (strcasecmp(elem->Value(),"resultpoints"))
+  if (!strcasecmp(elem->Value(),"function")) {
+    std::string name;
+    utl::getAttribute(elem,"name",name);
+    RealFunc* f = nullptr;
+    if (!name.empty()) {
+      IFEM::cout << "Writing additional function with name '" << name << "' to VTF";
+      f = utl::parseRealFunc(utl::getValue(elem,"function"),"expression");
+      if (f)
+        myAddScalars[name] = f;
+    }
+
+    return f != nullptr;
+  }
+  else if (strcasecmp(elem->Value(),"resultpoints"))
     return this->SIMinput::parseOutputTag(elem);
 
   bool newGroup = true;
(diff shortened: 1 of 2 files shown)


src/SIM/SIMoutput.C
```diff
@@ -83,9 +86,23 @@ bool SIMoutput::parseOutputTag (const TiXmlElement* elem)
 {
   IFEM::cout <<"  Parsing <"<< elem->Value() <<">"<< std::endl;
 
-  if (strcasecmp(elem->Value(),"resultpoints"))
+  if (strcasecmp(elem->Value(),"resultpoints") && strcasecmp(elem->Value(),"function"))
     return this->SIMinput::parseOutputTag(elem);
 
+  if (!strcasecmp(elem->Value(),"function")) {
```
kmokstad: A better logic would be:
```
  if (!strcasecmp(elem->Value(),"function"))
  {
    ...do the new function stuff...
  }
  else if (strcasecmp(elem->Value(),"resultpoints"))
    return this->SIMinput::parseOutputTag(elem);

  ...do the rest...
```
and you don't need to check for "function" tag more than once

src/SIM/SIMoutput.C
```diff
@@ -513,8 +530,11 @@ bool SIMoutput::writeGlvS (const Vector& psol, int iStep, int& nBlock,
 {
   idBlock = this->writeGlvS1(psol,iStep,nBlock,time,
                              pvecName,idBlock,psolComps);
+
   if (idBlock < 0)
     return false;
+  else if (!this->writeAddFuncs(iStep,nBlock,50,time))
```
kmokstad: Careful with the idBlock value here (50). Ensure it does not conflict with others.

src/SIM/SIMoutput.C
```diff
@@ -1801,3 +1821,18 @@ bool SIMoutput::serialize (std::map<std::string,std::string>&) const
             <<"     Restart not supported for "<< this->getName() << std::endl;
   return false;
 }
+
+
+bool SIMoutput::writeAddFuncs(int iStep, int& nBlock, int idBlock, double time)
+{
+  if (myAddScalars.empty())
+    return true;
```
kmokstad: This check is not needed. If empty, the loop below will handle it correctly?

src/SIM/SIMoutput.C
```diff
@@ -1801,3 +1821,18 @@ bool SIMoutput::serialize (std::map<std::string,std::string>&) const
...
+
+  for (const auto& func : myAddScalars) {
+    if (!this->writeGlvF(*func.second, func.first.c_str(), iStep,
+                         nBlock, idBlock, time))
+        return false;
+  }
```
kmokstad: Braces are not needed since just single-statement (cosmetic)

src/SIM/SIMoutput.h
```diff
@@ -314,6 +314,9 @@ class SIMoutput : public SIMinput
                          int& nBlock, std::vector< std::vector<int> >& sID,
                          size_t* i = nullptr);
 
+  //! \brief Private helper to write additional functions to VTF-file.
+  bool writeAddFuncs(int iStep, int& nBlock, int idBlock = 50, double time = 0.0);
```
kmokstad: Since it is a private helper, it does not need default values either.

src/SIM/SIMoutput.C
akva2: okay

src/SIM/SIMoutput.C
akva2: ok

src/SIM/SIMoutput.C
akva2: indeed.

src/SIM/SIMoutput.C
akva2: 50 is the default from writeGlvF, but i now make sure to increase it for each registered function (starting at 50).

src/SIM/SIMoutput.h
akva2: indeed

PR #362, opened 2019-05-27
added: allow system matrix dump when doing eigenvalue solves

akva2:


src/SIM/SIMbase.C
```diff
@@ -2071,3 +2037,50 @@ void SIMbase::registerDependency (const std::string& name, SIMdependency* sim,
...
+        strcpy(matName,"A");
+      else
+        sprintf(matName,"A%d",dmp.count);
+      for (int i = 0; M; M = myEqSys->getMatrix(++i), ++matName[0])
+        M->dump(os,dmp.format,matName); // label matrices as A,B,C,...
+      utl::zero_print_tol = old_tol;
+    }
+
+  if (!both)
+    return;
```
kmokstad: Do you actually need this? In case of eigenvalue solver there are usually no right-hand-side vectors, so maybe you can use `if (myEqSys->getNoRHS() =0) return;`instead?

akva2: indeed, changed.

akva2:
jenkins build this please

PR #365, opened 2019-06-27
added: also store the eigen modes in equation ordering

akva2:
so we can output these to hdf5. these are required when doing reduced
order models. this makes it easy to obtain them without having to use
hackery every time.

### DIFF
--- a/src/SIM/SIMbase.C
+++ b/src/SIM/SIMbase.C
@@ -1617,7 +1617,8 @@ bool SIMbase::systemModes (std::vector<Mode>& solution,
   for (int i = 1; i <= nev && ok; i++)
   {
     solution[i-1].eigNo = i;
-    if (!mySam->expandVector(eigVec.getColumn(i),solution[i-1].eigVec))
+    solution[i-1].eqnVec = eigVec.getColumn(i);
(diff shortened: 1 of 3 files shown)


akva2:

akva2:

akva2:
jenkins build this with downstreams ifem-poisson=37 please

src/SIM/SIMbase.C
```diff
@@ -1617,6 +1617,7 @@ bool SIMbase::systemModes (std::vector<Mode>& solution,
   for (int i = 1; i <= nev && ok; i++)
   {
     solution[i-1].eigNo = i;
+    solution[i-1].eqnVec = eigVec.getColumn(i);
     if (!mySam->expandVector(eigVec.getColumn(i),solution[i-1].eigVec))
```
kmokstad: sub-sub optimizing: Could then use `solution[i-1].eqnVec` in the `expandVector()` call to avoid calling `getColumn()` twice

akva2: yup yup. done.

akva2:
jenkins build this with downstreams ifem-poisson=37 please

PR #366, opened 2019-07-01
fixed: index check

akva2:
we only dereference the second entry, so no reason to require
size to be 3


akva2:

akva2:
uhm, no. (2) is the second index, which is equivalent to [1].

kmokstad:
you're right. I got it the wrong way. Then I don't know why this one failed.

akva2:
failure is due to instability in the graph partitioning algoritm. i have removed the problematic stuff from the reg files in stokes after the fact. i'll rerun

akva2:

kmokstad:
OK, then I guess this fix is in order

PR #369, opened 2019-07-10
fixed: properly setup state in some unit tests

akva2:
when we run through the IFEM-test executable, the state
is shared between tests so we have to explicitly set
these values.


akva2:
jenkins build this please

PR #371, opened 2019-07-30
changes needed for basic mixed FSI

akva2:
use of an additional MADOF is required

### DIFF
--- a/src/ASM/ASMs2Dmx.C
+++ b/src/ASM/ASMs2Dmx.C
@@ -1205,3 +1205,13 @@ void ASMs2Dmx::getBoundaryNodes (int lIndex, IntVec& nodes, int basis,
     for (size_t b = 1; b <= this->getNoBasis(); ++b)
       this->ASMs2D::getBoundaryNodes(lIndex, nodes, b, thick, 0, local);
 }
+
+
+bool ASMs2Dmx::updateCoords (const Vector& displ)
+{
+  if (!surf) return true; // silently ignore empty patches
+  if (shareFE) return true;
+
+  surf->deform(displ,nsd);
+  return true;
+}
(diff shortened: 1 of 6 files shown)


src/SIM/SIMbase.C
```diff
@@ -704,7 +705,12 @@ bool SIMbase::updateGrid (const Vector& displ)
   Vector locdisp;
   for (size_t i = 0; i < myModel.size() && ok; i++)
   {
-    myModel[i]->extractNodeVec(displ,locdisp,myModel[i]->getNoSpaceDim(),-1);
+    if (this->mixedProblem())
+      this->extractPatchSolution(displ, locdisp, myModel[i],
+                                 nsd, ASMmxBase::geoBasis);
```
kmokstad: In case of the very unlikely event that patches have different number of space dimensions, `nsd` should perhaps be replaced by `myModel[i]->getNoSpaceDim()`, like in the extractNodeVec call? Mostly for consistency.

src/ASM/ASMs2Dmx.C
```diff
@@ -1205,3 +1205,13 @@ void ASMs2Dmx::getBoundaryNodes (int lIndex, IntVec& nodes, int basis,
...
+
+
+bool ASMs2Dmx::updateCoords (const Vector& displ)
+{
+  if (!surf) return true; // silently ignore empty patches
+  if (shareFE) return true;
+
+  surf->deform(displ,nsd);
+  return true;
+}
```
kmokstad: Suggest that the sanity check in ASMs2D is changed from `!=` to `<` and you don't have to do this reimplementation.
Similar for 1D and 3D.

kmokstad: My main point was that the ASMs2Dmx overload was not needed when changing the sanity check in ASMs2D, but it appears to still be in this commit?

akva2:
whoops. that was meant to be removed indeed, forgot a git checkout.

src/ASM/ASMs2Dmx.C
akva2: remedied.

PR #373, opened 2019-08-01
Fix triangle quadrature

akva2:
This fixes a typo in the 4 point triangle quadrature rule and adds a simple unit test.

### DIFF
--- a/src/Utility/Test/TestTriangleQuadrature.C
+++ b/src/Utility/Test/TestTriangleQuadrature.C
@@ -1,0 +1,35 @@
+//==============================================================================
+//!
+//! \file TestTriangleQuadrature.C
+//!
+//! \date Jul 31 2019
+//!
+//! \author Arne Morten Kvarving / SINTEF
+//!
+//! \brief Tests for triangle quadratures.
+//!
+//==============================================================================
+
+#include "TriangleQuadrature.h"
+
+#include "gtest/gtest.h"
+#include <numeric>
+
+
+TEST(TestTriangleQuadrature, Sanity)
+{
+  // Quadratures that are intended to not exist
+  for (int i : {-1,2,5}) {
+    EXPECT_EQ(TriangleQuadrature::getCoord(i), nullptr);
+    EXPECT_EQ(TriangleQuadrature::getWeight(i), nullptr);
+  }
+
+  for (int i : {1,3,4}) {
+    const double* c = TriangleQuadrature::getCoord(i);
+    const double* w = TriangleQuadrature::getWeight(i);
+    double sum = std::accumulate(w, w+i, 0.0);
+    EXPECT_FLOAT_EQ(sum, 1.0);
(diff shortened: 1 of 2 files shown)


akva2:
jenkins build this please

src/Utility/TriangleQuadrature.C
```diff
@@ -42,7 +42,7 @@ static const double T4[8] = {
 };
 //! \brief 4-point rule weights.
 static const double W4[4] = {
- -0.5635,
+ -0.5625,
```
kmokstad: Most likely you're right, but double check with Tom Hughes' FEM book where these are taken from before approval.

akva2: yeah, i already did. also the sanity check should make it clear - the weights should sum to 1. they don't do currently, they do after the fix.

akva2: btw, in cause you wonder, the reason i ran into this is that coverage is now part of the static analysis jobs, and i noticed this class had no tests.

src/Utility/Test/TestTriangleQuadrature.C
```diff
@@ -0,0 +1,33 @@
...
+#include "TriangleQuadrature.h"
+
+#include "gtest/gtest.h"
+#include <numeric>
+
+
+TEST(TestTriangleQuadrature, Sanity)
+{
+  EXPECT_EQ(TriangleQuadrature::getCoord(-1), nullptr);
+  EXPECT_EQ(TriangleQuadrature::getWeight(-1), nullptr);
```
kmokstad: Doesn't it make more sense to check getCoord(2)/getWeigth(2) instead, as there is no two-point rule?

src/Utility/Test/TestTriangleQuadrature.C
```diff
@@ -0,0 +1,33 @@
...
+  for (int i : {1,3,4}) {
+    const double* c = TriangleQuadrature::getCoord(i);
+    const double* w = TriangleQuadrature::getWeight(i);
+    double sum = std::accumulate(w, w+i, 0.0);
+    EXPECT_FLOAT_EQ(sum, 1.0);
+    for (int j = 0; j < i; ++j)
+      EXPECT_TRUE(c[j] >= 0 && c[j] <= 1.0);
```
kmokstad: Cosmetic: 0.0 here, since 1.0

src/Utility/Test/TestTriangleQuadrature.C
akva2: i can do both.

src/Utility/Test/TestTriangleQuadrature.C
akva2: fixed

akva2:
jenkins build this please

akva2:
jenkins build this please

PR #380, opened 2019-09-18
Support for immersed boundaries from scalar functions

kmokstad:
This adds support for the input
```
<immersedboundary>
   <levelset threshold="0.5" type="file">my_phasefield.dat</levelset>
</immersedboundary>
```
to read in the coefficients of a scalar spline field defining the inside/outside condition.
One can also use `type="expression"` and then enter a general expression in `x` and `y`.
`file` is the default, as well as the threshold value 0.5.


kmokstad:

kmokstad:

src/Utility/Point.h
```diff
@@ -0,0 +1,101 @@
...
+{
+  /*!
+    \brief Class for representing points in 3D space.
+    \details The only difference between this class and the parent class Vec4
+    is that it keeps the spline parameters in an internal array.
+  */
+
+  class Point : public Vec4
+  {
+    Real par[3]; //!< Internal storage of spline parameters
```
akva2: if you use std::array you don't have to write any copy constructor or assignment operators, it will generate just fine then.

kmokstad: Probably, but then there is the `u` pointer in the parent class Vec4. It should not just be copied. In case of this sub-class I want it to point to `par` instead.

akva2: Was referring to the point(const point&)

kmokstad: I know, but in my understanding the generated copy constructor will do `this->u = X.u` while what I do want is `this->u = this->par`. Therefore the explicit implementation.

kmokstad:
jenkins build this with downstreams ifem-elasticity=96 please

src/SIM/SIM2D.C
```diff
@@ -285,10 +296,47 @@ bool SIM2D::parseGeometryTag (const TiXmlElement* elem)
...
+        const char* val = child->FirstChild()->Value();
+        RealFunc* lfunc = nullptr;
+
+        IFEM::cout <<"\tLevel set ("<< type <<")";
+
+        if (type == "file")
+        {
+          IFEM::cout <<" \""<< val <<"\""<< std::endl;
+          std::ifstream ibs(val);
+          if (ibs)
```
akva2: .good()

src/ASM/ASMs2D.C
```diff
@@ -1082,8 +1082,32 @@ void ASMs2D::constrainCorner (int I, int J, int dof, int code, char basis)
   if (basis < 1) basis = 1;
 
   int node = this->getCorner(I,J,basis);
-  if (node > 0)
```
akva2: (ignore line number); s/Tweek/Tweak/g in commit msg.

src/ASM/IBGeometries.h
```diff
@@ -112,4 +114,26 @@ class PerforatedPlate2D : public Immersed::Geometry
...
+class GeoFunc2D : public Immersed::Geometry
+{
+public:
+  //! \brief Default constructor.
+  explicit GeoFunc2D(RealFunc* f = nullptr, double p = 1.0, double eps = 0.5);
+  //! \brief The destructor deletes the function.
+  virtual ~GeoFunc2D();
+
+  //! \brief Performs the inside-outside test for the geometric object.
+  virtual double Alpha(double X, double Y, double Z) const;
```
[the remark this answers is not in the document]
kmokstad: Could be, but it is only in Utility I've been using the `Real` macro. Anyway, this is outdated.

src/Utility/FieldFunctions.h
```diff
@@ -135,6 +135,33 @@ class FieldFunction : public RealFunc, private FieldFuncHDF5
...
 
+/*!
+  \brief A scalar-valued spatial function, defined through scalar fields.
+  \details This class reads its control point values from a stream.
+*/
+
+class FieldFuncStream : public RealFunc, private FieldFuncBase
+{
+public:
+  //! \brief The constructor creates a field from the provided HDF5-file.
```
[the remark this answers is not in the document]
kmokstad: Indeed, copy-paste sloppy. Fixed.

src/SIM/SIM2D.C
kmokstad: Fixed.

src/ASM/ASMs2D.C
kmokstad: ok, corrected.

kmokstad:

kmokstad:

Hoping for a better day :-)

kmokstad:

PR #385, opened 2019-10-19
Miscellaneous updates and fixes

kmokstad:
This gathers a few commits that currently also sit in other branches, but are of a more generic character than those feature branches. Therefore a separate PR for these ones first, as they also affect some downstreams.

### DIFF
--- a/src/ASM/ASMs2D.h
+++ b/src/ASM/ASMs2D.h
@@ -89,6 +89,7 @@ private:
     int                dof;   //!< Local DOF to constrain along the boundary
     int                code;  //!< Inhomogeneous Dirichlet condition code
     std::vector<Ipair> nodes; //!< Nodes subjected to projection on the boundary
+
     //! \brief Default constructor.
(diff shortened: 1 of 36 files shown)


kmokstad:
jenkins build this with downstreams ifem-stokes=125 please

kmokstad:
jenkins build this with downstreams ifem-stokes=125 please

kmokstad:
That's true for the first two commits, but the last three do not sit in other PRs as far I can see.

kmokstad:
Regarding `evalSol1`, we could in principle have a mixed version also in `IntegrandBase`. But it will be harder to generalize than for non-mixed, since the `npv` variable is not used, etc. As far I can see, there are at present only three integrands (PoroElasticity, Stokes and NavierStokes) that override the mixed evalSol method, and with quite difference implementations. So not sure there is much to gain by pulling common things there into a IntegrandBase-method.

kmokstad:
jenkins build this with downstreams ifem-stokes=125 please

PR #387, opened 2019-10-23
added: handle multiple right hand sides in umfpack solve

### LINKED ISSUE #386: UMFPACK mode does not work with multiple right-hand-sides
With #385 I also include the option to use UMFPACK in place of SuperLU in the L2-projection, motivated by the fact that SuperLU decides to produce NaN results (or core dump) instead of returning an error state when something is wrong with the coefficient matrix. However, with UMFAPCK I only got the first stress component out when projecting the stress tensor, all the others were zero! Probably an error in our UMFPACK wrapping. Guess that package itself is supposed to manage multiple RHSs.

akva2:
Fixes #386

### DIFF
--- a/src/LinAlg/SparseMatrix.C
+++ b/src/LinAlg/SparseMatrix.C
@@ -1266,13 +1266,19 @@ bool SparseMatrix::solveUMF (Vector& B, Real* rcond)
     return false;
   if (rcond)
     *rcond = info[UMFPACK_RCOND];
-  umfpack_di_solve(UMFPACK_A,
-                   IA.data(), JA.data(), A.data(),
-                   &X[0], &B[0], numeric, nullptr, info);
-  if (info[UMFPACK_STATUS] == UMFPACK_OK)
+  const size_t nrhs = B.size() / nrow;
+
+  bool okAll = true;
+  for (size_t i = 0; i < nrhs && okAll; ++i) {
+    umfpack_di_solve(UMFPACK_A,
+                     IA.data(), JA.data(), A.data(),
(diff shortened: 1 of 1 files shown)


src/LinAlg/SparseMatrix.C
```diff
@@ -1266,11 +1266,15 @@ bool SparseMatrix::solveUMF (Vector& B, Real* rcond)
...
-    B = X;
+  const size_t nrhs = B.size() / nrow;
+
+  for (size_t i = 0; i < nrhs; ++i) {
+    umfpack_di_solve(UMFPACK_A,
+                     IA.data(), JA.data(), A.data(),
+                     &X[i*nrow], &B[i*nrow], numeric, nullptr, info);
+    if (info[UMFPACK_STATUS] == UMFPACK_OK)
+      std::copy(X.begin()+i*nrow, X.begin()+(i+1)*nrow, B.begin()+i*nrow);
+  }
```
kmokstad: Alternatively:
```
  bool okAll = true;
  for (size_t i = 0; i < nrhs && okAll; ++i) {
    umfpack_di_solve(UMFPACK_A,
                     IA.data(), JA.data(), A.data(),
                     &X[i*nrow], &B[i*nrow], numeric, nullptr, info);
    okAll = info[UMFPACK_STATUS] == UMFPACK_OK;
  }
  if (okAll) B = X;
```

akva2:
umfpack solve is divided in 3 steps;
symbolic factorization, numeric factorization, solution. we only loop the solution step - the factorizations are not repeated.

PR #388, opened 2019-10-31
Fixed: Incorrect sanity check correction

kmokstad:
It turns out I was wrong in e3da110cb9ea7f4fa3634257f64abcdf531782eb (see my review in #371).
The check should instead use the number of coefficient in the patch, which will work both for mixed and regular problems. The check against `MLGN.size()` is still needed as a secondary condition, in case of Dirichlet conditions in local axes (with extra nodes around).


akva2:
let's merge this to restore ci.

kmokstad:

PR #390, opened 2019-11-11
changed: use L2 projection instead of Greville with LR splines

akva2:
the Greville points are often not unique for locally refined
splines. thus L2 is a better default as it will always work.

### DIFF
--- a/src/ASM/LR/ASMu2D.C
+++ b/src/ASM/LR/ASMu2D.C
@@ -2155,7 +2155,9 @@ bool ASMu2D::evalSolution (Matrix& sField, const IntegrandBase& integrand,
   if (project == 'S')
     s = this->scRecovery(integrand);
   else if (project == 'D' || !npe)
-    s = this->projectSolution(integrand);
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

VikingScientist:
The only objection I have is that this might cause identic input-file to produce different results with and without the `-LR` flag.

Also: the tests seem to fail

akva2:
in practice, this only affects output to hdf5. i knew there would be failing tests, cause some of them (hdf5-tests) are broken due to greville point projection being used.

akva2:
i'll fix the downstream failures in companion PRs, but no sense making those before things are approved.

kmokstad:
And I agree with @VikingScientist concern, that an execution with and without the -LR option ideally should produce the same output (when the discretizations are equivalent). It is then better with an "Sorry, Greville projection is not supported for LR" exit instead. Then it is clear.

akva2:
i'll revisit this if it becomes necessary. i found the actual culprit elsewhere.

PR #391, opened 2019-11-12
fixed: use proper basis for evaluation in regularInterpolation

akva2:
woopsie.


akva2:

akva2:
3D does things differently, so it is not affected.

PR #392, opened 2019-11-13
Fixed: Result point evaluation for multi-patch models

kmokstad:
This (the SIMoutput fix) is needed for the digital-twin model, for instance, when there are more than one result points and not all one the same patch. The other commit is more cosmetic, but enables the possibility run without the -controller also (for debugging, etc.)

### DIFF
--- a/src/IFEM.C
+++ b/src/IFEM.C
@@ -13,6 +13,7 @@
 
 #include "IFEM.h"
 #include "LinAlgInit.h"
+#include "ControlFIFO.h"
 #include <iostream>
 #include <cstring>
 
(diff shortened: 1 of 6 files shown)


kmokstad:

src/IFEM.C
```diff
@@ -141,15 +150,48 @@ int IFEM::Init (int arg_c, char** arg_v, const char* title)
...
 
   std::cout << std::endl;
 
   return 0;
 }
 
 
+void IFEM::Close ()
+{
+  if (fifo) delete fifo;
```
akva2: delete is null safe.

kmokstad:
The thought was that it in most cases the FIFO object is not used (only when run with -controller), and then it is more plausible to allocate it dynamically when needed, instead of an always present static member. And I wanted an easy easy to detect if run with controller not not, without needing to check the command-line arguments. Then we don't need the `enableController` member in SIMoptions also.

kmokstad:

PR #393, opened 2019-11-14
Quasi-static modal simulation when beta=0

kmokstad:
This enables the possibility to run NewmarkSIM with SIMmodal in quasi-static mode,
by setting beta to zero, in which case there are no velocity and acceleration vectors around.


kmokstad:
jenkins build this with downstreams ifem-elasticity=98 please

PR #394, opened 2019-12-09
Separate projection basis for 3D

kmokstad:
To complete #323 and #325, here is the equivalent for 3D.


kmokstad:

PR #396, opened 2020-01-23
Fixed: Memory leak in ASMs2Dmx::getBoundary

kmokstad:
Also provide spline parameters to the Vec4 object in the boundary
integration methods, to account for possible field evaluation there.
The remaining changes are cosmetics only.

This (partly) replaces #289 which has been rebased beyond its original state.
It will remove a memory leak from the mixed NonLinEl tests (tQ2-Q1 and Q2Q1).
Have also tried to make the 2D and 3D implementations more similar to ease comparison.
Guess not that well covered by unit/regression tests yet, so please take look at it before approving.

### DIFF
--- a/src/ASM/ASMs2Dmx.C
+++ b/src/ASM/ASMs2Dmx.C
@@ -27,7 +27,6 @@
 #include "Utilities.h"
 #include "Profiler.h"
 #include "Vec3Oper.h"
-#include "Vec3.h"
 #include <array>
 #include <numeric>
 #ifdef USE_OPENMP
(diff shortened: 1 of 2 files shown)


kmokstad:

kmokstad:
I think the failure here is only due to OPM/IFEM-AdvectionDiffusion#40 which also affects the Boussinesq test. @akva2 you fix?

akva2:
sorry about that, fixed now.

akva2:

src/ASM/ASMs2Dmx.C
```diff
@@ -59,12 +59,21 @@ Go::SplineSurface* ASMs2Dmx::getBasis (int basis) const
...
 
   int iedge = dir > 0 ? dir : 3*dir+6;
-
-  return m_basis[basis-1]->edgeCurve(iedge);
+  if (basis > 1)
+    return m_basis[basis-1]->edgeCurve(iedge);
+  else if (!bou[iedge])
+    // Store the pointer to the first basis' edge in bou, to avoid memory leak
+    // since it is dynamically allocated (will be deallocated by parent class).
+    // If invoked with basis > 1 (less likely), we still leak though..
```
akva2: that certainly happens in stokes applications. but it's an improvement of course.

kmokstad: yes, but only when inhomo dirichlet on pressure (or when different basis for x and y-direction).
Ideally, it should have been handled inside the SplineSurface class (like for the trivariate class).

PR #398, opened 2020-02-06
Changed: Use a polymorphic lambda instead of the static function

kmokstad:
and make the parse[Expression|Field]Functions methods private.
Fixed: Avoid crash if the file attribute is forgotten.

This just some cosmetics that should not affect behavior, I hope..


kmokstad:

kmokstad:

PR #400, opened 2020-03-05
Fixed: Avoid writing unnecessary restart file

kmokstad:
In the restart regression tests, a restart hdf5-file is written in both the original execution, and in the restart. It is not needed in the latter, and it ends up in the source code tree where it does not belong.


kmokstad:

PR #403, opened 2020-04-03
changed: allow specifying direction="-1" on robin conditions

akva2:
use this to indicate that the two coefficients alpha and g are packed
as one vector function (alpha | g)


akva2:
jenkins build this with downstreams ifem-poisson=40 please

PR #412, opened 2020-04-28
added: support for saving the local-to-global node mapping in HDF5

akva2:
this is on a per-basis level, ie, not interleaved on a per-patch
basis as used in IFEM

### DIFF
--- a/src/Utility/DataExporter.h
+++ b/src/Utility/DataExporter.h
@@ -46,14 +46,15 @@ public:
 
   //! \brief An enum used to describe the results to write from a SIM
   enum Results {
-    PRIMARY      = 1,   //!< Storage of primary solutions
-    DISPLACEMENT = 2,   //!< Storage of vector fields as displacements
-    SECONDARY    = 4,   //!< Storage of secondary field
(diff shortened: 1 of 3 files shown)


TheBB:
I think the parallel blocks must be protected with ifdefs.

Rebased here: https://github.com/TheBB/IFEM/commit/82bfc71d64a473e90cc904f98f3e4cafad85cb79

akva2:
did you actually get a compile error or just a gut feeling. all those adm methods should be stubbed without mpi.

akva2:
jenkins build this please

akva2:
in any case. ifdefs do no harm, the code has been tested, is non-intrusive, has been available for review for a long time and on "my turf". merging.

TheBB:
Yeah I got a compile error. The definitions and declarations are all behind a guard:

https://github.com/OPM/IFEM/blob/392655583a83380a29ed665d4a263124bc69430d/src/LinAlg/ProcessAdm.h#L72
https://github.com/OPM/IFEM/blob/392655583a83380a29ed665d4a263124bc69430d/src/LinAlg/ProcessAdm.C#L81

PR #413, opened 2020-04-28
Fixed output of projected fields to HDF5

kmokstad:
This (i.e., the last commit, the others of more cosmetic character), should fix the issue with bullshit projected stress fields from the LinearElasticity app. The problem was that the LinEl registers the projected fields in a separate loop (since there in theory can be more than one of them). But inside the HDF5Writer they are treated as the primary solution and projected once again. The result then became of order O(2n) of instead of O(n) (E+18 instead of E+9, etc.)

You need to very this comparing with direct VTF output before approving.

### DIFF
--- a/src/Utility/HDF5Base.C
+++ b/src/Utility/HDF5Base.C
@@ -13,10 +13,8 @@
 
 #include "HDF5Base.h"
 
-#include <iostream>
-
 #ifdef HAS_HDF5
-#include <numeric>
+#include <iostream>
 #include <unistd.h>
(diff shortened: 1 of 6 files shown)


kmokstad:

kmokstad:

kmokstad:
The third commit is now rewritten. Only the cosmetic/refactor remains (in addition to the fix from #415), as the origin problem now is fixed by OPM/IFEM-Elasticity#105.

kmokstad:

kmokstad:

PR #421, opened 2020-06-12
fixed: TestPETScMatrix.Assemble

akva2:
for some reason newer PETSc requires an explicit call to zero out the
matrix.


akva2:
jenkins build this please

PR #422, opened 2020-06-16
added: print reason for stopping adaptive cycles

akva2:
Less chance of people not paying to details being confused.


akva2:

PR #437, opened 2020-11-11
changed: cache L2 projection matrix

akva2:
This is useful when doing repeated projection, like in my SIMRA postprocessing app.


akva2:

akva2:

PR #439, opened 2020-11-25
fixed: we need a separate bool to track if partitioning is enabled

akva2:
empty myElms is not good enough - we might own 0 elements
on the patch

### DIFF
--- a/src/ASM/LR/ASMu3D.C
+++ b/src/ASM/LR/ASMu3D.C
@@ -2499,6 +2499,10 @@ void ASMu3D::generateThreadGroupsFromElms (const IntVec& elms)
     if (this->getElmIndex(elm+1) > 0)
       myElms.push_back(this->getElmIndex(elm+1)-1);
 
+  // We need myElms to be non-empty to flag that partitioning is enabled
+  if (myElms.empty())
+    myElms.push_back(-1);
+
   if (projThreadGroups.size() == 0 || projThreadGroups[0].empty())
     projThreadGroups = threadGroups;
 
--- a/src/ASM/ASMs2D.C
+++ b/src/ASM/ASMs2D.C
@@ -3205,5 +3205,9 @@ void ASMs2D::generateThreadGroupsFromElms (const IntVec& elms)
     if (this->getElmIndex(elm+1) > 0)
       myElms.push_back(this->getElmIndex(elm+1)-1);
 
+  // We need myElms to be non-empty to flag that partitioning is enabled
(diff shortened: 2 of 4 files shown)


akva2:

src/ASM/LR/ASMu3D.C
```diff
@@ -2498,6 +2501,7 @@ void ASMu3D::generateThreadGroupsFromElms (const IntVec& elms)
   for (int elm : elms)
     if (this->getElmIndex(elm+1) > 0)
       myElms.push_back(this->getElmIndex(elm+1)-1);
+  hasElms = true;
```
kmokstad, line 2504: An alternative (more elegant?, at least with less footprint) solution could be to add, e.g. 0 or -1 as a value in myElms here, such that it is never empty if we do generateThreadGroupsFromElms and the test in ASMu3D::integrate will still work.

kmokstad (changes requested):
This would apply for ASMu2D also I persume?
Consider my alternative solution(here I'm not 100% sure about the logic).
At least, clean up the confusing myElms member in both ASMbase and ASMu3D.

src/ASM/LR/ASMu3D.C
kmokstad: The you don't need the hasElms member

src/ASM/LR/ASMu3D.h
```diff
@@ -632,6 +632,7 @@ class ASMu3D : public ASMLRSpline, public ASM3D
   ThreadGroups threadGroups; //!< Element groups for multi-threaded assembly
   ThreadGroups projThreadGroups; //!< Element groups for multi-threaded assembly - projection basis
   IntVec       myElms;       //!< Elements on patch - used with partitioning
```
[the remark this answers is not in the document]
akva2: yeah. in my original code there was no myElms member in ASMbase, they were only stored down in the DomainDecomposition member. I added it after your review comment, and then forgot to remove it in ASMu3D.
I will remove this first, before I redo the rest of this pr, https://github.com/OPM/IFEM/pull/443

src/ASM/LR/ASMu3D.C
akva2: Okay, did this instead.

akva2:
This applies to all ASMs. Just did not include it initially since I wanted review comments first. It's applied everywhere now.

akva2:

kmokstad (approved):
:+1:

PR #445, opened 2020-12-08
fixed: avoid unnecessary expensive refinement parameter calculations

akva2:
if we are shared, these are complete ignored further down. no reason
to calculate them.

useful for adaptive multi-patch fracture dynamics


VikingScientist:
I don't see the place where these results are ignored

akva2:
https://github.com/OPM/IFEM/blob/f6c847cf75862d1c5bbb3f1ae681746923044de1/src/ASM/LR/ASMLRSpline.C#L206

VikingScientist:
jenkins build this please

PR #446, opened 2020-12-09
changed: optimize multi-patch adaptive loop

akva2:
looping over basis functions in LRspline is not cheap.
prefer slightly more convoluted code for a huge speedup.

A small example:

old code:  Total time            |    819.50          |    787.63          |
new code: Total time            |     90.02          |     55.36          |


akva2:
a cache has been added inside liblr instead. new builds have been pushed in ppa (1.6.0). so this is no longer required.

PR #448, opened 2020-12-11
fixed: result point output with graph partitioning

akva2:


akva2:
jenkins build this please

PR #449, opened 2020-12-16
Fix restart for adaptive simulators

akva2:
Continuation of #310


kmokstad:
Replaced by #472 which now is merged.

PR #454, opened 2021-02-03
added: (non-internal) L2 projection support for functions

akva2:
I need this in the Stokes application (to use projected source functions). And since I have separate projection bases there, the current support does not cut it.

@kmokstad I am sure you have some opinion here. I can do it however you see fit, this does the job...

### DIFF
--- a/src/ASM/GlbL2projector.C
+++ b/src/ASM/GlbL2projector.C
@@ -17,6 +17,7 @@
 #include "FiniteElement.h"
 #include "TimeDomain.h"
 #include "ASMbase.h"
+#include "ASMunstruct.h"
 #include "ElmMats.h"
 #include "IntegrandBase.h"
 #include "Function.h"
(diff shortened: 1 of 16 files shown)


akva2:

akva2:

kmokstad (changes requested):
Github says that `GlbL2projector.h` now is binary, so It's hard to review it efficiently.

image

That said, the general concept looks OK, but I suggest the class name `L2Integrand` instead of the length `IntegrandOrFunction`.

src/ASM/GlbL2projector.C
```diff
@@ -33,6 +34,103 @@ LinAlg::MatrixType GlbL2::MatrixType   = LinAlg::SPARSE;
 LinSolParams*      GlbL2::SolverParams = nullptr;
 
 
+namespace {
+
+static void expandTensorGrid2 (const RealArray* in, RealArray* out)
```
kmokstad, line 39: Is it necessary with the anonymous `namespace` scoping here? If declaring the functions static, they will be "visible" only for the code in this file in any case (at least, that's what I thought).

akva2: static functions is the 'C' way to hide functions. anonymous namespace is the C++ way to do it. just gotten used to doing the latter as of late, but i can of course do it static functions instead if you prefer.

akva2:
So abstract base class named L2Integrand, how about the implementations?

src/ASM/GlbL2projector.C
kmokstad: OK, but when you do both, like here, it looks like "smør på flesk"/confusing. I have always used the "static" approach elsewhere in the code, so the only reason I can see is consistency, if the result is equivalent. But you can use namespace if you prefer that.

kmokstad:
yea, the names IntegrandNotFunction and FunctionNotIntegrand are perhaps also a bit cumbersome (allthough the meaning is clear). Both classes are integrands in this context, so those names could be confusing. How about changing the base class name to L2IntegrandBase and then L2Integrand and L2Function for the respective sub-classes, or alternatively L2ProbIntegrand and L2FuncIntegrand?

src/ASM/GlbL2projector.C
akva2: i prefer consistency, so made them static. the reason for the combination is that these are copy-pasted from elsewhere (but i did not want to expose them to global namespace..). updated one, forgot the other.

akva2:
I went for L2Integrand (base class), L2ProbIntegrand and L2FuncIntegrand.

akva2:

akva2:

src/ASM/GlbL2projector.C
```diff
@@ -33,6 +34,102 @@ LinAlg::MatrixType GlbL2::MatrixType   = LinAlg::SPARSE;
 LinSolParams*      GlbL2::SolverParams = nullptr;
 
 
+static void expandTensorGrid2 (const RealArray* in, RealArray* out)
```
kmokstad, line 37: Fine. But you may want to add some small doxy for the these static functions? Or else I assume some jenkins check will complain about it.

src/ASM/GlbL2projector.C
```diff
@@ -33,6 +34,102 @@ LinAlg::MatrixType GlbL2::MatrixType   = LinAlg::SPARSE;
...
+    for (j = 0; j < in[1].size(); j++)
+      for (i = 0; i < in[0].size(); i++, ip++) {
+        out[0][ip] = in[0][i];
+        out[1][ip] = in[1][j];
+        out[2][ip] = in[2][k];
+      }
+}
+
+
+L2ProbIntegrand::L2ProbIntegrand (const ASMbase& ASM, const IntegrandBase& itg) :
```
kmokstad, line 68: I prefer to use `patch` or `pch` as variable name her for the AMSbase object. ASM is already defined as a namespace elsewhere and could be potentially confusing in the doxy processing (or only in my head...)

src/ASM/GlbL2projector.C
```diff
@@ -33,6 +34,102 @@ LinAlg::MatrixType GlbL2::MatrixType   = LinAlg::SPARSE;
...
+}
+
+
+int L2ProbIntegrand::dim () const
+{
+  return m_itg.getNoFields(2);
+}
+
+
+L2FuncIntegrand::L2FuncIntegrand (const ASMbase& ASM, const FunctionBase& func) :
```
kmokstad, line 86: same here

src/ASM/GlbL2projector.C
```diff
@@ -33,6 +34,102 @@ LinAlg::MatrixType GlbL2::MatrixType   = LinAlg::SPARSE;
...
+    std::vector<Real> vals = m_func.getValue(X);
+    for (size_t j = 1; j <= m_func.dim(); ++j)
+      sField(j,i+1) = vals[j-1];
+  }
+
+  return true;
+}
+
+
+int L2FuncIntegrand::dim () const
```
kmokstad, line 127: `size_t`as return type? (since FunctionBase::dim() returns size_t)

src/ASM/GlbL2projector.h
```diff
@@ -29,6 +30,75 @@ typedef std::vector<int>           IntVec;      //!< Vector of integers
...
 
+/*!
+  \brief Abstract class for evaluating integrand or function.
+*/
+
+class L2Integrand {
+public:
+  //! \brief Constructor.
+  //! \param ASM ASM holding geometry to evaluate on
+  L2Integrand(const ASMbase& ASM) : m_asm(ASM) {}
```
kmokstad, line 41: Here also, I don't think i like ASM as variable name, suggest `patch` or `pch`and then use `myPatch` (or `m_patch` if you prefer) as name on the `const ASMbase&` member

src/ASM/GlbL2projector.h
```diff
@@ -29,6 +30,75 @@ typedef std::vector<int>           IntVec;      //!< Vector of integers
...
+
+  //! \brief Evaluates the entity in a set of points.
+  //! \param sField Matrix with results
+  //! \param gpar Points to evaluate in
+  virtual bool evaluate(Matrix& sField, const RealArray* gpar) const = 0;
+
+  //! \brief Returns dimension of entity to evaluate.
+  virtual int dim() const = 0;
```
kmokstad, line 49: size_t as return type

src/ASM/GlbL2projector.h
```diff
@@ -29,6 +30,75 @@ typedef std::vector<int>           IntVec;      //!< Vector of integers
...
+/*!
+  \brief Evaluation class for secondary solutions of an integrand.
+*/
+
+class L2ProbIntegrand : public L2Integrand {
+public:
+  //! \brief Constructor.
+  //! \param ASM ASM holding geometry to evaluate on
+  //! \param itg Integrand to evaluate
+  L2ProbIntegrand(const ASMbase& ASM, const IntegrandBase& itg);
```
kmokstad, line 64: same is in base class

src/ASM/GlbL2projector.h
```diff
@@ -29,6 +30,75 @@ typedef std::vector<int>           IntVec;      //!< Vector of integers
...
+
+  //! \brief Evaluates the secondary solutions in a set of points.
+  //! \param sField Matrix with results
+  //! \param gpar Points to evaluate in
+  bool evaluate(Matrix& sField, const RealArray* gpar) const override;
+
+  //! \brief Returns number of secondary solutions.
+  int dim() const override;
```
kmokstad, line 72: size_t

src/ASM/GlbL2projector.h
```diff
@@ -29,6 +30,75 @@ typedef std::vector<int>           IntVec;      //!< Vector of integers
...
+/*!
+  \brief Evaluation class for functions.
+*/
+
+class L2FuncIntegrand : public L2Integrand {
+public:
+  //! \brief Constructor.
+  //! \param ASM ASM holding geometry to evaluate on
+  //! \param func Function to evaluate
+  L2FuncIntegrand(const ASMbase& ASM, const FunctionBase& func);
```
kmokstad, line 88: same

src/ASM/GlbL2projector.h
```diff
@@ -29,6 +30,75 @@ typedef std::vector<int>           IntVec;      //!< Vector of integers
...
+
+  //! \brief Evaluates the function in a set of points.
+  //! \param sField Matrix with results
+  //! \param gpar Points to evaluate in
+  bool evaluate(Matrix& sField, const RealArray* gpar) const override;
+
+  //! \brief Returns number of function components.
+  int dim() const override;
```
kmokstad, line 96: size_t

kmokstad (approved):
Went through all now and spotted a few minor things to consider. I don't have full overview of the implications but trust that is is consistent and don't break anything. So you are free to merge at any convenience after resolving/rejecting the smaller things

src/ASM/GlbL2projector.C
akva2: yeah, copied from originals.

src/ASM/GlbL2projector.C
akva2: patch and m_patch.

src/ASM/GlbL2projector.C
akva2: changed to size_t

akva2:

akva2:

PR #455, opened 2021-02-08
changed: make isRefined an int

akva2:
256 refinements can happen, in particular in fracturedynamics.
this leads to overflow and a very confusing error message.


akva2:

kmokstad (approved):
OK, I did not foresee that, when the variable was changed from a bool in 75b4689faf7ae3d5598fca77a8af44dc3d7e7d8d

PR #458, opened 2021-03-15
VTF geometry block admin change

kmokstad:
Using a class member variable that is incremented each time a geometry block is added, instead of trying to provide a valid partID, which turns out difficult with a variety of geometry block types around. (Element blocks, points for tractions, additional geometry, etc.) Unfortunately, it affects some downstream apps, so be on alert..


kmokstad:
jenkins build this with downstreams ifem-poisson=43 ifem-elasticity=110 ifem-finitedeformation=26 ifem-stokes=162 ifem-fsi=48 please

PR #460, opened 2021-03-25
Fixed: Account for boundary MPC couplings when choosing stripe direction

kmokstad:
With this the K-joint model with rigid couplings at the ends also works when using multi-threading.


kmokstad:

kmokstad:

PR #464, opened 2021-04-08
Changed: Let IntegrandBase::parse() return false by default

kmokstad:
To ensure consistent behavior when parsing app-specific input


kmokstad:
jenkins build this with downstreams ifem-elasticity=111 ifem-fsi=50 ifem-beamex=24 please

PR #465, opened 2021-04-12
fixed: need to copy projBasis as well

akva2:


kmokstad (approved):
Approved, but amend similar fix for ASMu3D as well.

akva2:

PR #467, opened 2021-04-27
Added: Alternative superelement file format

kmokstad:
Supports the syntax proposed by DnV-GL for the reduced-order-model matrix files.
Also a missing `#include <numeric>` causing build failure on Ubuntu 20.04.


kmokstad:

PR #468, opened 2021-04-27
draft for SSV

akva2:


src/Utility/DataExporter.h
```diff
@@ -192,6 +192,12 @@ class DataWriter
   //! \param[in] level Level we just wrote to the file
   virtual void closeFile(int level) = 0;
 
+  //! \brief Prepare writer for writing a time level.
```
kmokstad, line 195: Doxy for the parameters?

src/Utility/HDF5WriterSSV.C
```diff
@@ -0,0 +1,89 @@
...
+                             bool geometryUpdated, const std::string& prefix)
+{
+  if (geometryUpdated) {
+    H5Fclose(m_file);
+    m_file = -1;
+    openFile(0);
+    level = 0;
+  }
+  if (level == 0)
+      this->writeSIM2(0, entry, geometryUpdated, prefix, true);
```
kmokstad, line 79: wrong indent

src/Utility/HDF5WriterSSV.h
```diff
@@ -0,0 +1,67 @@
...
+  HDF5WriterSSV(const std::string& name, const ProcessAdm& adm);
+
+  //! \brief Destructor.
+  virtual ~HDF5WriterSSV();
+
+  //! \brief Returns the last time level stored in the HDF5 file.
+  virtual int getLastTimeLevel();
+
+  virtual bool prepare(int level, const DataEntry& entry,
+                       bool geometryUpdated, const std::string& prefix);
```
kmokstad, line 44: MIssing doxy

src/Utility/DataExporter.h
```diff
@@ -192,6 +192,12 @@ class DataWriter
...
+  //! \details Used by SVV where we have to preallocate datasets
+  virtual bool prepare(int level, const DataEntry& entry,
+                       bool geometryUpdated, const std::string& prefix)
+  { return true; }
```
kmokstad, line 199: Think the return value can be void instead, since the only implementation always returns true, and the return value is not checked either. Or is it for future extensions?

src/Utility/DataExporter.C
```diff
@@ -125,8 +125,15 @@ bool DataExporter::dumpTimeLevel (const TimeStep* tp, bool geometryUpdated)
...
 
   std::map<std::string,FileEntry>::iterator it;
+
   for (DataWriter* writer : m_writers) {
     writer->openFile(m_level);
+    for (it = m_entry.begin(); it != m_entry.end(); ++it) {
+      if (it->second.field == SIM)
+        if (writeData)
+          writer->prepare(m_level,*it,geometryUpdated,it->second.prefix);
+    }
```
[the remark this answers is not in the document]
akva2: it's not looping over writers here, it's looping over registered SIMs. we need to preallocate for *all* sims before we start writing any data in the ssv writer. so yeah, the separate loop is kind of the whole point of this.

src/Utility/DataExporter.h
akva2: this, and probably all of the currently void methods should ideally be made bool and return values checked. i'll add checking instead.

src/Utility/DataExporter.C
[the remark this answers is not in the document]
kmokstad: sure (was early in the morning), please disregard

src/Utility/DataExporter.h
akva2: added

src/Utility/HDF5WriterSSV.C
akva2: fixed

src/Utility/HDF5WriterSSV.h
akva2: added

src/Utility/DataExporter.h
akva2: done.

akva2:
all comments addressed. i also drafted how to handle the want ssv, but no viz, hdf5 file.

kmokstad:
The most relevant scenario is perhaps: We want both ssv and viz-files,where viz-file contains "all" results wheres ssv typically contains the most interesting field only (one chosen scalar field + deformations when relevant).

akva2:
sure, that's trivial to add when we need it, assuming the most interesting fields means 'primary solutions'.

akva2:
there we go, always only outputting primary solution to ssv file.

PR #469, opened 2021-04-28
simplification in DataExporter

akva2:
- remove unused members
- use ranged for
- remove unnecessary if (we return early if writeData is false)


akva2:

kmokstad:
Included in #474 which now is merged.

PR #474, opened 2021-05-07
Changed: Set initial time level for restart runs from previous

kmokstad:
The first commit here is a polished version of #469.

Then a proposed solution to the mismatching time level flags on hdf5-files for restart and visualization in restart runs.
Now we use the time level specified for the restart as the initial time level for the DataExporter, instead of always starting from 0.
Perhaps a similar approach should be used from HDF5Restart as well (there istep and stride are used to compute the time level, which will yield incorrect results if the stride is changed in a restart, not an unusual scenario).


kmokstad:

PR #479, opened 2021-06-16
Fixed: Use a separate MADOF array in extractPatchSolution()

kmokstad:
...when extraordinary nodes and/or monolithic coupled simulators.

Not sure about side-effects here so, so be alert. Could also be the addMADOF() method should have been used.


kmokstad:

akva2 (approved):

kmokstad:

PR #489, opened 2021-08-10
added: empty protected constructor in FunctionSum

akva2:
to give flexibility in subclasses

### DIFF
--- a/src/SIM/FunctionSum.C
+++ b/src/SIM/FunctionSum.C
@@ -15,16 +15,15 @@
 #include "matrix.h"
 
 
-FunctionSum::FunctionSum (FunctionBase* f, double w)
-{
-  comps.push_back(std::make_pair(f,w));
-  ncmp = f->dim();
-}
-
-
 bool FunctionSum::add (FunctionBase* f, double w)
 {
-  if (f->dim() == comps.front().first->dim())
+  if (comps.empty())
+  {
+    comps.push_back(std::make_pair(f,w));
+    ncmp = f->dim();
+    return true;
+  }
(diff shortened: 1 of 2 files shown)


src/SIM/FunctionSum.h
```diff
@@ -51,6 +51,10 @@ class FunctionSum : public FunctionBase
   //! \brief Returns a representative scalar equivalent of the function value.
   virtual double getScalarValue(const Vec3& X) const;
 
+protected:
+  //! \brief Constructor to allow empty initialization in subclasses.
+  FunctionSum() {}
+
```
kmokstad, line 57: It's OK, but move L54-L57 before L28, such that both constructors are declared first (kinda IFEM standard)?

src/SIM/FunctionSum.C
```diff
@@ -17,14 +17,19 @@
 
 FunctionSum::FunctionSum (FunctionBase* f, double w)
 {
-  comps.push_back(std::make_pair(f,w));
-  ncmp = f->dim();
+  this->add(f,w);
```
kmokstad, line 20: Consider to inline this constructor in the h-file instead, since now only a singleliner:
```
FunctionSum(FunctionBase* f, double w = 1.0) { this->add(f,w); }
```

src/SIM/FunctionSum.h
akva2: done

src/SIM/FunctionSum.C
akva2: done.

akva2:
jenkins build this with downstreams ifem-darcy=26 please

PR #492, opened 2021-08-13
Add support for matrix/vector dumps in matrix-market format

akva2:
Matrix-market format is useful for testing as it is supported by various linalg packages (in my case amgcl).

### DIFF
--- a/src/LinAlg/DiagMatrix.C
+++ b/src/LinAlg/DiagMatrix.C
@@ -30,19 +30,18 @@ void DiagMatrix::initAssembly (const SAM& sam, bool)
 }
 
 
-void DiagMatrix::dump (std::ostream& os, char format, const char* label)
+void DiagMatrix::dump (std::ostream& os, LinAlg::StorageFormat format, const char* label)
 {
   switch (format)
-    {
-    case 'M':
-    case 'm':
+  {
+    case LinAlg::MATLAB:
       utl::writeMatlab(label,myMat,os);
       break;
 
-    default:
+    case LinAlg::FLAT:
       if (label) os << label <<" =";
       os << myMat;
-    }
+  }
 }
 
 
--- a/src/LinAlg/DiagMatrix.C
+++ b/src/LinAlg/DiagMatrix.C
@@ -38,6 +38,15 @@ void DiagMatrix::dump (std::ostream& os, LinAlg::StorageFormat format, const cha
       utl::writeMatlab(label,myMat,os);
       break;
 
+    case LinAlg::MATRIX_MARKET:
+      os << "%%MatrixMarket matrix coordinate real general\n";
+      if (label)
+        os << "% label = " << label << '\n';
+      os << myMat.size() << ' ' << myMat.size() << ' ' << myMat.size();
+      for (size_t row = 1; row <= myMat.size(); ++row)
+        os << '\n' << row << ' ' << row << ' ' << myMat(row);
+      break;
+
     case LinAlg::FLAT:
       if (label) os << label <<" =";
       os << myMat;
--- a/src/LinAlg/SparseMatrix.C
+++ b/src/LinAlg/SparseMatrix.C
@@ -344,30 +344,51 @@ const Real& SparseMatrix::operator () (size_t r, size_t c) const
 
 void SparseMatrix::dump (std::ostream& os, LinAlg::StorageFormat format, const char* label)
 {
-  if (label) os << label <<" = [\n";
+  if (label && format != LinAlg::MATRIX_MARKET) os << label <<" = [\n";
+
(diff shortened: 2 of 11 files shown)


akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

src/LinAlg/DiagMatrix.C
```diff
@@ -38,6 +38,17 @@ void DiagMatrix::dump (std::ostream& os, LinAlg::StorageFormat format, const cha
       utl::writeMatlab(label,myMat,os);
       break;
 
+    case LinAlg::MATRIX_MARKET:
+    {
```
kmokstad, line 42: Not necessary with braces around this case block (since no local variable definitions within it). To be consistent with the other case-clocks and also for the similar added block for DenseMatrix above.

src/LinAlg/SparseMatrix.C
```diff
@@ -344,31 +344,45 @@ const Real& SparseMatrix::operator () (size_t r, size_t c) const
 
 void SparseMatrix::dump (std::ostream& os, LinAlg::StorageFormat format, const char* label)
 {
-  if (label) os << label <<" = [\n";
+  if (label && format != LinAlg::MATRIX_MARKET) os << label <<" = [\n";
+  std::string trail = ";\n";
```
kmokstad, line 348: I did not see the need for introducing the `trail`variable here, except for slightly added complexity It is only for the new matrix format it is changed to '\n' such that the hard-coding ";\n" should still work fine for the Matlab format.

akva2: it's supposed to be "];\n" - with the bracket. so it's a bug.

src/LinAlg/DiagMatrix.C
akva2: removed.

akva2:
jenkins build this please

src/SIM/SIMinput.C
```diff
@@ -678,7 +678,10 @@ bool SIMinput::parseOutputTag (const TiXmlElement* elem)
     utl::getAttribute(elem,"step",steps);
     utl::getAttribute(elem,"eps",dmp.eps);
     dmp.step.insert(steps.begin(),steps.end());
-    dmp.format = format[0];
+    if (format == "matlab")
+      dmp.format = LinAlg::MATLAB;
+    else if (format[0] == 'M' || format[0] == 'm')
+      dmp.format = LinAlg::MATLAB;
```
kmokstad, line 686: This could be simplified to
```
  if (toupper(format[0]) == 'M')
    dmp.format = LinAlg::MATLAB;
```
will cover all cases above

akva2: i do not want to cover all cases cause it would also match matrix_market. i don't like one-letter magic codes, but i need to keep it for backwards compatibility. i can change to use toupper though (old code)

src/LinAlg/SparseMatrix.C
kmokstad: Right, I missed there was no break at the end of that block. Then it is clear. (this case fall-through is sometimes hard to spot, and the newer compilers warns against it also, for the good or bad depending on taste).

src/SIM/SIMinput.C
kmokstad: I agree, it is bad magic, and since the new format added in the subsequent commit also starts with 'm'. Then forget the backward- compatibility. The very few others that dumps the system matrices can take that change I believe (may consider at "unknown matrix format  (ignored)" warning instead"

akva2:
jenkins build this please

src/SIM/SIMinput.C
```diff
@@ -680,8 +680,12 @@ bool SIMinput::parseOutputTag (const TiXmlElement* elem)
     dmp.step.insert(steps.begin(),steps.end());
     if (format == "matlab")
       dmp.format = LinAlg::MATLAB;
-    else if (std::toupper(format[0]) == 'M')
-      dmp.format = LinAlg::MATLAB;
+    else if (format == "matrix_market")
+      dmp.format = LinAlg::MATRIX_MARKET;
+    else {
+      std::cerr << "** SIMinput::parseOutputTag: Unknown matrix dump format \"" << format << "\", ignored" << std::endl;
```
kmokstad, line 686: The IFEM standard is: Two spaces + two *-characters as prefix on warning messages to cerr (or cout).

akva2: i never get this right..

akva2:
jenkins build this please

kmokstad (approved):

PR #497, opened 2021-09-10
Fixes for warnings emitted by new clang/gcc versions

akva2:
- const for keys
- std::move to avoid some copies
- a missing using statement


akva2:
jenkins build this with downstreams ifem-elasticity=126 ifem-fsi=57 ifem-openfrac=68 ifem-thermoelasticity=21 ifem-thm=14 please

kmokstad (approved):
Looks fine :-)

PR #501, opened 2021-09-21
fixed: allow flat dump format

akva2:


akva2:

kmokstad (approved):
Alternatively, one could accept `format == "flat"` and/or `fornmat == "plain"`, etc. before giving the error.

akva2:
yeah, i was uncertain whether we want to force a format attribute in that case or not. if you think it's better, i can do that.

akva2:

PR #508, opened 2021-10-04
fixed: have to cast to int

akva2:
if not the shift makes little sense as everything is done
for an unsigned char.

this caused problems with clang.


akva2:

kmokstad (commented):
Shifting by 16 is the same as multiplying with 2^16, so why not just write the simpler `basis*65536` instead, or would that also cause issues. If so, I think a better solution would be to change the argument types of these two methods to (unsigned) int, such that the casting implicitly take place in the function call?

akva2:
while shifting more clearly conveys the intention (in terms of use 17 bits for basis, the rest for components), it's completely acceptable to use multiplication to have the cast happen implicitly. changed to *65536

PR #509, opened 2021-10-08
Modified Newton-Raphson for linear Newmark time integration

kmokstad:
This (mainly the third and fourth commit) implements the option to do Newmark time integration for linear problems without recalculating the LHS matrix for every step/iteration. Thereby speeding up the simulation.
Activate by setting (the new) parameter nupdat=0 in `<newmarksolver>` context.

The fourth commit provides an interface for telling the integrand to buffer up its element LHS matrices instead of recalculating them, when they are supposed to be unchanged, and the last commit was triggered by the need of passing the global element index to the integrand through the `finalizeElement` method.

### DIFF
--- a/src/ASM/ASMs2D.C
+++ b/src/ASM/ASMs2D.C
@@ -37,9 +37,6 @@
 #include "MPC.h"
 #include "IFEM.h"
 #include <array>
-#ifdef USE_OPENMP
-#include <omp.h>
-#endif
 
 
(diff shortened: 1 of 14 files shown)


akva2 (approved):
Apart from doxy issues this looks good to me.

kmokstad:
jenkins build this with downstreams ifem-elasticity=133 please

PR #511, opened 2021-10-13
Fixed: eval2ndSolution should do nothing if opt.pSolOnly is true

kmokstad:
Changed: Return true if myVtf is null in all the writeGlv methods.
Then there is no need to check for domain decomposition and procId>0
since myVtf will always be null in that case.
Removed evalProjSolution, use writeGlvP with iStep=0 instead.

These changes are needed to ensure consistent max-value calculation when hdf-export is activated and/or vtf export with stride.


kmokstad:
jenkins build this with downstreams ifem-elasticity=134 please

PR #518, opened 2021-11-03
Changed: Use std::vector<double> as reaction force container

kmokstad:
Only to make clear this is not a `Vector` subjected to algebraic operations.


kmokstad:
jenkins build this with downstreams ifem-finitedeformation=31 please

PR #520, opened 2021-11-12
added: allow for defining a function for textureproperties

akva2:
this allows for a generic f(texture(X)) where f can be scalar or vector
valued.

ideally f would be a function of property name but
as we do not have a base class for f: R -> R^3 we have to use the
f: R^3 -> R^3 type for vector functions. for this reason, both
scalar and vectorial functions have to refer to the texture value as 'x'.

### DIFF
--- a/src/Utility/TextureProperties.C
+++ b/src/Utility/TextureProperties.C
@@ -13,6 +13,7 @@
 
 #include "TextureProperties.h"
 #include "IFEM.h"
+#include "Functions.h"
 #include "HDF5Reader.h"
 #include "ProcessAdm.h"
 #include "Utilities.h"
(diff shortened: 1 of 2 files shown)


akva2:

akva2:

kmokstad:
Regarding "as we do not have a base class for f: R -> R^3", you can do, e.g., ` typedef utl::Function<Real,Vec3> Vec3Func;` to get Vec3Func a base class for such functions. See for instance [TractionField.h](https://github.com/OPM/IFEM/blob/f4d4c7247aa4e4e858e0aac28a531ac38e7928c8/src/Utility/TractionField.h#L142) where I use that to represent a time-dependent vector field.

akva2:
yeah, i realize that it's doable but i'd have to implement a ExprFunc version of it.

PR #522, opened 2021-12-01
fixed: forgotten condition in 3D subgrid with H mesh projection

akva2:
meant projection was done on h/2 grid after adaptation.


akva2:
jenkins build this with downstreams ifem-stokes=188 please

PR #529, opened 2022-01-27
re-enable multithreaded mixed 3D assembly

akva2:
the problems leading to it being disabled should have been fixed now

### DIFF
--- a/src/ASM/ASMs3Dmx.C
+++ b/src/ASM/ASMs3Dmx.C
@@ -1249,15 +1249,6 @@ void ASMs3Dmx::generateThreadGroups (const Integrand& integrand, bool silence,
 }
 
 
-void ASMs3Dmx::generateThreadGroups (char lIndex, bool silence, bool)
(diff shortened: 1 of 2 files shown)


akva2:

src/ASM/ASMs3Dmx.C
```diff
@@ -1251,9 +1251,6 @@ void ASMs3Dmx::generateThreadGroups (const Integrand& integrand, bool silence,
 
 void ASMs3Dmx::generateThreadGroups (char lIndex, bool silence, bool)
 {
-#ifdef USE_OPENMP
-  omp_set_num_threads(1);
-#endif
```
kmokstad, line 1256: But then you could just remove the overriding and just use the parent class method, since it be just the same anyway?

akva2: yeah, right you are.

akva2:

PR #531, opened 2022-02-07
Added: Extract internal forces from ReactionsOnly integral

kmokstad:
For calculation of interface forces between patches, instead of using the surface integral which is not that accurate

### DIFF
--- a/src/ASM/ReactionsOnly.h
+++ b/src/ASM/ReactionsOnly.h
@@ -36,7 +36,9 @@ public:
   //! \param[in] rf The reaction force vector to be assembled
   //! \param[in] sam Data for FE assembly management
   //! \param[in] adm Parallell processing administrator
-  ReactionsOnly(Vector& rf, const SAM* sam, const ProcessAdm& adm);
+  //! \param[in] sf Internal force vector to be assembled
+  ReactionsOnly(Vector& rf, const SAM* sam, const ProcessAdm& adm,
+                Vector* sf = nullptr);
   //! \brief Empty destructor.
   virtual ~ReactionsOnly() {}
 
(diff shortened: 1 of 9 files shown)


src/ASM/ReactionsOnly.h
```diff
@@ -36,7 +36,9 @@ class ReactionsOnly : public GlobalIntegral
   //! \param[in] rf The reaction force vector to be assembled
   //! \param[in] sam Data for FE assembly management
   //! \param[in] adm Parallell processing administrator
-  ReactionsOnly(Vector& rf, const SAM* sam, const ProcessAdm& adm);
+  //! \param[in] sf Internal force vector to be assembled
+  ReactionsOnly(Vector& rf, const SAM* sam, const ProcessAdm& adm,
+                Vector* sf = NULL);
```
akva2, line 41: nullptr

kmokstad: Fixed. Also a similar case in SIMbase.

kmokstad:
jenkins build this with downstreams ifem-elasticity=138 ifem-poisson=50 please

src/ASM/ASMs3D.C
```diff
@@ -1720,45 +1663,118 @@ void ASMs3D::getBoundaryNodes (int lIndex, IntVec& nodes,
...
+  int node = this->findStartNode(n1,n2,n3,basis);
+  if (node < 1)
+    return;
+  else if (local)
+    node = 1;
+
+  if (swapW && lEdge <= 8) // Account for swapped parameter direction
+    lEdge += (lEdge-1)%4 < 2 ? 2 : -2;
+
+  std::pair<int,int> range;
```
akva2, line 1739: std::array<int,2> seems more natural to me.

akva2 (approved):

kmokstad:
jenkins build this with downstreams ifem-elasticity=138 ifem-poisson=50 please

PR #532, opened 2022-02-07
changed: print iteration history in NonLinSIM even at msgLevel -1

akva2:
allows suppressing norm and solution summary output, while still
keeping the iteration history. useful in split simulators
such as Chorin

### DIFF
--- a/src/SIM/NonLinSIM.C
+++ b/src/SIM/NonLinSIM.C
@@ -434,7 +434,7 @@ ConvStatus NonLinSIM::checkConvergence (TimeStep& param)
   if (param.iter > 1 && prevNorm > 0.0 && fabs(norm) > prevNorm*0.1)
     status = SLOW;
 
-  if (msgLevel > 0)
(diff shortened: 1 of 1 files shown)


src/SIM/NonLinSIM.C
```diff
@@ -434,7 +434,7 @@ ConvStatus NonLinSIM::checkConvergence (TimeStep& param)
   if (param.iter > 1 && prevNorm > 0.0 && fabs(norm) > prevNorm*0.1)
     status = SLOW;
 
-  if (msgLevel > 0)
+  if (msgLevel >= -1)
```
kmokstad, line 437: I would prefer `if (msgLevel > 0 || msgLevel == -1)` such that `msgLevel == 0` still retains silence.

akva2: done.

akva2:

kmokstad (approved):

PR #533, opened 2022-02-09
fixed: we need to free the MPI_Comms even in serial

akva2:
I guess at some point we did not dupe comms in serial, but we do now.


akva2:

akva2:
this is a fix and in mpi code which is my turf -> merging.

PR #534, opened 2022-02-10
added: extend element matrix cache support to mixed

akva2:
And add necessary tracking of assembled status in BlockElmMats.

### DIFF
--- a/src/ASM/GlbL2projector.h
+++ b/src/ASM/GlbL2projector.h
@@ -147,14 +147,21 @@ public:
                            LocalIntegral& elmInt);
   //! \brief Initializes current element for numerical integration (mixed).
   //! \param[in] MNPC1 Matrix of nodal point correspondance for current element
+  //! \param[in] fe Nodal and integration point data for current element
   //! \param[in] elem_sizes Size of each basis on the element
   //! \param[in] basis_sizes Size of each basis on the patch
   //! \param elmInt Local integral for element
   virtual bool initElement(const IntVec& MNPC1,
+                           const MxFiniteElement& fe,
                            const uIntVec& elem_sizes,
                            const uIntVec& basis_sizes,
                            LocalIntegral& elmInt);
 
+  //! \brief Dummy implementation.
+  virtual bool initElement(const IntVec& MNPC1,
+                           const uIntVec& elem_sizes,
+                           const uIntVec& basis_sizes,
+                           LocalIntegral& elmInt) { return false; }
   //! \brief Dummy implementation.
   virtual bool initElement(const IntVec&, LocalIntegral&) { return false; }
   //! \brief Dummy implementation.
--- a/src/ASM/IntegrandBase.C
+++ b/src/ASM/IntegrandBase.C
@@ -159,6 +159,16 @@ bool IntegrandBase::initElement (const std::vector<int>& MNPC,
 }
 
 
+bool IntegrandBase::initElement (const std::vector<int>& MNPC,
+                                 const MxFiniteElement&,
+                                 const std::vector<size_t>& elem_sizes,
+                                 const std::vector<size_t>& basis_sizes,
+                                 LocalIntegral& elmInt)
+{
+  return this->initElement(MNPC, elem_sizes, basis_sizes, elmInt);
+}
+
+
 /*!
   The default implementation extracts the element-level vector only for the
   first (current) primary solution vector.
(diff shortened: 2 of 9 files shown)


akva2:

src/ASM/GlbL2projector.h
```diff
@@ -147,15 +147,22 @@ class GlbL2 : public Integrand
...
+                           const MxFiniteElement& fe,
                            const uIntVec& elem_sizes,
                            const uIntVec& basis_sizes,
                            LocalIntegral& elmInt);
 
   //! \brief Dummy implementation.
+  virtual bool initElement(const IntVec& MNPC1,
+                           const uIntVec& elem_sizes,
+                           const uIntVec& basis_sizes,
+                           LocalIntegral& elmInt) { return false; }
```
kmokstad, line 164: Is this dummy strictly necessary if you changed all the `initElement` calls in the ASM*mx* methods?

src/ASM/IntegrandBase.C
```diff
@@ -159,6 +159,16 @@ bool IntegrandBase::initElement (const std::vector<int>& MNPC,
 }
 
 
+bool IntegrandBase::initElement (const std::vector<int>& MNPC,
+                                 const MxFiniteElement& fe,
```
kmokstad, line 163: `fe` is not used here

src/ASM/IntegrandBase.h
```diff
@@ -359,6 +370,14 @@ class NormBase : public Integrand
                            const std::vector<size_t>& elem_sizes,
                            const std::vector<size_t>& basis_sizes,
                            LocalIntegral& elmInt);
+  //! \brief Initializes current element for numerical integration (mixed).
+  virtual bool initElement(const std::vector<int>& MNPC,
+                           const MxFiniteElement& fe,
```
kmokstad, line 375: `fe` is unused (will give compiler warning?)

src/ASM/BlockElmMats.C
```diff
@@ -168,12 +171,16 @@ const Matrix& BlockElmMats::getNewtonMatrix () const
   std::cout <<"Resulting Newton matrix"<< A.front();
 #endif
 
+  newtonMatAssembled = true;
```
kmokstad, line 174: So where are the bool members reset to false again when assembling a new system? Ideally, that should be handled my a member method as well.

kmokstad (commented):
Guess it's fine and all downstreams have been verified. However, as far I can see, only the element index is used in the sub-class implementations of the mixed initElement methods, so an alternative could be to just pass that as parameter and not the whole MxFiniteElement object.

src/ASM/GlbL2projector.h
akva2: yes. GlbL2projector inherits Integrand only, where this is pure virtual.

src/ASM/IntegrandBase.C
akva2: made anonymous.

src/ASM/IntegrandBase.h
akva2: made anonymous.

src/ASM/BlockElmMats.C
akva2: it's not really necessary as we never reuse LocaIntegrals. I can add it of course, but it will be completely unused.

akva2:
As for only passing the iel, I stuck to what you did for non-mixed, which was to pass the whole fe object.

src/ASM/BlockElmMats.C
akva2: maybe overriding redim is sufficient?

src/ASM/GlbL2projector.h
kmokstad: OK, the alternative would be to remove the Integrand member also, but then all other downstreams using mixed integrands would be affected. Guess you want to avoid that..

src/ASM/BlockElmMats.C
kmokstad: I see,its fine as is then, although I don't see the big picture why this is needed since the LocaIntegral objects are deleted after each element in any case.

kmokstad:
Yep, just stick with passing the fe object, it will be consistent with the other (non-mixed) version

src/ASM/BlockElmMats.C
akva2: when we assemble the rhs vector using the element matrices, getNewtonMatrix() is called, https://github.com/OPM/IFEM/blob/master/src/ASM/AlgEqSystem.C#L171

which would overwrite the cached matrix set in initElement (unless we instead cache all the blocks which is not very convenient).

akva2:

PR #535, opened 2022-02-11
fixed: wrong variable used in 3D mixed tensor spline fields

akva2:
woopsie.


akva2:

akva2:
i have added regression test so this is under testing.

kmokstad (approved):

PR #536, opened 2022-02-14
Fix SuperLU option

akva2:
if we have previously reused the matrix factorization,
the FACTORED value is stored. this leads to segfaults.
we now properly flag that we only want to reuse the
sparsity pattern.


akva2:

kmokstad (approved):
Looks fine, although I have no clue about consequences of setting this options.

PR #537, opened 2022-02-16
changed: do not resize the newton matrix up front

akva2:
this way we can use its non-emptyness to signal that
it has already been assembled.


kmokstad (approved):
Yep, I rather prefer this way instead of the #534 approach

akva2:
jenkins build this with downstreams ifem-boussinesq=73 ifem-darcy=45 ifem-navierstokes=173 ifem-poroelasticity=55 ifem-stokes=207 ifem-thm=17 please

PR #538, opened 2022-02-16
changed: suppress all output in RK solver if not standalone

akva2:
This to aid prettifying in user code.


akva2:

PR #540, opened 2022-02-18
Boundarynodes fix

kmokstad:
Split out from #531 which will be rebased later.

This only deals with the boundary node extraction for a given property code. Now it also account for 1D boundaries in 3D, and the node uniqueness check is applied for all dimensions - not only for ldim+1==nsd. Notice also the mod'ing (by 10) of the topology index is removed for edge boundaries in 3D. This will give wrong results since the index may be in range 1 to 12.


kmokstad:

akva2 (approved):
Note minor comment in other PR

kmokstad:

PR #547, opened 2022-04-20
fixed: check number of components in function

akva2:
in particular for mixed integrands using anasol boundary conditions
nf can be greater than the number of components in the function. this is
because nf is set to the total number of fields and not the number
of fields on a single basis


akva2:

akva2:

akva2:
Merging without min().

PR #548, opened 2022-04-22
fixed: clear additional functions in clearProperties

akva2:
this is necessary to not register multiple functions if
used in adaptive simulations

### DIFF
--- a/src/SIM/SIMoutput.C
+++ b/src/SIM/SIMoutput.C
@@ -52,6 +52,9 @@ SIMoutput::~SIMoutput ()
 void SIMoutput::clearProperties ()
 {
   myPoints.clear();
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this with downstreams ifem-boussinesq=84 please

akva2:

kmokstad (approved):
Yep, this would be a memory leak when changing grids. The only concern is if someone uses `addAddFunc` with a pointer to a function which is deallocated locally (or not allocated dynamically at all), it will crash. But as far I can see it is only used in SIMStokes where I assume you have full control.

akva2:
ownership is already taken by this class (it does the delete in the destructor) so that's nothing new.

PR #549, opened 2022-04-22
changed: add 'symmetric' as an alias for 'symmetrized' beta type

akva2:
Users keep getting this wrong. Having the alias is harmless..


akva2:

PR #550, opened 2022-04-25
fixed: memory leak in mixed LR integrands

akva2:
tensorspline was not cleared up after LR bases
had been established


akva2:

kmokstad (approved):

PR #554, opened 2022-05-06
Some fixes for Subgrid

akva2:
- fix wrong projection basis with tensor splines
- remove support for subgridH. this is not correct, we need to use two separate projection bases.

### DIFF
--- a/src/ASM/ASMs2Dmx.C
+++ b/src/ASM/ASMs2Dmx.C
@@ -203,8 +203,11 @@ bool ASMs2Dmx::generateFEMTopology ()
 
     // we need to project on something that is not one of our bases
     if (ASMmxBase::Type == ASMmxBase::REDUCED_CONT_RAISE_BASIS1 ||
+        ASMmxBase::Type == ASMmxBase::REDUCED_CONT_RAISE_BASIS2 ||
         ASMmxBase::Type == ASMmxBase::DIV_COMPATIBLE)
       projB = proj = ASMmxBase::raiseBasis(surf);
+    else if (ASMmxBase::Type == ASMmxBase::SUBGRID)
+      projB = proj = m_basis.front()->clone();
     else
       projB = proj = m_basis[2-ASMmxBase::geoBasis]->clone();
   }
--- a/src/ASM/ASMs3Dmx.C
+++ b/src/ASM/ASMs3Dmx.C
@@ -204,8 +204,11 @@ bool ASMs3Dmx::generateFEMTopology ()
 
     // we need to project on something that is not one of our bases
     if (ASMmxBase::Type == ASMmxBase::REDUCED_CONT_RAISE_BASIS1 ||
+        ASMmxBase::Type == ASMmxBase::REDUCED_CONT_RAISE_BASIS2 ||
         ASMmxBase::Type == ASMmxBase::DIV_COMPATIBLE)
       projB = proj = ASMmxBase::raiseBasis(svol);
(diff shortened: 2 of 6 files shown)


akva2:
jenkins build this with downstreams ifem-stokes=217 please

akva2:
jenkins build this with downstreams ifem-stokes=217 please

src/ASM/ASMs2Dmx.C
```diff
@@ -206,7 +206,7 @@ bool ASMs2Dmx::generateFEMTopology ()
         ASMmxBase::Type == ASMmxBase::DIV_COMPATIBLE)
       projB = proj = ASMmxBase::raiseBasis(surf);
     else
-      projB = proj = m_basis[2-ASMmxBase::geoBasis]->clone();
+      projB = proj = m_basis[0]->clone();
```
kmokstad, line 209: or `.front()` (instead of `[0]`), like it was before d8d4d40d8ff9435a6971779913b6431135f568b7. So that fix (from Oct 2021) is no longer valid since reverted here? Or are there other cases when it still apply.

src/ASM/ASMs3Dmx.C
```diff
@@ -207,7 +207,7 @@ bool ASMs3Dmx::generateFEMTopology ()
         ASMmxBase::Type == ASMmxBase::DIV_COMPATIBLE)
       projB = proj = ASMmxBase::raiseBasis(svol);
     else
-      projB = proj = m_basis[2-ASMmxBase::geoBasis]->clone();
+      projB = proj = m_basis[0]->clone();
```
kmokstad, line 210: same question as above.

src/ASM/LR/ASMu2Dmx.C
```diff
@@ -1011,8 +1011,7 @@ bool ASMu2Dmx::refine (const LR::RefineData& prm, Vectors& sol)
     // Uniformly refine to find basis 1
     if (ASMmxBase::Type == ASMmxBase::SUBGRID) {
       m_basis[0].reset(refBasis->copy());
-      if (!ASMmxBase::subgridH)
-        projBasis = m_basis[0];
+      projBasis = m_basis[0];
```
kmokstad, line 1014: Be consistent at least (either `[0]` or `.front()`, not a "salig blanding"), for the readability.

src/ASM/LR/ASMu3Dmx.C
```diff
@@ -970,8 +970,7 @@ bool ASMu3Dmx::refine (const LR::RefineData& prm, Vectors& sol)
     // Uniformly refine to find basis 1
     if (ASMmxBase::Type == ASMmxBase::SUBGRID) {
       m_basis[0].reset(refBasis->copy());
-      if (!ASMmxBase::subgridH)
-        projBasis = m_basis[0];
+      projBasis = m_basis[0];
```
kmokstad, line 973: Here also, suggest using `.front()` for consistency, whenever referring to the first basis, which is a special case. Only when looping I use the index operator (probably a underlying hunch that index-operator may in some times be marginally more costly since it needs to deal with array bounds, at least in debug mode)

kmokstad (approved):
..with some comments/suggestions, mostly of cosmetics kind.

src/ASM/ASMs2Dmx.C
akva2: the fix was partially broken, as in it fixes what is was supposed to fix for FRTH but it broke subgrid in the process. and now i almost broke FRTH while fixing subgrid.. too focused on problems at hand, not enough on the world

PR #558, opened 2022-05-11
Some updates to make unbutu jammy (22.04) happy

akva2:
- bump c++ to c++17 and do so through standard cmake mechanisms instead of homegrow FindCXX11.cmake
- fix support for dune-istl > 2.6
- quell various cmake warnings emitted in the newer cmake in jammy


akva2:

PR #562, opened 2022-05-19
added: proper preallocation for block matrices with graph partitioning

akva2:
Also avoid some memory usage for unused arrays (G2LEQ for blocks).


akva2:

PR #563, opened 2022-05-19
added: set parameters in SplineUtils::project

akva2:
Need this when projecting a field function.


akva2:

PR #566, opened 2022-05-19
fix initial conditions with partitioning

akva2:
calculate these on root node and broadcast instead of calculating
separately on each process. this can cause desync between processes,
for instance if the initial condition uses random numbers


akva2:

akva2:
jenkins build this with downstreams ifem-boussinesq=92 ifem-advectiondiffusion=74 ifem-stokes=226 ifem-navierstokes=188 please

akva2:

PR #569, opened 2022-05-23
Call initPatch in SIMbase::project

akva2:
Needed when projecting a multi-patch field function.


akva2:

akva2:

PR #570, opened 2022-06-02
draft of basis function cache

akva2:
This is a draft so we can discuss how to do this.

In largeish 3D simulations the patch-global gauss point cache of basis function values becomes a memory bottleneck.
In particular this is a problem when running with graph partitioning, since all processes will calculate all values across the patch,
not limited to the element it actually has to process.

For this reason I have draft this cache layer. It allows for several modes of operation;
1) Policy::PRE_CACHE: previous behavior, calculating for all gauss points on the patch up front and clearing when integrate ends
2) Policy::FULL_CACHE: retaining the gauss-point cache between calls to integrate (memory intensive but saves time)
3) Policy::ON_THE_FLY: Like 2) but we build the cache on first call. In particular this only builds for the elements we are integrating on this process, see above
4) Policy::NO_CACHE: completely disabling caching

I have only implemented it for ASMs3D yet, but the idea would be to use it for all ASMs where relevant (3D struct and LR, maybe 2D).


src/ASM/ASMs3D.C
```diff
@@ -1978,52 +1978,21 @@ bool ASMs3D::integrate (Integrand& integrand,
...
-  std::array<const double*,3> xg, wg;
-  for (int d = 0; d < 3; d++)
-  {
-    ng[d] = this->getNoGaussPt(svol->order(d));
-    xg[d] = GaussQuadrature::getCoord(ng[d]);
-    wg[d] = GaussQuadrature::getWeight(ng[d]);
-    if (!xg[d] || !wg[d]) return false;
-  }
+  if (myCache.empty())
+    myCache.emplace_back(std::make_unique<BasisFunctionCache>(*this, integrand, 1));
```
kmokstad, line 1982: What is emblace_back doing here? Same as push_back? If not, I lost the logic...
Also, the make_unique calls for some explanation. All these newer C++ construct I'm not updated with.

src/ASM/ASMs3D.C
```diff
@@ -2100,21 +2069,21 @@ bool ASMs3D::integrate (Integrand& integrand,
...
                 // Fetch basis function derivatives at current integration point
-                SplineUtils::extractBasis(spline[ip],fe.N,dNdu);
+                const BasisFunctionVals& bfs = cache.getVals(iel-1,ip);
 
                 // Compute Jacobian determinant of coordinate mapping
                 // and multiply by weight of current integration point
-                double detJac = utl::Jacobian(Jac,fe.dNdX,Xnod,dNdu,false);
-                double weight = dV*wg[0][i]*wg[1][j]*wg[2][k];
+                double detJac = utl::Jacobian(Jac,fe.dNdX,Xnod,bfs.dNdu,false);
+                double weight = dV*wr[i]*wr[j]*wr[k];
```
kmokstad, line 2083: This looks suspicious, when comparing with old code, with different integration order in different directions (not much used, but still..)

src/ASM/ASMs3D.C
```diff
@@ -3791,3 +3759,165 @@ bool ASMs3D::addRigidCpl (int lindx, int ldim, int basis,
...
   return this->ASMstruct::addRigidCpl(lindx,ldim,basis,gMaster,Xmaster,extraPt);
 }
+
+
+ASMs3D::BasisFunctionCache::BasisFunctionCache (const ASMs3D& pch,
+                                                const Integrand& itg,
+                                                int b) :
+  patch(pch), integrand(itg), basis(b)
+{
+
```
kmokstad, line 3769: no blank line here

src/ASM/ASMs3D.C
```diff
@@ -3791,3 +3759,165 @@ bool ASMs3D::addRigidCpl (int lindx, int ldim, int basis,
...
+std::array<size_t, 3> ASMs3D::BasisFunctionCache::elmIndex (size_t el) const
+{
+  std::array<size_t, 3> result;
+
+  result[0] = el % nel1;
+  result[1] = (el / nel1) % nel2;
+  result[2] = el / (nel1*nel2);
+
+  return result;
```
kmokstad, line 3877: could shorten this to just `return { el % nel1, (el/nel1) % nel2, el / (nel1*nel2) };` ?

src/ASM/ASMs3D.C
```diff
@@ -3791,3 +3759,165 @@ bool ASMs3D::addRigidCpl (int lindx, int ldim, int basis,
...
+std::array<size_t, 3> ASMs3D::BasisFunctionCache::gpIndex (size_t gp, bool reduced) const
+{
+  const Quadrature& q = reduced ? reducedQ : mainQ;
+  std::array<size_t, 3> result;
+
+  result[0] = gp % q.ng[0];
+  result[1] = (gp / q.ng[0]) % q.ng[1];
+  result[2] = gp/ (q.ng[0]*q.ng[1]);
+
+  return result;
```
kmokstad, line 3890: same, consider one-liner implementation

kmokstad (commented):
I like the general idea, and with the option to to as before, as well as the newer "calculate only when needed option". Did not go into details of the implementation (I trust it will work as intended), only some comments here and there. Not sure I like too much this smart pointer management used here (and other places). I know you like them ;-) but personally I like to control the memory management explicitly to make the logic less fuzzy (in my eyes).

src/ASM/ASMs3D.C
akva2: emplace_back is used to construct in-place. with a push_back you first construct an object, and then copy that into the vector element. with emplace_back() this copy is avoided. make_unique creates a unique pointer in-place, avoiding default construction followed by assignment from bare pointer. so the sum effect is that an unique pointer is placed at the back of the vector without any copying.

src/ASM/ASMs3D.C
akva2: nice catch. bad copy-paste

src/ASM/ASMs3D.C
akva2: sure

src/ASM/ASMs3D.C
akva2: Yes. Strictly speaking this is not a shared_ptr, this is a unique_ptr. They are not copyable, only a single reference can exist, hence the requirement to create in-place.

PR #572, opened 2022-08-01
changed: facilitate reuse of ForceBase derived integrands

akva2:
previously there was no way to clear the buffers used during
integration. now initBuffer resizes only if necessary, but
always zeros the underlying buffer.

closes https://github.com/OPM/IFEM-Darcy/issues/51


akva2:
jenkins build this with downstreams ifem-costa=21 please

kmokstad (approved):

PR #574, opened 2022-09-13
Quell some cmake warnings

akva2:
Ubuntu 22.04 warns some more. let's quell them.


akva2:

PR #575, opened 2022-10-03
fixed: periodic single patch models with multiple bases

akva2:
If we pass dirs=123 (the default) with less fields, stuff meets fan.

### DIFF
--- a/src/Utility/Utilities.C
+++ b/src/Utility/Utilities.C
@@ -445,6 +445,16 @@ std::set<int> utl::getDigits (int num)
 }
 
 
+int utl::getDirs (int ncmp)
+{
+  int res = 0, mul = 1;
+  for (int i = ncmp; i >= 1; --i, mul *= 10)
+    res += i*mul;
+
+  return res;
+}
+
+
 namespace utl
 {
   /*!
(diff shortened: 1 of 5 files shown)


akva2:

akva2:

src/Utility/Utilities.C
```diff
@@ -445,6 +445,16 @@ std::set<int> utl::getDigits (int num)
 }
 
 
+int utl::getDirs (int ncmp)
+{
+  int res = 0, mul = std::pow(10,ncmp-1);
+  for (int i = 1; i <= ncmp; ++i, mul /= 10)
+    res += i*mul;
```
kmokstad, line 452: Sub-sub optimizing, but this would result in the same without the pow-function(?)
```
  int res = 0, mul = 1;
  for (int i = ncmp; i >= 1; i--, mul *= 10)
    res *= i*mul;
```

kmokstad (approved):
This looks fine, only with one suggestion for "improvement"

src/Utility/Utilities.C
akva2: yes. at the cost of some readability (imo). but changed.

akva2:

PR #578, opened 2022-10-11
Quell some warnings

akva2:
- Annoying bitwise instead of logical warnings
- One wrong return value

### DIFF
--- a/src/SIM/ModelGenerator.C
+++ b/src/SIM/ModelGenerator.C
@@ -33,8 +33,9 @@ static std::string parseKnots (const TiXmlElement* geo, int num)
   std::stringstream str;
   for (int i = 0; i < num; ++i, ++min[0], ++max[0]) {
     double pmin = 0.0, pmax = 1.0;
-    if (utl::getAttribute(geo,min.c_str(),pmin) |
-        utl::getAttribute(geo,max.c_str(),pmax))
+    bool parsed = utl::getAttribute(geo,min.c_str(),pmin);
+    parsed &= utl::getAttribute(geo,max.c_str(),pmax);
+    if (parsed)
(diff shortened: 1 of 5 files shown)


akva2:
jenkins build this please

src/SIM/ModelGenerator.C
```diff
@@ -33,8 +33,9 @@ static std::string parseKnots (const TiXmlElement* geo, int num)
   std::stringstream str;
   for (int i = 0; i < num; ++i, ++min[0], ++max[0]) {
     double pmin = 0.0, pmax = 1.0;
-    if (utl::getAttribute(geo,min.c_str(),pmin) |
-        utl::getAttribute(geo,max.c_str(),pmax))
+    bool parsed = utl::getAttribute(geo,min.c_str(),pmin);
+    parsed &= utl::getAttribute(geo,max.c_str(),pmax);
```
kmokstad, line 37: Think maybe you need  `|=` here to retain exactly the same logic?

src/SIM/SIMinput.C
```diff
@@ -740,7 +740,7 @@ FunctionBase* SIMinput::parseDualTag (const TiXmlElement* elem, int ftype)
     ftype = 3;
     depth = (Xd.second-Xd.first).length();
   }
-  else if (utl::getAttribute(elem,"u0",xi[0])|
+  else if (static_cast<int>(utl::getAttribute(elem,"u0",xi[0]))|
```
kmokstad, line 743: Maybe a better solution would be to change the return value of `utl::getAttribute()` to `int`, so this casting would not be needed. And the above ModelGenerator changes would not be needed either (an possibly other/future similarities).

PR #579, opened 2022-10-19
Added: Optional element shrinkage factor for LR meshes to VTF

kmokstad:
This is needed for models with internal C^{-1} lines, to properly visualize the discontinuity with deformed geometry.
Without it, one often picks the wrong nodal results along the discontinuity.


kmokstad:

PR #582, opened 2023-01-12
fixed: allow system CFLAGS in petsc pkg config

### LINKED ISSUE #581: Suggestion for FindPetsc.cmake
The FindPetsc.cmake Module currently depends on pkg_check_modules returning PETSC_INCLUDE_DIRS. However, pkg-config deduplicates Paths and strips the include path if it is a default include path or a path already contained in the CFLAGS and similar environment variables when calling pkg-config. In HPC centers using environment modules with LMod or similar, this is very often the case. That leads to PETSC_INCLUDE_DIRS being empty and PETSc thus being disabled, although it is found. The behaviour can be circumvented by defining 

set(ENV{PKG_CONFIG_ALLOW_SYSTEM_CFLAGS} 1)

It could be set and unset similar to the PKG_CONFIG path before. 

Greetings
André

akva2:
these can end up empty if petsc is installed to system paths and stripping is enabled

Closes #581


akva2:

PR #584, opened 2023-05-03
Fix external load tratment for HHTSIM

kmokstad:
The important thing here is the third commit. Without it, the external forces (when resulting from internal and neumann integrals - not for direct point loads) are added twice to the residual force vector, since it is also done in `HHTSIM::finalizeRHSvector()`. The other two commits are for debug print handling and added unit test for `NewmarkNLSIM`.


kmokstad:
jenkins build this with downstreams ifem-elasticity=144 ifem-beamex=32 ifem-finitedeformation=36 please

PR #588, opened 2023-05-22
fixed: use internal basis evaluation in jump integration method

akva2:
to handle NURBS


akva2:

PR #589, opened 2023-05-22
changed: unify computeBasis signatures

akva2:
always take a spline pointer parameter. use reference in the NURBS interfaces.


akva2:

kmokstad (approved):
Looks fine as well, although I did not see the (good) reason for changing from pointer to reference on the spline argument for the computeBasisNurbs methods

kmokstad:
and I assume the checks failure is unrelated

akva2:
it's mostly my coding principles. using a pointer signals that null is an option and it's not an option in those methods. failure is the unit tests failing with clang, i'll rerun the build now.

PR #592, opened 2023-05-23
Fix rational projection basis in ASMu2D

akva2:
Manually specified projection basis can also be rational.


akva2:

src/ASM/LR/ASMu2D.h
```diff
@@ -643,6 +643,9 @@ class ASMu2D : public ASMLRSpline, public ASM2D
   //! \brief Converts current tensor spline object to LR-spline.
   LR::LRSplineSurface* createLRfromTensor();
 
+  //! \brief Converts a rational spline surfce to a LR NURBS surface.
+  LR::LRSplineSurface* createLRNurbs(const Go::SplineSurface& srf);
```
kmokstad, line 647: At least this method can be const (or even better; static), since it does not access any of the class members - only the argument. And check spelling in the doxy "surfce".

akva2: true, i will make it static. i need to add some more stuff to really make this usable first though so i'll work on that first.

akva2: okay, fixed.

akva2:

kmokstad (approved):

PR #593, opened 2023-05-23
Add support for NURBS in LRSplineField(s)2D

akva2:
And needed refactorings.

PR #594, opened 2023-05-24
ASMu2Dmx: add NURBS support

akva2:
Sits on top of https://github.com/OPM/IFEM/pull/593


akva2:
jenkins build this with downstreams ifem-stokes=230 please

akva2:
jenkins build this with downstreams ifem-stokes=230 please

src/ASM/LR/ASMu2Dmx.C
```diff
@@ -212,7 +218,7 @@ bool ASMu2Dmx::generateFEMTopology ()
     SurfaceVec svec = ASMmxBase::establishBases(tensorspline, ASMmxBase::Type);
     m_basis.resize(svec.size());
     for (size_t b = 0; b < svec.size(); b++)
-      m_basis[b].reset(new LR::LRSplineSurface(svec[b].get()));
+    m_basis[b].reset(createLR(*svec[b]));
```
kmokstad, line 221: wrong indent

src/ASM/LR/ASMu2Dmx.C
```diff
@@ -201,9 +201,15 @@ bool ASMu2Dmx::generateFEMTopology ()
...
+  auto createLR = [this](Go::SplineSurface& srf)
+                  {
+                    return srf.rational() ? this->createLRNurbs(srf)
+                                          : new LR::LRSplineSurface(&srf);
+                  };
+
   if (tensorPrjBas)
   {
-    projBasis.reset(new LR::LRSplineSurface(tensorPrjBas));
+    projBasis.reset(createLR(*tensorPrjBas));
```
kmokstad, line 212: Here you first dereference the tensorPrjBas pointer, before you (inside the lambda) use & to extract the pointer again (back and forth), wouldn't in this case at least be better to let the lambda accept the pointer argument instead? Since it is local for this method. In spite of your policy of using reference whenever the pointer can't be NULL.

kmokstad (approved):
with one question and (only) one indent error to fix

src/ASM/LR/ASMu2Dmx.C
akva2: I did initially, but it comes at the cost of .get() for the shared_ptrs. So i went with the least visual noise, deref and one ref (for the evil lr constructor that takes a ptr where null makes no sense 😁)

akva2: And lr even takes a mutable pointer hence i cant pass a const ref :/

src/ASM/LR/ASMu2Dmx.C
akva2: fixed.

akva2:
jenkins build this with downstreams ifem-stokes=230 please

PR #595, opened 2023-05-30
Some cleanup of mixed LR

kmokstad:
This originates from #397 more than 3 years ago, and contains some cleanup (including fix of potential issues) of the ASMu[23]mx code, but is not directly related to the scope of that PR. Motivation: Less code-duplication, and easiear to compare 2D and 3D implementations (diff).


kmokstad:

PR #596, opened 2023-08-09
fixed: support establishing RT basis for NURBS

akva2:
Need to deal with the weights properly when raising the order of the bases.

Currently unused as RT can only be used with Piola but logically separate so I open the PR early.


akva2:

akva2:
done better in https://github.com/OPM/IFEM/pull/629

PR #598, opened 2023-08-21
Allow for geometry definition separate from bases

akva2:
This allows for a geometry that is separate from solution bases.
In particular this will be needed for proper RT support where we need the geometry to quadratic for the lowest-order RT basis (ie where the solution bases are (2,1) x (1,2) x (1,1) we need geometry to be (2,2)).

Very drafty still, only done structured 2D ASMs and only handling first order derivatives, ie there is no hessian handling yet. I'm opening early to get feedback on how I have approached it.

### DIFF
--- a/src/ASM/ASMs2D.C
+++ b/src/ASM/ASMs2D.C
@@ -89,6 +89,12 @@ ASMs2D::~ASMs2D ()
 }
 
 
+const Go::SplineSurface* ASMs2D::getGeometry () const
+{
+  return static_cast<const Go::SplineSurface*>(geomB);
+}
+
+
 Go::SplineCurve* ASMs2D::getBoundary (int dir, int)
 {
   if (dir < -2 || dir == 0 || dir > 2)
@@ -115,7 +121,7 @@ void ASMs2D::copyParameterDomain (const ASMbase* other)
 bool ASMs2D::read (std::istream& is, int)
 {
   if (shareFE) return true;
-  if (surf) delete surf;
+  delete surf;
 
   Go::ObjectHeader head;
   surf = new Go::SplineSurface;
@@ -154,7 +160,11 @@ bool ASMs2D::read (std::istream& is, int)
     nsd = surf->dimension();
   }
 
-  geomB = surf;
+  if (ASMbase::refineGeometry)
+    geomB = surf;
+  else
+    geomB = surf->clone();
+
   return true;
 }
 
@@ -177,6 +187,7 @@ void ASMs2D::clear (bool retainGeometry)
     // Erase spline data
     if (proj && proj != surf) delete proj;
     if (surf && !shareFE) delete surf;
+    if (geomB != surf) delete geomB;
     geomB = projB = surf = proj = nullptr;
   }
 
@@ -359,6 +370,8 @@ bool ASMs2D::checkRightHandSystem ()
   // This patch has a negative Jacobian determinant. Probably it is modelled
   // in a left-hand-system. Swap the v-parameter direction to correct for this.
   surf->reverseParameterDirection(false);
+  if (geomB != surf)
+    static_cast<Go::SplineSurface*>(geomB)->reverseParameterDirection(false);
   return swapV = true;
 }
 
@@ -952,7 +965,7 @@ size_t ASMs2D::constrainEdgeLocal (int dir, bool open, int dof, int code,
     // Compute the surface normal at this edge point.
     // That will be the local z-axis of the local coordinate system.
     u[tdir] = gpar[i];
-    surf->point(pts,u[0],u[1],1);
+    static_cast<Go::SplineSurface*>(geomB)->point(pts,u[0],u[1],1);
     Vec3 Zaxis(SplineUtils::toVec3(pts[1],nsd),SplineUtils::toVec3(pts[2],nsd));
     gdata[k+3] = Zaxis.x;
     gdata[k+4] = Zaxis.y;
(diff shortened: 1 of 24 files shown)


src/ASM/FiniteElement.C
```diff
@@ -80,15 +80,17 @@ bool MxFiniteElement::Jacobian (Matrix& Jac, const Matrix& Xnod,
...
-                         gBasis > 1 ? dMdX[gBasis-2] : dNdX, Xnod,
-                         bf ? (*bf)[gBasis-1]->dNdu : (*dNxdu)[gBasis-1]);
-  if (detJxW == 0.0) return false; // singular point
+  size_t nBasis = bf ? bf->size() : dNxdu->size();
+  if (gBasis <= nBasis) {
+    detJxW = utl::Jacobian(Jac,
+                           gBasis > 1 ? dMdX[gBasis-2] : dNdX, Xnod,
+                           bf ? (*bf)[gBasis-1]->dNdu : (*dNxdu)[gBasis-1]);
+    if (detJxW == 0.0) return false; // singular point
+  }
```
kmokstad, line 89: Wouldn't it make more sense to just do `if (gBasis > nBasis) return false;` here? The calculations below will otherwise be meaningless since the `Jac` then is not set, or do you make the assumption it has the correct value as input (dangerous). Not quite sure what the intention is here.

src/ASM/ASMs2D.C
```diff
@@ -3362,26 +3436,29 @@ void ASMs2D::BasisFunctionCache::calculateAll ()
...
+    const Go::SplineSurface* surf = basis == 0 ? patch.getGeometry() : patch.getBasis(basis);
+    surf->computeBasisGrid(par1,par2,spline);
     size_t idx = 0;
     for (const Go::BasisDerivsSf& spl : spline) {
       SplineUtils::extractBasis(spl,values[idx].N,values[idx].dNdu);
       ++idx;
     }
   };
-  if (nderiv == 1)
+  if (nderiv == 1) {
```
kmokstad, line 3447: you insist on braces around single-statements?

src/ASM/ASMs2D.C
```diff
@@ -3362,26 +3436,29 @@ void ASMs2D::BasisFunctionCache::calculateAll ()
...
     extract1(mainQ->gpar[0],mainQ->gpar[1],values);
-  else if (nderiv == 2) {
+  } else if (nderiv == 2) {
     std::vector<Go::BasisDerivsSf2> spline;
-    patch.getBasis(basis)->computeBasisGrid(mainQ->gpar[0],mainQ->gpar[1],spline);
+    const Go::SplineSurface* surf = basis == 0 ? patch.getGeometry() : patch.getBasis(basis);
+    surf->computeBasisGrid(mainQ->gpar[0],mainQ->gpar[1],spline);
```
kmokstad, line 3452: The local scope variable surf here may conflict with member with same name (at least on newer compilers)?
An alternative could be to redefine the `getBasis()` method to return the value of `getGeometry()` when the argument is zero, if that always holds. Then the change here would not be needed at all.

src/ASM/ASMs2D.C
```diff
@@ -3362,26 +3436,29 @@ void ASMs2D::BasisFunctionCache::calculateAll ()
...
     size_t idx = 0;
     for (const Go::BasisDerivsSf2& spl : spline) {
       SplineUtils::extractBasis(spl,values[idx].N,values[idx].dNdu,values[idx].d2Ndu2);
       ++idx;
     }
   } else if (nderiv == 3) {
     std::vector<Go::BasisDerivsSf3> spline;
-    patch.getBasis(basis)->computeBasisGrid(mainQ->gpar[0],mainQ->gpar[1],spline);
+    const Go::SplineSurface* surf = basis == 0 ? patch.getGeometry() : patch.getBasis(basis);
+    surf->computeBasisGrid(mainQ->gpar[0],mainQ->gpar[1],spline);
```
kmokstad, line 3461: same comment as above

src/ASM/ASMs2D.C
```diff
@@ -3314,22 +3387,23 @@ BasisFunctionVals ASMs2D::BasisFunctionCache::calculatePt (size_t el,
   std::array<size_t,2> elIdx = this->elmIndex(el);
 
   BasisFunctionVals result;
+  const Go::SplineSurface* surf = basis == 0 ? patch.getGeometry() : patch.getBasis(basis);
```
kmokstad, line 3390: Same comment as on L3451 (see below).

src/ASM/ASMs2Dmx.C
```diff
@@ -222,7 +222,10 @@ bool ASMs2Dmx::generateFEMTopology ()
       return false; // Logic error
   }
   delete surf;
-  geomB = surf = m_basis[elmBasis-1]->clone();
+  const Go::SplineSurface* surfOld = surf;
```
kmokstad, line 225: Maybe a `bool geoAndItgBasisAreSame = surf == geomB;` instead of copying the pointer would be more readable? That variable could also be a class member as I assume it will be rather constant (but the name is up for discussion).

Or can `ASMbase::refineGeometry` be used instead?

src/ASM/ASMs2D.C
```diff
@@ -89,6 +89,12 @@ ASMs2D::~ASMs2D ()
 }
 
 
+const Go::SplineSurface* ASMs2D::getGeometry () const
+{
+  return static_cast<const Go::SplineSurface*>(geomB);
+}
```
kmokstad, line 95: Consider merging this method with `getBasis()` (see comments further down).

kmokstad (commented):
Added some preliminary comments. Could some are only due to not understanding the full logic here. It becomes quite complicated now with so many different bases flouting around ;)

On the firsts commit, maybe `intBasis` or `itgBasis` is a even more precise name than `elmBais` which I feel is a bit round...

PR #601, opened 2023-08-24
Some deduplication in ASMsxDmx

akva2:
Introduce a method to obtain last node on integration basis and use this to deduplicate the methods that obtains parametric area/volume/length. Likewise introduce a method to obtain first node on integration basis and use this to deduplicate getElementCoordinates().

### DIFF
--- a/src/ASM/ASMs2D.C
+++ b/src/ASM/ASMs2D.C
@@ -1336,10 +1336,11 @@ bool ASMs2D::getElementCoordinates (Matrix& X, int iel) const
 
   X.resize(nsd,surf->order_u()*surf->order_v());
 
+  int lnod0 = this->getFirstItgElmNode();
   RealArray::const_iterator cit = surf->coefs_begin();
   for (size_t n = 0; n < X.cols(); n++)
   {
-    int ip = this->coeffInd(MNPC[iel-1][n])*surf->dimension();
+    int ip = this->coeffInd(MNPC[iel-1][n + lnod0])*surf->dimension();
     if (ip < 0) return false;
 
     for (size_t i = 0; i < nsd; i++)
--- a/src/ASM/ASMs2D.C
+++ b/src/ASM/ASMs2D.C
@@ -1233,7 +1233,7 @@ double ASMs2D::getParametricArea (int iel) const
   if (MNPC[iel-1].empty())
     return 0.0;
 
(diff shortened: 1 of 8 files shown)


akva2:

akva2:

src/ASM/ASMs2Dmx.C
```diff
@@ -1162,6 +1126,12 @@ void ASMs2Dmx::swapProjectionBasis ()
 }
 
 
+int ASMs2Dmx::getFirstItgElmNode () const
+{
+  return std::accumulate(elem_size.begin(), elem_size.begin() + geoBasis-1, 0);
+}
+
+
```
kmokstad, line 1134: This implementation (as well as for getLastItgElmNode below) could also have been places in the common base ASMmxBase, since they only use members of that class, and the 3D version is identical. But you still need to forward to that method from here though (consider if worthwhile).

kmokstad (approved):
yes, looks great, although it took me half a second extra to realize that "deduplication" was the opposite of "duplication" (is there a better word perhaps?).

Also, I would suggest to squash the four commits into two; one for the get-first methods, and one for the get-last methods.

src/ASM/ASMs2D.C
```diff
@@ -1274,13 +1274,13 @@ double ASMs2D::getParametricLength (int iel, int dir) const
...
 
   switch (dir)
-    {
-    case 1: return surf->knotSpan(0,nodeInd[inod1].I);
-    case 2: return surf->knotSpan(1,nodeInd[inod1].J);
-    }
+  {
+  case 1: return surf->knotSpan(0,nodeInd[inod1].I);
+  case 2: return surf->knotSpan(1,nodeInd[inod1].J);
+  }
```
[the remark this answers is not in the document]
akva2: uhm, all i did was nuke the tabs, but i guess i forgot to set the project tabstop to 8 instead of the default 4 so the autoindenter got confused. fixed.

src/ASM/ASMs2Dmx.C
akva2: yeah i know but i decided it was just an extra jump with little gain.

akva2:
deduplication is the proper english word so i can't do much about that. squashed.

akva2:

PR #602, opened 2023-08-25
fixed: make sure to always set geoBasis

akva2:
if not we risk getting the wrong setting if we instance a SIM with e.g. subgrid followed by instancing a SIM with Taylor-Hood.


akva2:

PR #603, opened 2023-08-25
Changed: Second (mostly dummy) argument in read method is removed.

kmokstad:
Instead, introducing a separate `readBasis()` method.
Consider if this simplifies the logic, and does not break any future intentions.


kmokstad:

akva2:
this is fine and makes things a bit clearer. read() is now for the geometry from which we derive the bases, while readBasis() is used for the special case of serialization.

kmokstad:
jenkins build this with downstreams simra-postproc=12 please

akva2:
You have write access in downstream now.

PR #604, opened 2023-08-28
Unify geometry pointer in ASMLRSpline with ASMstruct

akva2:
This is stored as geomB in ASMstruct. Do the same in ASMLRSpline for consistency (and to facilitate further upcoming changes).

### DIFF
--- a/src/ASM/LR/ASMu2DC1.C
+++ b/src/ASM/LR/ASMu2DC1.C
@@ -18,13 +18,14 @@
 
 bool ASMu2DC1::generateFEMTopology ()
 {
-  if (!(geo = this->createLRfromTensor()))
+  geomB = this->createLRfromTensor();
+  if (!geomB)
     return false;
-  else if (geo->order(0) > 2 || geo->order(1) > 2)
+  else if (geomB->order(0) > 2 || geomB->order(1) > 2)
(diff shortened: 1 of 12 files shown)


akva2:

src/ASM/LR/ASMu2DC1.C
```diff
@@ -18,7 +18,8 @@
 
 bool ASMu2DC1::generateFEMTopology ()
 {
-  if (!(geomB = this->createLRfromTensor()))
+    geomB = std::shared_ptr<LR::LRSpline>(this->createLRfromTensor());
+    if (!geomB)
```
kmokstad, line 22: Still wrong indentation in L21,22..
And since the return type of `createLRfromTensor()` is already changed to a `shared_ptr`, do you need to pass it through its copy constructor again?

kmokstad (approved):
Since the first commit is only the renaming, and the second commit is change of type of the same variable, I feel these two commits are better off squashed into a single commit.

src/ASM/LR/ASMu2DC1.C
akva2: fixed, and no. and squashed.

akva2:

PR #605, opened 2023-08-28
ASMsxD: move projection bases into ASMstruct

akva2:
Make this consistent with other bases. This introduces some more casts than comfortable but they will go away later.

### DIFF
--- a/src/ASM/ASMs2Dmx.C
+++ b/src/ASM/ASMs2Dmx.C
@@ -44,12 +44,6 @@ ASMs2Dmx::ASMs2Dmx (const ASMs2Dmx& patch, const CharVec& n_f)
 }
 
 
-ASMs2Dmx::~ASMs2Dmx ()
-{
-  delete altProjBasis;
-}
-
-
 Go::SplineSurface* ASMs2Dmx::getBasis (int basis) const
 {
   if (basis < 1 || basis > (int)m_basis.size())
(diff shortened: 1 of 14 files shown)


akva2:

src/ASM/ASMs2Dmx.C
```diff
@@ -96,7 +96,7 @@ bool ASMs2Dmx::readBasis (std::istream& is, size_t basis)
 bool ASMs2Dmx::write (std::ostream& os, int basis) const
 {
   if (basis == -1)
-    os <<"200 1 0 0\n" << *proj;
+    os <<"200 1 0 0\n" << *static_cast<const Go::SplineSurface*>(projB);
```
kmokstad, line 99: No need to cast here, since the `operator<<`  is defined in the base class `Go::Streamable`?

src/ASM/ASMs3Dmx.C
```diff
@@ -97,7 +97,7 @@ bool ASMs3Dmx::readBasis (std::istream& is, size_t basis)
 bool ASMs3Dmx::write (std::ostream& os, int basis) const
 {
   if (basis == -1)
-    os <<"700 1 0 0\n" << *proj;
+    os <<"700 1 0 0\n" << *static_cast<const Go::SplineVolume*>(projB);
```
kmokstad, line 100: No need to cast here, since the operator<< is defined in the base class Go::Streamable?

kmokstad (approved):
Yes, this will lessen the confusion on the projection basis handling a bit. But consider that name change.

src/ASM/ASMstruct.h
```diff
@@ -109,6 +109,7 @@ class ASMstruct : public ASMbase
 protected:
   Go::GeomObject* geomB; //!< Pointer to spline object of the geometry basis
   Go::GeomObject* projB; //!< Pointer to spline object of the projection basis
+  Go::GeomObject* altProjB; //!< Pointer to spline object of the alternative projection basis
```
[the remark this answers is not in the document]
akva2: good point. renamed to projB2.

src/ASM/ASMs3Dmx.C
akva2: indeed.

akva2:

PR #607, opened 2023-08-29
rename some variables to make code more readable

akva2:
and reuse getFirstItgElmNode while at it


akva2:

akva2:
In some contexts that is not the number of element nodes (multipliers etc). But yeah, we can add it, but will do it elsewhere.

PR #608, opened 2023-08-29
ASMuxD: move projection and refinement bases into ASMLRSpline

akva2:
LR variant of #605

PR #609, opened 2023-08-29
Override separateProjectionBasis()

akva2:
This to fix a bug with two-field formulations where the second basis is the higher order basis, and to prepare for having a separate geometry basis.

Sits on top of https://github.com/OPM/IFEM/pull/608


akva2:

kmokstad (approved):
Looks good (assuming the running tests pass)

PR #611, opened 2023-08-29
Extended getBasis

akva2:
This extends the getBasis method in ASMs to be able to return more than just the finite element bases.

### DIFF
--- a/src/ASM/ASMenums.h
+++ b/src/ASM/ASMenums.h
@@ -46,6 +46,15 @@ namespace ASM //! Assembly scope
     FULL_CACHE  //!< Cache basis function values up front
   };
 
+  //! \brief Enumeration of different basis types.
+  //! \details Entries should have non-positive values
+  enum BasisType {
+    GEOMETRY_BASIS     =  0, //!< Geometry basis
+    PROJECTION_BASIS   = -1, //!< Projection basis
+    PROJECTION_BASIS_2 = -2, //!< Second projection basis
+    REFINEMENT_BASIS   = -3, //!< Refinement basis
+    INTEGRATION_BASIS  = -4, //!< Integration basis
+  };
 }
 
 #endif
--- a/src/ASM/LR/ASMu2D.h
+++ b/src/ASM/LR/ASMu2D.h
@@ -138,10 +138,10 @@ public:
   //! \brief Returns the spline surface representing the geometry of this patch.
   const LR::LRSplineSurface* getSurface() const { return lrspline.get(); }
 
-  //! \brief Returns the spline surface representing the basis of this patch.
-  virtual const LR::LRSplineSurface* getBasis(int = 1) const { return lrspline.get(); }
-  //! \brief Returns the spline surface representing the basis of this patch.
-  virtual LR::LRSplineSurface* getBasis(int = 1) { return lrspline.get(); }
+  //! \brief Returns the spline surface representing a basis of this patch.
+  virtual const LR::LRSplineSurface* getBasis(int basis = 1) const;
+  //! \brief Returns the spline surface representing a basis of this patch.
+  virtual LR::LRSplineSurface* getBasis(int basis = 1);
 
 
   // Methods for model generation and refinement
--- a/src/ASM/LR/ASMu2D.h
+++ b/src/ASM/LR/ASMu2D.h
@@ -133,11 +133,6 @@ public:
   //! \brief Empty destructor.
   virtual ~ASMu2D() { geomB = nullptr; }
 
-  //! \brief Returns the spline surface representing the geometry of this patch.
-  LR::LRSplineSurface* getSurface() { return this->createLRfromTensor().get(); }
-  //! \brief Returns the spline surface representing the geometry of this patch.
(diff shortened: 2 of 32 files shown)


akva2:

src/ASM/LR/ASMu3Dmx.h
```diff
@@ -66,10 +66,7 @@ class ASMu3Dmx : public ASMu3D, private ASMmxBase
   virtual ~ASMu3Dmx() {}
 
   //! \brief Returns the spline surface representing the basis of this patch.
```
kmokstad, line 68: Correct the doxy also, here we are dealing with volumes

src/ASM/LR/ASMu2D.h
```diff
@@ -139,9 +139,7 @@ class ASMu2D : public ASMLRSpline, public ASM2D
   const LR::LRSplineSurface* getSurface() const { return lrspline.get(); }
 
   //! \brief Returns the spline surface representing the basis of this patch.
-  virtual const LR::LRSplineSurface* getBasis(int = 1) const { return lrspline.get(); }
-  //! \brief Returns the spline surface representing the basis of this patch.
-  virtual LR::LRSplineSurface* getBasis(int = 1) { return lrspline.get(); }
+  virtual LR::LRSplineSurface* getBasis(int basis = 1) const;
```
kmokstad: So the const version is not needed, I presume..

src/ASM/SplineFields2D.C
```diff
@@ -26,7 +26,9 @@
 SplineFields2D::SplineFields2D (const ASMs2D* patch,
                                 const RealArray& v, char nbasis,
                                 int nnf, const char* name)
-  : Fields(name), basis(patch->getBasis(nbasis)), surf(patch->getSurface())
+    : Fields(name),
+      basis(patch->getBasis(nbasis)),
+      surf(patch->getBasis(ASM::GEOMETRY_BASIS))
```
kmokstad, line 31: Something happened with the indentation here? compared to SplineField2D above

src/ASM/LR/Test/TestASMu2D.C
```diff
@@ -107,7 +107,7 @@ TEST(TestASMu2D, InterfaceChecker)
   pch->raiseOrder(1,1);
   pch->uniformRefine(0, 3);
   pch->uniformRefine(1, 3);
-  LR::LRSplineSurface* lr = pch->getSurface();
+  LR::LRSplineSurface* lr = pch->getBasis(1);
```
kmokstad, line 110: Could also drop the 1-argument here, since it is the default.

src/ASM/LR/ASMu3D.h
```diff
@@ -116,11 +116,6 @@ class ASMu3D : public ASMLRSpline, public ASM3D
   //! \brief Empty destructor.
   virtual ~ASMu3D() {}
 
-  //! \brief Returns the spline volume representing the geometry of this patch.
-  LR::LRSplineVolume* getVolume() { return this->createLRfromTensor().get(); }
```
kmokstad, line 120: Same comment as for 2D

kmokstad (approved):
OK, now checked all. Looks fine, but I would suggest squashing the first 2 commits and the 4 last ones, such that only 2 remains.

src/ASM/ASMenums.h
```diff
@@ -46,6 +46,15 @@ namespace ASM //! Assembly scope
     FULL_CACHE  //!< Cache basis function values up front
   };
 
+  //! \brief Enumeration of different basis types.
+  //! \details Entries should have non-positive values
+  enum BasisType {
+    GEOMETRY_BASIS     =  0,  //!< Geometry basis
```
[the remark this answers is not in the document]
akva2: Fixed

src/ASM/SplineFields2D.C
akva2: fixed.

src/ASM/LR/ASMu2D.h
```diff
@@ -133,11 +133,6 @@ class ASMu2D : public ASMLRSpline, public ASM2D
   //! \brief Empty destructor.
   virtual ~ASMu2D() { geomB = nullptr; }
 
-  //! \brief Returns the spline surface representing the geometry of this patch.
-  LR::LRSplineSurface* getSurface() { return this->createLRfromTensor().get(); }
```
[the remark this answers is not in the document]
akva2: Actually getBasis(ASM::INTEGRATION_BASIS) is the same as the old getSurface, not ASM::GEOMETRY_BASIS. The non-mutable getBasis does the same as getSurface, ie creates the LR surface if we are called between read and generateFEMTopology(). This only happens in a test or two and they still work so all should be well.

src/ASM/LR/ASMu3D.h
akva2: And same response.

akva2:

src/ASM/LR/ASMu3Dmx.h
akva2: Fixed.

PR #612, opened 2023-08-29
Unify ASMsxDmx and ASMsxD write methods

akva2:
This removes some unnecessary overrides and adds tests for write/getBasis implementations in structured ASMs.

### DIFF
--- a/src/ASM/ASMs2D.C
+++ b/src/ASM/ASMs2D.C
@@ -181,12 +181,14 @@ bool ASMs2D::read (std::istream& is)
 }
 
 
-bool ASMs2D::write (std::ostream& os, int) const
+bool ASMs2D::write (std::ostream& os, int basis) const
 {
   if (!surf) return false;
+  if (basis > static_cast<int>(this->getNoBasis())) return false;
+  const Go::SplineSurface* spline = this->getBasis(basis);
+  if (!spline) return false;
 
-  os <<"200 1 0 0\n";
-  os << *surf;
+  os <<"200 1 0 0\n" << *spline;
 
   return os.good();
 }
--- a/src/ASM/ASMs3D.C
+++ b/src/ASM/ASMs3D.C
@@ -152,12 +152,14 @@ bool ASMs3D::read (std::istream& is)
 }
 
 
-bool ASMs3D::write (std::ostream& os, int) const
+bool ASMs3D::write (std::ostream& os, int basis) const
 {
   if (!svol) return false;
+  if (basis > static_cast<int>(this->getNoBasis())) return false;
+  const Go::SplineVolume* spline = this->getBasis(basis);
+  if (!spline) return false;
 
-  os <<"700 1 0 0\n";
-  os << *svol;
+  os <<"700 1 0 0\n" << *spline;
 
   return os.good();
 }
--- a/src/ASM/Test/TestASMs2Dmx.C
+++ b/src/ASM/Test/TestASMs2Dmx.C
@@ -1,0 +1,297 @@
+//==============================================================================
+//!
+//! \file TestASMs2Dmx.C
+//!
+//! \date Aug 25 2023
+//!
+//! \author Arne Morten Kvarving / SINTEF
+//!
+//! \brief Tests for structured 2D mixed spline FE models.
+//!
+//==============================================================================
+
+#include "ASMmxBase.h"
+#include "ASMSquare.h"
+
+#include "gtest/gtest.h"
+
+
+namespace {
+
+// formatting matches write routine, do not change
+const char* squareTH_1 =
+  "200 1 0 0\n"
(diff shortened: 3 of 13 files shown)


akva2:

src/ASM/ASMs2D.C
```diff
@@ -181,12 +181,15 @@ bool ASMs2D::read (std::istream& is)
...
 
 
-bool ASMs2D::write (std::ostream& os, int) const
+bool ASMs2D::write (std::ostream& os, int basis) const
 {
   if (!surf) return false;
+  if (basis > static_cast<int>(this->getNoBasis())) return false;
+  const Go::SplineSurface* spline = this->getBasis(basis);
+  if (!spline)
+    return false;
```
kmokstad, line 190: Inconsistent style, return on the same line as the if, or on the next line, L187 vs. L189,190

src/ASM/ASMs2D.C
```diff
@@ -181,12 +181,15 @@ bool ASMs2D::read (std::istream& is)
 }
 
 
-bool ASMs2D::write (std::ostream& os, int) const
+bool ASMs2D::write (std::ostream& os, int basis) const
```
kmokstad, line 184: maybe writeBasis would be a better suited name now...

src/ASM/ASMs3D.C
```diff
@@ -152,12 +152,15 @@ bool ASMs3D::read (std::istream& is)
...
 {
   if (!svol) return false;
+  if (basis > static_cast<int>(this->getNoBasis())) return false;
+  const Go::SplineVolume* spline = this->getBasis(basis);
+  if (!spline)
+    return false;
 
-  os <<"700 1 0 0\n";
-  os << *svol;
+  os <<"700 1 0 0\n" << *spline;
```
kmokstad, line 163: Same comments as for the 2D commit apply here as well.

src/ASM/Test/TestASMs2Dmx.C
```diff
@@ -0,0 +1,297 @@
...
+//==============================================================================
+
+#include "ASMmxBase.h"
+#include "ASMSquare.h"
+
+#include "gtest/gtest.h"
+
+
+namespace {
+
```
kmokstad, line 20: what is the effect of this namespace with no name? You refer to its members below as if the namespace was not there. Is it really needed? And if so, why?

src/ASM/Test/TestASMs3Dmx.C
```diff
@@ -0,0 +1,529 @@
...
+//! \brief Tests for structured 3D mixed spline FE models.
+//!
+//==============================================================================
+
+#include "ASMCube.h"
+
+#include "gtest/gtest.h"
+
+
+namespace {
```
kmokstad, line 18: Same namespace question applies here too. There is probably a good reason, I just don't see it.

kmokstad (approved):
Mostly OK, except some cosmetics/questions.

kmokstad:
and yes, the first two commits should be squashed. Don't see any point splitting the same change into two commits for 2D and 3D.

src/ASM/ASMs2D.C
akva2: okay, i guess i have to stick to the one-line ifs (i really dislike them as it makes stepping through the code in a debugger less convenient and it makes human parsing of code more painful).

src/ASM/ASMs2D.C
akva2: we can consider renaming that but i'd rather do that separately. this is a virtual all the way down in ASMbase..

akva2:
as for commit structure, they are split this way for development purposes. i have lots of code changes in flight and the smaller the units, the easier it is to rebase, move around, review and compare etc. i will gladly squash before merge.

akva2:

src/ASM/ASMs2D.C
kmokstad: I know, I also tend to prefer separate line for the return (unless it is so obvious I will never consider a need for breaking the debugger at that line). I only asked for consistency (at least within the same method, not to enforce 1-line policy)

src/ASM/Test/TestASMs2Dmx.C
kmokstad: Right, that was what I missed, the `static` keyword was gone.
The content could heave been indented though, to make that anonymous namespace more "visible" for reader.

For the class-scoped string you however have `static constexpr const` prefix, which looks abit over-specified (but is probably with reason as well)

akva2: i'm not used to indenting for namespaces. and yes, but i need that because only static constexpr variables can be inline initialized, if not i'd need a compile unit to instance in.

PR #613, opened 2023-08-29
Unify ASMuxDmx and ASMuxD write methods

akva2:


kmokstad (approved):
This is just the LR-version of #612, so the comments there applies here as well.

akva2:

PR #614, opened 2023-08-30
use a lambda in ASMu3D::getBoundary1Nodes

akva2:
switch is preferable to ifs


akva2:

kmokstad (approved):

PR #615, opened 2023-08-30
Use getBasis(xxx) in various spots

akva2:
To make code clearer and to kill temporarily introduced casts.

### DIFF
--- a/src/ASM/ASMs2Dmx.C
+++ b/src/ASM/ASMs2Dmx.C
@@ -251,7 +251,6 @@ bool ASMs2Dmx::generateFEMTopology ()
   else if (shareFE == 'F')
     return true;
 
-  int i1, i2, j1, j2;
 #ifdef SP_DEBUG
   size_t nbasis = 0;
   for (auto& it : m_basis) {
@@ -259,17 +258,19 @@ bool ASMs2Dmx::generateFEMTopology ()
     std::cout <<":\nnumCoefs: "<< it->numCoefs_u() <<" "<< it->numCoefs_v();
     std::cout <<"\norder: "<< it->order_u() <<" "<< it->order_v();
     std::cout <<"\ndu:";
-    for (i1 = 0; i1 < it->numCoefs_u(); i1++)
+    for (int i1 = 0; i1 < it->numCoefs_u(); i1++)
       std::cout <<' '<< it->knotSpan(0,i1);
     std::cout <<"\ndv:";
-    for (i2 = 0; i2 < it->numCoefs_v(); i2++)
+    for (int i2 = 0; i2 < it->numCoefs_v(); i2++)
       std::cout <<' '<< it->knotSpan(1,i2);
     std::cout << std::endl;
   }
 #endif
 
-  nel = (m_basis[geoBasis-1]->numCoefs_u()-m_basis[geoBasis-1]->order_u()+1)*
-        (m_basis[geoBasis-1]->numCoefs_v()-m_basis[geoBasis-1]->order_v()+1);
+  const Go::SplineSurface* itg = this->getBasis(ASM::INTEGRATION_BASIS);
+
+  nel = (itg->numCoefs_u() - itg->order_u() + 1) *
+        (itg->numCoefs_v() - itg->order_v() + 1);
 
   myMLGE.resize(nel,0);
   myMLGN.resize(nnod);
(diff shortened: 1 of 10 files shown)


akva2:

src/ASM/ASMs2Dmx.C
```diff
@@ -277,60 +278,59 @@ bool ASMs2Dmx::generateFEMTopology ()
...
         myMLGN[inod++]    = ++gNod;
       }
 
-  int firstItgNode = this->getFirstItgElmNode();
-  int nElmNode = std::accumulate(elem_size.begin(), elem_size.end(), 0);
+  const size_t firstItgElmNode = this->getFirstItgElmNode();
+  const size_t nElmNode = std::accumulate(elem_size.begin(), elem_size.end(), 0);
 
   // Create nodal connectivities for bases
+  auto&& enumerate_elem = [this,&iel](const Go::SplineSurface& spline,
```
kmokstad, line 293: since `iel` is just an int and is not updated by the lambda, do we need to catch the reference here?

src/ASM/ASMs2Dmx.C
```diff
@@ -277,60 +278,59 @@ bool ASMs2Dmx::generateFEMTopology ()
...
-                    myMNPC[iel][lnod++] = iinod - m_basis[bas]->numCoefs_u()*j2 - j1;
-              } else
-                lnod += elem_size[bas];
-              basis_ofs += nb[bas];
+  auto knotv = itg->basis_v().begin();
+  for (int i2 = 1; i2 <= itg->numCoefs_v(); i2++, ++knotv) {
+    auto knotu = itg->basis_u().begin();
+    for (int i1 = 1; i1 <= itg->numCoefs_u(); i1++, inod++, ++knotu)
+      if (i1 >= itg->order_u() && i2 >= itg->order_v()) {
+        if (itg->knotSpan(0,i1-1) > 0.0 && itg->knotSpan(1,i2-1) > 0.0) {
```
kmokstad, line 307: Can maybe combine the  two if statements into a single one. Or even better, move the check on `i2` and `knotSpan(1,i2-1)` outside the inner-most loop, since it will evaluate to the the same each time, thereby saving some looping (microoptimization, think the results will be the same but needs to be verified), but you need to add a `else inod += itg->numCoefs_u();` though to compensate for the `inod++`

src/ASM/ASMs3Dmx.C
```diff
@@ -271,75 +271,79 @@ bool ASMs3Dmx::generateFEMTopology ()
...
+          myNodeInd[inod].I = i;
+          myNodeInd[inod].J = j;
+          myNodeInd[inod].K = k;
           myMLGN[inod++]    = ++gNod;
         }
 
-  int firstItgElmNode = this->getFirstItgElmNode();
-  int nElmNode = std::accumulate(elem_size.begin(), elem_size.end(), 0);
+  const int firstItgElmNode = this->getFirstItgElmNode();
+  const size_t nElmNode = std::accumulate(elem_size.begin(), elem_size.end(), 0);
```
kmokstad, line 298: why const ?

src/ASM/ASMs3Dmx.C
```diff
@@ -271,75 +271,79 @@ bool ASMs3Dmx::generateFEMTopology ()
...
 
   // Create nodal connectivities for bases
+  auto&& enumerate_elem = [this,&iel](const Go::SplineVolume& spline,
```
kmokstad, line 301: since `iel` is just a `size_t` and is not updated by the lambda, do we need to catch the reference here?

src/ASM/ASMs3Dmx.C
```diff
@@ -271,75 +271,79 @@ bool ASMs3Dmx::generateFEMTopology ()
...
+  auto knotw = itg->basis(2).begin();
+  for (int k = 1; k <= itg->numCoefs(2); k++, ++knotw) {
+    auto knotv = itg->basis(1).begin();
+    for (int j = 1; j <= itg->numCoefs(1); j++, ++knotv) {
+      auto knotu = itg->basis(0).begin();
+      for (int i = 1; i <= itg->numCoefs(0); i++, inod++, ++knotu)
+        if (i >= itg->order(0) && j >= itg->order(1) && k >= itg->order(2)) {
+          if (itg->knotSpan(0,i-1) > 0.0 &&
+              itg->knotSpan(1,j-1) > 0.0 &&
+              itg->knotSpan(2,k-1) > 0.0) {
```
kmokstad, line 322: Similar comment as in the 2D case, consider revising the if-statements

src/ASM/LR/ASMu2D.C
```diff
@@ -2764,12 +2764,15 @@ const RealArray& ASMu2D::InterfaceChecker::getIntersections (int iel, int edge,
 bool ASMu2D::refine (const LR::RefineData& prm, Vectors& sol)
 {
   bool ok = this->ASMLRSpline::refine(prm,sol);
-  if (!ok || !this->separateProjectionBasis() ||
-      prm.elements.size() + prm.errors.size() == 0)
+
+  // check if refinement was actually done
+  if (prm.elements.size() + prm.errors.size() == 0 || !ok)
     return ok;
```
kmokstad, line 2772: To make it even more readable, we can drop the `ok` variable entirely:
```
if (!this->ASMLRSpline::refine(prm,sol))
  return false;

// check if refinement was actually done
if (prm.elements.empty() && prm.errors.empty())
  return true;
```

src/ASM/LR/ASMu3D.C
```diff
@@ -2337,13 +2337,23 @@ void ASMu3D::extendRefinementDomain (IntSet& refineIndices,
 bool ASMu3D::refine (const LR::RefineData& prm, Vectors& sol)
 {
   bool ok = this->ASMLRSpline::refine(prm,sol);
-  if (!ok || !this->separateProjectionBasis() ||
-      prm.elements.size() + prm.errors.size() == 0)
+
+  // check if refinement was actually done
+  if (prm.elements.size() + prm.errors.size() == 0 || !ok)
     return ok;
```
kmokstad, line 2343: same here, consider replace with
```
if (!this->ASMLRSpline::refine(prm,sol))
  return false;

// check if refinement was actually done
if (prm.elements.empty() && prm.errors.empty())
  return true;
```

kmokstad (commented):
Probably fine, although I did not quite see the logic behind the changes regarding the spl0, spl1 and spl2 objects in the `assembleL2matrices()` methods.

src/ASM/ASMs2Dmx.C
akva2: heh. this is a common question and a common mistake to make. if we do not capture by reference, we capture the value when the lambda is instanced, and it will have that value in all subsequent calls. so yes, i do need to capture by reference.

src/ASM/ASMs2Dmx.C
```diff
@@ -277,60 +278,59 @@ bool ASMs2Dmx::generateFEMTopology ()
...
       {
         myNodeInd[inod].I = i1;
         myNodeInd[inod].J = i2;
         myMLGN[inod++]    = ++gNod;
       }
 
-  int firstItgNode = this->getFirstItgElmNode();
-  int nElmNode = std::accumulate(elem_size.begin(), elem_size.end(), 0);
+  const size_t firstItgElmNode = this->getFirstItgElmNode();
+  const size_t nElmNode = std::accumulate(elem_size.begin(), elem_size.end(), 0);
```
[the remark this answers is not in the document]
akva2: well, the normal reasons, const helps avoid mistakes and it aids the compiler by telling it that it can put this in a register. i figured i could add since i updated the types anyways, but i'm not insisting.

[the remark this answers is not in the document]
kmokstad: no, by all means, keep them if there is a reason (and helps the compiler).

src/ASM/ASMs3Dmx.C
kmokstad: just keep them

src/ASM/ASMs2Dmx.C
kmokstad: really? I though it was captured when invoked. That's a dangerous difference I was unaware of. Hopefully I have not introduced such lambdas elsewhere..

Thank's for the lesson ;)

src/ASM/ASMs3Dmx.C
kmokstad: yes, we do

src/ASM/ASMs2Dmx.C
akva2: while i can do the first part, it's .. tricky to the combine ifs. reason is we need to increase iel even if the knotspan is empty.

src/ASM/ASMs3Dmx.C
akva2: similar answer as well. sorry no can do.

src/ASM/LR/ASMu2D.C
akva2: done.

kmokstad (approved):

akva2:
as for the assembleL2matrices;

t's to save one set of BasisDerivsSf. Either we are not continuous so we only want spl0, now spl1. if not we are continuous but proj and integration basis is the same. then we just need spl2. if not we need spl2,
but we only need spl1 (ie just basis functions, no derivatives) for the projection basis. previously we evaluated the derivatives for the projection basis.

akva2:
should I squash the last 4?

akva2:
i squashed them.

akva2:

PR #617, opened 2023-09-02
Quell warnings with newer PETsc and gtest versions

akva2:
Avoid some deprecation warnings with newer versions of gtest/petsc.


akva2:
jenkins build this with downstreams ifem-stokes=237 please

PR #621, opened 2023-09-04
ASMu2D(mx): use lrspline instead of geomB in various spots

akva2:
these ops intends to use the integration basis


akva2:

akva2:
yes u3D needs the same. was in another branch..

akva2:

PR #624, opened 2023-09-06
fixed: reading nodes from HDF5 file

akva2:
we need to instance the ProcessAdm and not pass a temporary. holding a reference to a temporary is not enough to keep it alive, which means it is killed and consequently the reader holds an invalid reference.


akva2:

kmokstad (approved):

PR #626, opened 2023-09-07
ASMu2Dmx: add support for a separate geometry basis

akva2:
These are the changes required for 2DLR.

### DIFF
--- a/src/ASM/LR/ASMu2Drecovery.C
+++ b/src/ASM/LR/ASMu2Drecovery.C
@@ -108,9 +108,9 @@ bool ASMu2D::assembleL2matrices (SparseMatrix& A, StdVector& B,
 {
   size_t nnod = this->getNoProjectionNodes();
 
-  const LR::LRSplineSurface* itg = this->getBasis(ASM::INTEGRATION_BASIS);
+  const LR::LRSplineSurface* geo = this->getBasis(ASM::GEOMETRY_BASIS);
   const LR::LRSplineSurface* proj = this->getBasis(ASM::PROJECTION_BASIS);
-  const bool separateProjBasis = proj != itg;
+  const bool separateProjBasis = proj != geo;
   const bool useModelMNPC = !separateProjBasis && this->getNoBasis() == 1;
 
   const int p1 = proj->order(0);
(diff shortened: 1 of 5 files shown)


akva2:

src/ASM/LR/ASMu2Drecovery.C
```diff
@@ -153,23 +153,24 @@ bool ASMu2D::assembleL2matrices (SparseMatrix& A, StdVector& B,
...
+      int ielG = geo->getElementContaining(elm->midpoint())+1;
 
       if (continuous)
       {
         // Set up control point (nodal) coordinates for current element
         if (!this->getElementCoordinates(Xnod,iel)) {
           ok = false;
           continue;
-        } else if ((dA = 0.25*this->getParametricArea(iel)) < 0.0) {
+        } else if ((dA = 0.25*elm->area()) < 0.0) {
```
kmokstad, line 164: Consider same change on other uses of `getParametricArea()` for the LR-classes. Maybe we don't need that method at all since we probably always fetches the elm pointer for other uses anyway.

kmokstad (approved):

src/ASM/LR/ASMu2Drecovery.C
akva2: Yeah, done.

akva2:

src/ASM/LR/ASMu2Drecovery.C
kmokstad: great, the same sing for ASMu3D also then, maybe, although there the `getParametricVolume()` method in addition checks if `MNPC[iel-1].empty())`. Not sure if this is needed, or why the same was not done for 2D. At least there is an inconsitency here that should be sorted out, one way or the other.

akva2: i don't think that makes sense at all. We don't have empty knotspans represented as elements with LR so we also do not have empty mnpc-entries. i think it's just a copy-pasta from the struct code.

akva2: the fact that the method is completely unused also kinda shortcuts the discussion :)

akva2: yes, let me rephrase. it's completely unused after my refactorings.

PR #627, opened 2023-09-07
ASMu3Dmx: add support for a separate geometry basis

akva2:
Sits on top of https://github.com/OPM/IFEM/pull/626

### DIFF
--- a/src/ASM/LR/ASMLRSpline.h
+++ b/src/ASM/LR/ASMLRSpline.h
@@ -16,9 +16,9 @@
 
 #include "ASMbase.h"
 #include "ASMunstruct.h"
-#include "GoTools/geometry/BsplineBasis.h"
+#include "ThreadGroups.h"
 
-class ThreadGroups;
+#include "GoTools/geometry/BsplineBasis.h"
 
 
 namespace LR //! Utilities for LR-splines.
@@ -207,6 +207,10 @@ protected:
   std::shared_ptr<LR::LRSpline> projB;  //!< Pointer to spline object of the projection basis
   std::shared_ptr<LR::LRSpline> projB2; //!< Pointer to spline object of the secondary projection basis
   std::shared_ptr<LR::LRSpline> refB;   //!< Pointer to spline object of the refinement basis
+
+  ThreadGroups threadGroups; //!< Element groups for multi-threaded assembly
+  ThreadGroups projThreadGroups; //!< Element groups for multi-threaded assembly - projection basis
+  ThreadGroups proj2ThreadGroups; //!< Element groups for multi-threaded assembly - second projection basis
 };
 
 #endif
--- a/src/ASM/LR/ASMLRSpline.h
+++ b/src/ASM/LR/ASMLRSpline.h
@@ -181,6 +181,9 @@ public:
   //! \brief Computes the total number of integration points in this patch.
   virtual void getNoIntPoints(size_t& nPt, size_t& nIPt);
 
+  //! \brief Swaps between the first and second projection basis.
+  virtual void swapProjectionBasis();
+
 protected:
   //! \brief Refines the mesh adaptively.
   //! \param[in] prm Input data used to control the mesh refinement
--- a/src/ASM/LR/ASMu3Dmx.C
+++ b/src/ASM/LR/ASMu3Dmx.C
@@ -688,7 +688,6 @@ bool ASMu3Dmx::evalSolution (Matrix& sField, const Vector& locSol,
     return false;
 
   Vector   ptSol;
-  std::vector<Matrix> dNxdu(m_basis.size()), dNxdX(m_basis.size());
   Matrix   Jac, Xnod, eSol, ptDer;
 
   std::vector<Go::BasisPts> splinex(m_basis.size());
@@ -752,6 +751,9 @@ bool ASMu3Dmx::evalSolution (Matrix& sField, const IntegrandBase& integrand,
   if (nPoints != gpar[1].size() || nPoints != gpar[2].size())
     return false;
 
+  const LR::LRSplineVolume* geo = this->getBasis(ASM::GEOMETRY_BASIS);
+  const bool separateGeometry = geo != lrspline.get();
+
   // Evaluate the secondary solution field at each point
   for (size_t i = 0; i < nPoints; i++)
   {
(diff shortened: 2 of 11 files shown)


akva2:

akva2:

src/ASM/LR/ASMu3Dmx.C
```diff
@@ -352,9 +354,11 @@ bool ASMu3Dmx::integrate (Integrand& integrand,
       double   param[3] = { 0.0, 0.0, 0.0 };
       Vec4     X(param,time.t);
 
-      int iEl = el->getId();
+      int iEl = els[itgBasis-1]-1;
       fe.iel = MLGE[iEl];
 
+      el = lrspline->getElement(iEl);
```
kmokstad, line 360: Since the `el` variable is declared const (in L344), is it fine to assign a new value to it here?  Thought the point of const'ing it was to prevent if from being redefined (or maybe there is a pointer issue that still allows it).

Consider changing L348 into
```
this->getElementsAt(threadBasis->getElement(groups[t][e])->midpoint(),els,elem_sizes);
```
instead (yes, it then becomes longer than 80 chars), and moving the type declaration of el into L360, such that the code is comparable with the equivalent 2D version.

src/ASM/LR/ASMu3Drecovery.C
```diff
@@ -157,25 +157,26 @@ bool ASMu3D::assembleL2matrices (SparseMatrix& A, StdVector& B,
...
+      int iel = lrspline->getElementContaining(elm->midpoint()) + 1;
+      int ielG = geo->getElementContaining(elm->midpoint()) + 1;
 
       if (continuous)
       {
         // Set up control point (nodal) coordinates for current element
         if (!this->getElementCoordinates(Xnod,iel)) {
           ok = false;
           continue;
-        } else if ((dV = 0.125*this->getParametricVolume(iel)) < 0.0) {
```
kmokstad, line 168: OK, here goes the last `getParametricVolume()` usage.

src/ASM/LR/ASMu3Drecovery.C
```diff
@@ -157,25 +157,26 @@ bool ASMu3D::assembleL2matrices (SparseMatrix& A, StdVector& B,
...
+        } else if ((dV = 0.125*elm->volume()) < 0.0) {
           ok = false; // topology error (probably logic error)
```
kmokstad, line 170: ...but then this comment should  vanish as well. Think that also is inherited from the struct case where it was associated with the possible zero knot-spans, which we don't have in the LR-case.

src/ASM/LR/ASMu3Drecovery.C
```diff
@@ -533,6 +535,9 @@ bool ASMu3D::faceL2projection (const DirichletFace& face,
   for (size_t ie = 0; ie < face.MLGE.size(); ie++) // for all face elements
   {
     int iel = 1 + face.MLGE[ie];
+    int ielG = iel;
+    if (geo != lrspline.get())
+      ielG = geo->getElementContaining(lrspline->getElement(iel-1)->midpoint()) + 1;
```
kmokstad, line 540: Here also, consider the shorter `int ielG = geo == lrspline.get() ? iel : ...;` syntac although the line then gets even longer. But it is already beyond the 80-char limit, afaict

src/ASM/LR/ASMLRSpline.h
```diff
@@ -207,6 +207,11 @@ class ASMLRSpline : public ASMbase, public ASMunstruct
   std::shared_ptr<LR::LRSpline> projB;  //!< Pointer to spline object of the projection basis
   std::shared_ptr<LR::LRSpline> projB2; //!< Pointer to spline object of the secondary projection basis
   std::shared_ptr<LR::LRSpline> refB;   //!< Pointer to spline object of the refinement basis
+
+  ThreadGroups threadGroups; //!< Element groups for multi-threaded assembly
+  ThreadGroups projThreadGroups; //!< Element groups for multi-threaded assembly - projection basis
+  ThreadGroups proj2ThreadGroups; //!< Element groups for multi-threaded assembly - second projection basis
+
```
kmokstad, line 217: Remove blank line L214.
With this change also the non-mixed classes gets a `proj2ThreadGroups` member. That is intentionally, or an affordable side effect only. Do the non-mixed classes ever need a second projection basis?

src/ASM/LR/ASMLRSpline.h
```diff
@@ -181,6 +181,9 @@ class ASMLRSpline : public ASMbase, public ASMunstruct
   //! \brief Computes the total number of integration points in this patch.
   virtual void getNoIntPoints(size_t& nPt, size_t& nIPt);
 
+  //! \brief Swaps between the first and second projection basis.
+  virtual void swapProjectionBasis();
```
kmokstad, line 185: The same comment applies here. Has this method relevance for the non-mixed case?

kmokstad: OK, maybe you already answered that in the commit message which I read a bit too late ;)

kmokstad (approved):
Great cleaning work!

src/ASM/LR/ASMu3Dmx.C
akva2: Yes. it's a pointer to a const LR::Element, not a const pointer to a const LR::Element. If it was
```
const LR::Element* const el
``` 
then it points to a const element AND i cannot re-assign the el variable.

in any case i've made the requested change.

src/ASM/LR/ASMu3Drecovery.C
akva2: indeed.

src/ASM/LR/ASMu3Drecovery.C
akva2: this can happen if we have a lhs geometry, which can be said to be a topology error, but i have removed it.

src/ASM/LR/ASMu3Drecovery.C
akva2: it get's obnoxiously long but okay.

src/ASM/LR/ASMLRSpline.h
akva2: see above.

akva2:

PR #628, opened 2023-09-08
changed: use ASM(mx)u(Cube|Square) in TestLRSplineField(s) 

akva2:
While at it also plug some memory leaks.


akva2:
jenkins build this please

kmokstad (approved):
Keep the SIM-classes away from the ASM classes and associated tests, like here, as long it is possible. So this is a good improvement in that way also :-) The dependencies were intended to be one-way only (SIM depends on ASM, but not vice versa).

PR #629, opened 2023-09-08
changed: establishing RT basis now uses order lowering 

akva2:
does not make sense to use order raising since the geometry needs to be the highest order

### DIFF
--- a/src/ASM/ASMmxBase.C
+++ b/src/ASM/ASMmxBase.C
@@ -133,7 +133,8 @@ ASMmxBase::SurfaceVec ASMmxBase::establishBases (Go::SplineSurface* surf,
   if (type == FULL_CONT_RAISE_BASIS1 || type == FULL_CONT_RAISE_BASIS2)
   {
     // basis1 should be one degree higher than basis2 and C^p-1 continuous
-    result[0].reset(ASMmxBase::raiseBasis(surf));
+    result[0].reset(ASMmxBase::adjustBasis(*surf,{SplineUtils::AdjustOp::Raise,
+                                                  SplineUtils::AdjustOp::Raise}));
     result[1].reset(new Go::SplineSurface(*surf));
     itgBasis = 2;
   }
@@ -260,7 +261,9 @@ ASMmxBase::VolumeVec ASMmxBase::establishBases(Go::SplineVolume* svol,
   if (type == FULL_CONT_RAISE_BASIS1 || type == FULL_CONT_RAISE_BASIS2)
   {
     // basis1 should be one degree higher than basis2 and C^p-1 continuous
-    result[0].reset(ASMmxBase::raiseBasis(svol));
+    result[0].reset(ASMmxBase::adjustBasis(*svol,{SplineUtils::AdjustOp::Raise,
+                                                  SplineUtils::AdjustOp::Raise,
+                                                  SplineUtils::AdjustOp::Raise}));
     result[1].reset(new Go::SplineVolume(*svol));
     itgBasis = 2;
   }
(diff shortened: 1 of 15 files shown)


akva2:
jenkins build this with downstreams ifem-stokes=242 ifem-navierstokes=195 ifem-boussinesq=97 please

akva2:
jenkins build this with downstreams ifem-stokes=242 ifem-navierstokes=195 ifem-boussinesq=97 please

src/Utility/Test/TestSplineUtils.C
```diff
@@ -450,3 +451,75 @@ TEST(TestSplineUtils, BuildKnotVector)
...
+
+
+TEST_P(TestSplineUtils, AdjustBasis)
+{
+  Go::BsplineBasis basis_in(GetParam().order_in,
+                            GetParam().knots_in.begin(),
+                            GetParam().knots_in.end());
+
+  const Go::BsplineBasis res = SplineUtils::adjustBasis(basis_in,GetParam().op);
+  const std::vector<double> out{res.begin(), res.end()};
```
kmokstad, line 520: This looks suspicious, `begin()` and `end()` returns iterators, afaik, and then you put them into a double array? Or it this another way of copying the array.

src/ASM/ASMmxBase.C
```diff
@@ -390,13 +394,15 @@ ASMmxBase::VolumeVec ASMmxBase::establishBases(Go::SplineVolume* svol,
 }
 
 
-Go::SplineSurface* ASMmxBase::raiseBasis (Go::SplineSurface* surf)
+Go::SplineSurface* ASMmxBase::adjustBasis (const Go::SplineSurface& surf,
+                                          const std::array<SplineUtils::AdjustOp,2>& ops)
```
kmokstad, line 260: check indent

kmokstad (approved):
Can't go into the details on all the simplifications here. I trust they do as intended. Only commented on some cosmetic issues.

src/Utility/SplineUtils.C
```diff
@@ -309,6 +309,30 @@ Go::SplineVolume* SplineUtils::project (const Go::SplineVolume* volume,
...
+  std::vector<int> mult;
+  basis.knotMultiplicities(mult);
+  std::vector<int> cont(knots.size());
+  cont.front() = cont.back() = -1;
+  int order = basis.order();
+  int start_cont = adjust == AdjustOp::Lower ? order-2 : order;
+  int mult_adjust = adjust == AdjustOp::Lower ? 0 : 1;
+  int new_order = adjust == AdjustOp::Lower ? order - 1 : order + 1;
+  for (size_t i = 1; i < knots.size()-1; ++i)
+    cont[i]  = start_cont - (mult[i] == 1 ? 1 : mult[i] + mult_adjust);
```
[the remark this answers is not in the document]
akva2: fixed.

src/ASM/ASMmxBase.C
akva2: fixed.

akva2:
jenkins build this with downstreams ifem-stokes=242 ifem-navierstokes=195 ifem-boussinesq=97 please

src/Utility/Test/TestSplineUtils.C
kmokstad: ok, I was confused by the syntax, I thought that constructor was invoked with parentheses, i.e. `out(res.begin(), res.end())`, and when using braces `{}` means a list of elements instead and here this would be a length-two vector with that start and stop iterator values as elements. But clearly that was not the case then..

akva2: right. it's c++-11 uniform initialization. you can initialize arrays or classes or anything really using {} . as it has advantages, such as not allowing implicit conversions of arguments, i've made it a habit to use it (see https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-list for all the itty-gritty details if you are interested).

PR #630, opened 2023-09-13
cleanup ASMuxDmx::evalSolution

akva2:
- use getElementsAt
- no reason to have a vector of evaluation objects

### DIFF
--- a/src/ASM/LR/ASMu2Dmx.C
+++ b/src/ASM/LR/ASMu2Dmx.C
@@ -362,7 +362,7 @@ bool ASMu2Dmx::integrate (Integrand& integrand,
 
       std::vector<int>    els;
       std::vector<size_t> elem_sizes;
-      this->getElementsAt(threadBasis->getElement(groups[t][e])->midpoint(),els,elem_sizes);
+      this->getElementsAt(threadBasis->getElement(groups[t][e])->midpoint(),els,&elem_sizes);
 
       MxFiniteElement fe(elem_sizes);
       Matrix   Xnod, Jac;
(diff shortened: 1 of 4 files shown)


akva2:

kmokstad (approved):
The only thing is that you now compute `elem_sizes` unnecessarily. One could consider making that an optional argument to `getElementsAt()` to avoid it, but not a big deal.

akva2:
yeah, i can do that. note that i will need them later when adding piola there (only when enabled of course) so it will be a little back and forth.

akva2:

src/ASM/LR/ASMu2Dmx.C
```diff
@@ -850,27 +850,26 @@ bool ASMu2Dmx::evalSolution (Matrix& sField, const Vector& locSol,
   sField.resize(std::accumulate(nc.begin(), nc.end(), 0), nPoints);
   for (size_t i = 0; i < nPoints; i++)
   {
+    std::vector<int>    els;
```
kmokstad, line 853: Now too many spaces between type and variable  (one is enough since elem_sizes now gone)

src/ASM/LR/ASMu3Dmx.C
```diff
@@ -709,27 +709,26 @@ bool ASMu3Dmx::evalSolution (Matrix& sField, const Vector& locSol,
   sField.resize(std::accumulate(nc.begin(), nc.end(), 0), nPoints);
   for (size_t i = 0; i < nPoints; i++)
   {
+    std::vector<int>    els;
```
kmokstad, line 712: same here

kmokstad (approved):

src/ASM/LR/ASMu2Dmx.C
akva2: done.

akva2:

PR #632, opened 2023-09-14
Some matrix utilities

akva2:
This exposes stride and offset for x vector in utl::vector<T>::add and adds an extractBlock method to utl::matrix<T>.
Useful for making code more readable while doing some of the piola stuff.

### DIFF
--- a/src/LinAlg/matrix.h
+++ b/src/LinAlg/matrix.h
@@ -565,6 +565,24 @@ namespace utl //! General utility classes and functions.
       }
     }
 
+    //! \brief Extract a block of the matrix to another matrix.
+    void extractBlock(matrix<T>& block, size_t r, size_t c,
+                      bool addTo = false,
+                      bool transposed = false) const
+    {
+      size_t nr = transposed ? block.cols() : block.rows();
+      size_t nc = transposed ? block.rows() : block.cols();
+      for (size_t i = 1; i <= nr && i+r-1 <= nrow; i++)
+      {
+        size_t ip = i+r-2 + nrow*(c-1);
+        for (size_t j = 1; j <= nc && j+c-1 <= ncol; j++, ip += nrow)
+          if (addTo)
+            (transposed ? block(j,i) : block(i,j)) += this->elem[ip];
+          else
+            (transposed ? block(j,i) : block(i,j)) = this->elem[ip];
+      }
+    }
+
     //! \brief Create a diagonal matrix.
     matrix<T>& diag(T d, size_t dim = 0)
     {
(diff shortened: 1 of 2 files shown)


akva2:

src/LinAlg/matrix.h
```diff
@@ -963,19 +964,21 @@ namespace utl //! General utility classes and functions.
 
   template<> inline
   vector<float>& vector<float>::add(const std::vector<float>& X,
-                                    const float& alfa)
+                                    const float& alfa,
+                                    int ofsx, int stridex)
   {
     size_t n = this->size() < X.size() ? this->size() : X.size();
-    cblas_saxpy(n,alfa,&X.front(),1,this->ptr(),1);
+    cblas_saxpy(n,alfa,&X[ofsx],stridex,this->ptr(),1);
```
kmokstad, line 989: Or isn't it better with `X.data()+ofsx` in this case?

src/LinAlg/matrix.h
```diff
@@ -963,19 +964,21 @@ namespace utl //! General utility classes and functions.
...
 
   template<> inline
   vector<double>& vector<double>::add(const std::vector<double>& X,
-                                      const double& alfa)
+                                      const double& alfa,
+                                      int ofsx, int stridex)
   {
     size_t n = this->size() < X.size() ? this->size() : X.size();
-    cblas_daxpy(n,alfa,&X.front(),1,this->ptr(),1);
+    cblas_daxpy(n,alfa,&X[ofsx],stridex,this->ptr(),1);
```
kmokstad, line 999: `x.data()+ofsx` instead?

src/LinAlg/matrix.h
```diff
@@ -1345,11 +1348,12 @@ namespace utl //! General utility classes and functions.
...
 
   template<class T> inline
-  vector<T>& vector<T>::add(const std::vector<T>& X, const T& alfa)
+  vector<T>& vector<T>::add(const std::vector<T>& X, const T& alfa,
+                            int ofsx, int stridex)
   {
     T* p = this->ptr();
-    const T* q = &X.front();
-    for (size_t i = 0; i < this->size() && i < X.size(); i++, p++, q++)
+    const T* q = &X.front() + ofsx;
```
kmokstad, line 1373: `x.data() + ofsx` ?

Maybe the `ptr()` method also should be replaced by `data()` which is inherited from `std::vector` and does the same. `ptr()` was added long ago and before `data()` was known to the author.

kmokstad (commented):
Maybe this is still in progress (since the test fails), but I added some comments anyway

akva2:
failure was with an old version (github didn't update properly).

akva2:

src/LinAlg/matrix.h
akva2: sure.

src/LinAlg/matrix.h
akva2: sure.

src/LinAlg/matrix.h
akva2: sure.

akva2:

kmokstad (approved):
The replacement for `ptr()` by `data()` elsewhere can be done later - it is used a lot.

PR #633, opened 2023-09-15
changed: use BasisFunctionValues also in boundary integrals

akva2:
this allows simplifying the MxFiniteElement::(Hessian|Jacobian) methods.

add a type alias and a helper class to make it convenient.

### DIFF
--- a/src/ASM/ASMs2Dmx.C
+++ b/src/ASM/ASMs2Dmx.C
@@ -551,7 +551,7 @@ bool ASMs2Dmx::integrate (Integrand& integrand,
             fe.v = param[1] = cache.getParam(1,i2-p2,j);
 
             // Fetch basis function derivatives at current integration point
-            std::vector<const BasisFunctionVals*> bfs(myCache.size());
+            BasisValuesPtrs bfs(myCache.size());
             for (size_t b = 0; b < myCache.size(); ++b) {
               bfs[b] = &myCache[b]->getVals(iel-1, ip);
               if (b < m_basis.size())
@@ -560,11 +560,11 @@ bool ASMs2Dmx::integrate (Integrand& integrand,
 
             // Compute Jacobian inverse of the coordinate mapping and
             // basis function derivatives w.r.t. Cartesian coordinates
-            if (!fe.Jacobian(Jac,Xnod,itgBasis,&bfs))
+            if (!fe.Jacobian(Jac,Xnod,itgBasis,bfs))
               continue; // skip singular points
 
             // Compute Hessian of coordinate mapping and 2nd order derivatives
-            if (use2ndDer && !fe.Hessian(Hess,Jac,Xnod,itgBasis,&bfs))
+            if (use2ndDer && !fe.Hessian(Hess,Jac,Xnod,itgBasis,bfs))
               ok = false;
 
             // Compute G-matrix
@@ -662,8 +662,7 @@ bool ASMs2Dmx::integrate (Integrand& integrand, int lIndex,
   fe.v = gpar[1](1,1);
   double param[3] = { fe.u, fe.v, 0.0 };
 
-  Matrices dNxdu(splinex.size());
-  Vector Ng;
+  BasisValues bfs(splinex.size());
   Matrix Xnod, Jac;
   Vec4   X(param,time.t);
   Vec3   normal;
@@ -729,21 +728,21 @@ bool ASMs2Dmx::integrate (Integrand& integrand, int lIndex,
         }
 
         if (separateGeometry)
-          SplineUtils::extractBasis(splinex.back()[ip],Ng,dNxdu.back());
+          SplineUtils::extractBasis(splinex.back()[ip],bfs.back().N,bfs.back().dNdu);
 
         // Fetch basis function derivatives at current integration point
         for (size_t b = 0; b < m_basis.size(); ++b)
-          SplineUtils::extractBasis(splinex[b][ip],fe.basis(b+1),dNxdu[b]);
+          SplineUtils::extractBasis(splinex[b][ip],fe.basis(b+1),bfs[b].dNdu);
 
         // Compute Jacobian inverse of the coordinate mapping and
         // basis function derivatives w.r.t. Cartesian coordinates
-        if (!fe.Jacobian(Jac,normal,Xnod,itgBasis,dNxdu,t1,t2))
+        if (!fe.Jacobian(Jac,normal,Xnod,itgBasis,bfs,t1,t2))
           continue; // skip singular points
 
         if (edgeDir < 0) normal *= -1.0;
 
         // Cartesian coordinates of current integration point
-        X.assign(Xnod * (separateGeometry ? Ng : fe.basis(itgBasis)));
+        X.assign(Xnod * (separateGeometry ? bfs.back().N : fe.basis(itgBasis)));
 
         // Evaluate the integrand and accumulate element contributions
         fe.detJxW *= dS*wg[i];
@@ -811,8 +810,7 @@ bool ASMs2Dmx::integrate (Integrand& integrand,
   std::copy(elem_size.begin(),elem_size.end(),std::back_inserter(elem_sizes2));
 
   MxFiniteElement fe(elem_sizes2);
-  Matrix        dNdu, Xnod, Jac;
-  Vector        dN;
+  Matrix        Xnod, Jac;
   Vec4          X(nullptr,time.t);
   Vec3          normal;
   double        u[2], v[2];
@@ -906,17 +904,17 @@ bool ASMs2Dmx::integrate (Integrand& integrand,
             }
 
             // Fetch basis function derivatives at current integration point
-            Matrices dNxdu(2*nB);
+            BasisValues bfs(2*nB);
             for (size_t b = 1; b <= nB; b++) {
               Go::BasisDerivsSf spline;
-              this->getBasis(b)->computeBasis(fe.u, fe.v, spline, edgeDir < 0);
-              SplineUtils::extractBasis(spline, fe.basis(b), dNxdu[b-1]);
-              this->getBasis(b)->computeBasis(fe.u, fe.v, spline, edgeDir > 0);
-              SplineUtils::extractBasis(spline, fe.basis(nB+b), dNxdu[nB+b-1]);
+              this->getBasis(b)->computeBasis(fe.u,fe.v,spline,edgeDir < 0);
+              SplineUtils::extractBasis(spline,fe.basis(b),bfs[b-1].dNdu);
+              this->getBasis(b)->computeBasis(fe.u,fe.v,spline,edgeDir > 0);
+              SplineUtils::extractBasis(spline,fe.basis(nB+b),bfs[nB+b-1].dNdu);
             }
 
             // Compute basis function derivatives and the edge normal
-            if (!fe.Jacobian(Jac,normal,Xnod,itgBasis,dNxdu,t1,t2,nB))
+            if (!fe.Jacobian(Jac,normal,Xnod,itgBasis,bfs,t1,t2,nB))
               continue; // skip singular points
 
             if (edgeDir < 0) normal *= -1.0;
@@ -1056,7 +1054,7 @@ bool ASMs2Dmx::evalSolution (Matrix& sField, const IntegrandBase& integrand,
 
   MxFiniteElement fe(elem_size,firstIp);
   Vector          solPt;
-  Matrices        dNxdu(m_basis.size());
+  BasisValues     bfs(m_basis.size());
   Matrix          Jac;
 
   // Evaluate the secondary solution field at each point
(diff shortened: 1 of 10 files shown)


akva2:

src/ASM/ASMs2Dmx.C
```diff
@@ -1086,11 +1085,11 @@ bool ASMs2Dmx::evalSolution (Matrix& sField, const IntegrandBase& integrand,
 
     // Fetch basis function derivatives at current integration point
     for (size_t b = 0; b < m_basis.size(); ++b)
-      SplineUtils::extractBasis(splinex[b][i],fe.basis(b+1),dNxdu[b]);
+      SplineUtils::extractBasis(splinex[b][i],fe.basis(b+1),bfs[b].dNdu);
```
kmokstad, line 1088: ...and here

src/ASM/ASMs2DmxLag.C
```diff
@@ -422,13 +422,13 @@ bool ASMs2DmxLag::integrate (Integrand& integrand, int lIndex,
         // Compute the basis functions and their derivatives, using
         // tensor product of one-dimensional Lagrange polynomials
         for (size_t b = 0; b < nxx.size(); ++b)
-          if (!Lagrange::computeBasis(fe.basis(b+1),dNxdu[b],
+          if (!Lagrange::computeBasis(fe.basis(b+1),bfs[b].dNdu,
```
kmokstad, line 425: and here (fe vs bfs usage)

src/ASM/ASMs2DmxLag.C
```diff
@@ -521,14 +521,14 @@ bool ASMs2DmxLag::evalSolution (Matrix& sField, const IntegrandBase& integrand,
 	double xi  = -1.0 + i*incx;
 	double eta = -1.0 + j*incy;
         for (size_t b = 0; b < nxx.size(); ++b)
-          if (!Lagrange::computeBasis(fe.basis(b+1),dNxdu[b],
+          if (!Lagrange::computeBasis(fe.basis(b+1),bfs[b].dNdu,
```
kmokstad, line 524: here the same

src/ASM/ASMs3Dmx.C
```diff
@@ -806,21 +805,21 @@ bool ASMs3Dmx::integrate (Integrand& integrand, int lIndex,
             }
 
             if (separateGeometry)
-              SplineUtils::extractBasis(splinex.back()[ip],Ng,dNxdu.back());
+              SplineUtils::extractBasis(splinex.back()[ip],bfs.back().N,bfs.back().dNdu);
 
             // Fetch basis function derivatives at current integration point
             for (size_t b = 0; b < m_basis.size(); ++b)
-              SplineUtils::extractBasis(splinex[b][ip],fe.basis(b+1),dNxdu[b]);
+              SplineUtils::extractBasis(splinex[b][ip],fe.basis(b+1),bfs[b].dNdu);
```
kmokstad, line 812: same

src/ASM/ASMs3Dmx.C
```diff
@@ -1021,17 +1020,17 @@ bool ASMs3Dmx::integrate (Integrand& integrand,
...
-                Matrices dNxdu(2*nB);
+                BasisValues bfs(2*nB);
                 for (size_t b = 1; b <= nB; b++) {
                   Go::BasisDerivs spline;
                   this->getBasis(b)->computeBasis(fe.u, fe.v, fe.w, spline, faceDir < 0);
-                  SplineUtils::extractBasis(spline, fe.basis(b), dNxdu[b-1]);
+                  SplineUtils::extractBasis(spline, fe.basis(b), bfs[b-1].dNdu);
                   this->getBasis(b)->computeBasis(fe.u, fe.v, fe.w, spline, faceDir > 0);
-                  SplineUtils::extractBasis(spline, fe.basis(nB+b), dNxdu[nB+b-1]);
+                  SplineUtils::extractBasis(spline, fe.basis(nB+b), bfs[nB+b-1].dNdu);
```
kmokstad, line 1029: more of the same (both L1027 and L1029)

src/ASM/ASMs3Dmx.C
```diff
@@ -1204,11 +1203,11 @@ bool ASMs3Dmx::evalSolution (Matrix& sField, const IntegrandBase& integrand,
 
     // Fetch basis function derivatives at current integration point
     for (size_t b = 0; b < m_basis.size(); ++b)
-      SplineUtils::extractBasis(splinex[b][i],fe.basis(b+1),dNxdu[b]);
+      SplineUtils::extractBasis(splinex[b][i],fe.basis(b+1),bfs[b].dNdu);
```
kmokstad, line 1206: also here

src/ASM/ASMs3DmxLag.C
```diff
@@ -487,14 +486,14 @@ bool ASMs3DmxLag::integrate (Integrand& integrand, int lIndex,
 	    // Compute the basis functions and their derivatives, using
 	    // tensor product of one-dimensional Lagrange polynomials
             for (size_t b = 0; b < nxx.size(); ++b)
-              if (!Lagrange::computeBasis(fe.basis(b+1),dNxdu[b],
+              if (!Lagrange::computeBasis(fe.basis(b+1),bfs[b].dNdu,
```
kmokstad, line 489: and here

src/ASM/ASMs3DmxLag.C
```diff
@@ -590,15 +589,15 @@ bool ASMs3DmxLag::evalSolution (Matrix& sField, const IntegrandBase& integrand,
 	  fe.eta  = -1.0 + j*incy;
 	  fe.zeta = -1.0 + k*incz;
           for (size_t b = 0; b < nxx.size(); ++b)
-            if (!Lagrange::computeBasis(fe.basis(b+1),dNxdu[b],
+            if (!Lagrange::computeBasis(fe.basis(b+1),bfs[b].dNdu,
```
kmokstad, line 592: and here

src/ASM/BasisFunctionVals.h
```diff
@@ -29,4 +29,28 @@ struct BasisFunctionVals
   Matrix4D d3Ndu3; //!< Third order derivatives of basis functions
 };
 
+
+//! \brief Convenience type alias
+using BasisValuesPtrs = std::vector<const BasisFunctionVals*>;
+
+
+//! \brief Utility class holding a vector of basis function values.
```
kmokstad, line 37: Here I had to pause a minute to understand what this class was about. Seems like the idea is to have a class which is a container holding BasisFunctionVals objects, which also can be cast into a an array of _pointers_ to the same BasisFunctionVals objects. Maybe describe this fact and why it is needed in a `\details` doxy section.

src/ASM/FiniteElement.h
```diff
@@ -14,6 +14,7 @@
 #ifndef _FINITE_ELEMENT_H
 #define _FINITE_ELEMENT_H
 
+#include "BasisFunctionVals.h"
```
kmokstad, line 17: OK, but then you can remove the forward declaration of the same type in L23 (not shown here)

src/ASM/FiniteElement.C
```diff
@@ -109,7 +107,7 @@ bool MxFiniteElement::Jacobian (Matrix& Jac, const Matrix& Xnod,
 bool MxFiniteElement::Jacobian (Matrix& Jac, Vec3& n,
                                 const Matrix& Xnod,
                                 unsigned short int gBasis,
-                                const std::vector<Matrix>& dNxdu,
+                                const BasisValuesPtrs& dNxdu,
```
kmokstad, line 110: Should maybe change the name of this parameter to `bfs` or `bf` like elsewhere, for consistency? Since this object also holds the basis function values themselves - not only the (first-order) derivatives

src/ASM/FiniteElement.h
```diff
@@ -122,25 +123,23 @@ class MxFiniteElement : public FiniteElement
...
-  //! \param[in] dNxdu First order derivatives of basis functions
+  //! \param[in] dNxdu Derivatives of basis functions
   //! \param[in] t1 First parametric tangent direction of the boundary
   //! \param[in] t2 Second parametric tangent direction of the boundary
   //! \param[in] nBasis Number of basis functions
   //! \param[in] Xnod2 Matrix of element nodal coordinates for neighbor element
   bool Jacobian(Matrix& Jac, Vec3& n, const Matrix& Xnod,
                 unsigned short int gBasis,
-                const std::vector<Matrix>& dNxdu,
+                const BasisValuesPtrs& dNxdu,
```
kmokstad, line 142: Consider name change to `bf` (like for the Hessian method). Then also remember to update the doxy

src/ASM/LR/ASMu2Dmx.C
```diff
@@ -777,17 +776,17 @@ bool ASMu2Dmx::integrate (Integrand& integrand,
...
-            std::vector<Matrix> dNxdu(2*nB);
+            BasisValues bfs(2*nB);
             for (size_t b = 0; b < nB; b++) {
               Go::BasisDerivsSf spline;
               this->computeBasis(fe.u+epsu, fe.v+epsv, spline, els[b]-1, m_basis[b].get());
-              SplineUtils::extractBasis(spline,fe.basis(b+1),dNxdu[b]);
+              SplineUtils::extractBasis(spline,fe.basis(b+1),bfs[b].dNdu);
               this->computeBasis(fe.u-epsu, fe.v-epsv, spline, els2[b]-1, m_basis[b].get());
-              SplineUtils::extractBasis(spline,fe.basis(nB+b+1),dNxdu[nB+b]);
+              SplineUtils::extractBasis(spline,fe.basis(nB+b+1),bfs[nB+b].dNdu);
```
kmokstad, line 785: similar (L783 and L785)

kmokstad (commented):
Sorry for all the comments (which mostly are on the same issue). Will approve it when things are a but more clear to me,

src/ASM/ASMs2Dmx.C
```diff
@@ -729,21 +728,21 @@ bool ASMs2Dmx::integrate (Integrand& integrand, int lIndex,
         }
 
         if (separateGeometry)
-          SplineUtils::extractBasis(splinex.back()[ip],Ng,dNxdu.back());
+          SplineUtils::extractBasis(splinex.back()[ip],bfs.back().N,bfs.back().dNdu);
 
         // Fetch basis function derivatives at current integration point
         for (size_t b = 0; b < m_basis.size(); ++b)
-          SplineUtils::extractBasis(splinex[b][ip],fe.basis(b+1),dNxdu[b]);
+          SplineUtils::extractBasis(splinex[b][ip],fe.basis(b+1),bfs[b].dNdu);
```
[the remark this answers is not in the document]
akva2: yes. i need the basis function values in the finite element. so putting them in the bfs object here would just mean an unnecessary copy operation. we don't need them in the bfs objects, we only need the parametric derivatives for the mapping calls. for the geometry however i don't have a vector in the finite element to store the basis function values, so there i use the bf vector for storage.

src/ASM/ASMs3DmxLag.C
```diff
@@ -430,7 +429,7 @@ bool ASMs3DmxLag::integrate (Integrand& integrand, int lIndex,
     for (size_t t = 0; t < threadGrp[g].size(); t++)
     {
       MxFiniteElement fe(elem_size);
-      Matrices dNxdu;
+      BasisValues bfs(elem_size.size());
```
akva2, line 432: this fixes a bug.

src/ASM/BasisFunctionVals.h
akva2: i can do that. this is because we get pointers (references) back from the cache, but we don't want deal with the pointers in other usages. but i still need to pass pointers into fe-methods. this utility class handles that transparently through the casting operator.

src/ASM/FiniteElement.C
akva2: yeah. i did that elsewhere but i guess i forgot to update this branch.

src/ASM/FiniteElement.h
akva2: done.

src/ASM/FiniteElement.h
akva2: done (bfs everywhere).

akva2: done.

src/ASM/ASMs2Dmx.C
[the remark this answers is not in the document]
kmokstad: I see, it is clear now (although not an ideal solution, but I don't see a better way right now). We can live with this and hope it does not create future confusion of any kind.

src/ASM/ASMs3DmxLag.C
kmokstad: what bug? The BasisValues object was introduced in this commit for this method, afics

akva2: the dNxdu was empty previously...

kmokstad: Indeed, which means that mixed 3D Lagrange (with Neumann conditions) is not covered by any of the regression tests.

akva2:

akva2:

akva2:

akva2:

PR #635, opened 2023-09-22
Add autodiff library and expression functions with auto-differentiation

akva2:
This is the start of a PR series where we ultimately end up with being able to do

```xml
<anasol type="expression_ad">
  <primary>some-expression-function</primary>
</anasol>
```
and autodiff takes care of derivation to obtain the secondary solution.

Likewise we will be able to combine this with
```xml
<source type="anasol"/>
```
where the source function is found by auto-differentation of the anasol.

This has been a major pain point in setting up cases for convergence studies.

In this PR the autodiff library is added and the expreval library is extended (templated) to be able to have autodifferentation of expression functions.

PR #636, opened 2023-09-22
ExprFunctions: some modernization

akva2:
- use language construct to manage pointers
- use anon namespace
- use override
- use type aliases, not typedefs

Sits on top of https://github.com/OPM/IFEM/pull/635

### DIFF
--- a/src/Utility/ExprFunctions.C
+++ b/src/Utility/ExprFunctions.C
@@ -91,7 +91,7 @@ int EvalFunc::numError = 0;
 
 
 EvalFunc::EvalFunc (const char* function, const char* x, Real eps)
-  : gradient(nullptr), dx(eps)
+  : dx(eps)
 {
   try {
 #ifdef USE_OPENMP
@@ -105,51 +105,31 @@ EvalFunc::EvalFunc (const char* function, const char* x, Real eps)
     expr.resize(nalloc);
     arg.resize(nalloc);
     for (size_t i = 0; i < nalloc; ++i) {
-      expr[i] = new Expression;
-      f[i] = new FunctionList;
-      v[i] = new ValueList;
+      expr[i] = std::make_unique<Expression>();
+      f[i] = std::make_unique<FunctionList>();
+      v[i] = std::make_unique<ValueList>();
       f[i]->AddDefaultFunctions();
       v[i]->AddDefaultValues();
       v[i]->Add(x,0.0,false);
-      expr[i]->SetFunctionList(f[i]);
-      expr[i]->SetValueList(v[i]);
+      expr[i]->SetFunctionList(f[i].get());
+      expr[i]->SetValueList(v[i].get());
       expr[i]->Parse(function);
       arg[i] = v[i]->GetAddress(x);
     }
   }
   catch (ExprEval::Exception& e) {
-    this->cleanup();
     ExprException(e,"parsing",function);
   }
 }
 
 
-EvalFunc::~EvalFunc ()
-{
-  this->cleanup();
-}
-
-
-void EvalFunc::cleanup ()
-{
-  for (Expression* it : expr)
-    delete it;
-  for (FunctionList* it : f)
-    delete it;
-  for (ValueList* it : v)
-    delete it;
-  delete gradient;
-  expr.clear();
-  f.clear();
-  v.clear();
-  arg.clear();
-}
+EvalFunc::~EvalFunc () = default;
 
 
 void EvalFunc::addDerivative (const std::string& function, const char* x)
 {
   if (!gradient)
-    gradient = new EvalFunc(function.c_str(),x);
+    gradient = std::make_unique<EvalFunc>(function.c_str(),x);
 }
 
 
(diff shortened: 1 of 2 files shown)


akva2:

akva2:

akva2:

src/Utility/ExprFunctions.C
```diff
@@ -87,7 +92,56 @@ static void ExprException (const ExprEval::Exception& exc, const char* task,
...
+
+  if (d1 < 1 || d2 > 3)
+    return -1; // Out-of-range
+  else if (d2-d1 == 0) // diagonal term, 11, 22 and 33
+    return d1-1;
+  else if (d2-d1 == 1) // off-diagonal term, 12 and 23
+    return d2+1;
+  else // off-diagonal term, 13
+    return 5;
+}
```
kmokstad, line 113: Since the `voigtIdx()` helper is only used once (in `addDerivative()`) it could as well be defined as a local Lamda in that method, just to make that more clear and since you here are moving the implementation anyway.

src/Utility/ExprFunctions.C
```diff
@@ -105,51 +105,31 @@ EvalFunc::EvalFunc (const char* function, const char* x, Real eps)
...
-    delete it;
-  for (ValueList* it : v)
-    delete it;
-  delete gradient;
-  expr.clear();
-  f.clear();
-  v.clear();
-  arg.clear();
-}
+EvalFunc::~EvalFunc () = default;
```
kmokstad, line 180: The now empty (or default) destructor should then be inlined in the header file (and the current destructor doxy's are no longer valid).

src/Utility/ExprFunctions.C
```diff
@@ -234,31 +213,7 @@ EvalFunction::EvalFunction (const char* function, Real epsX, Real epsT)
...
-  for (EvalFunction* it : dgradient)
-    delete it;
-  expr.clear();
-  f.clear();
-  v.clear();
-  arg.clear();
-  gradient.fill(nullptr);
-  dgradient.fill(nullptr);
-}
+EvalFunction::~EvalFunction () = default;
```
kmokstad, line 270: Iinline default destructor in header file instead

src/Utility/ExprFunctions.C
```diff
@@ -421,15 +376,11 @@ EvalFunctions::EvalFunctions (const std::string& functions,
...
+    p.emplace_back(std::make_unique<EvalFunction>(comp.c_str()));
 }
 
 
-EvalFunctions::~EvalFunctions ()
-{
-  for (EvalFunction* it : p)
-    delete it;
-}
+EvalFunctions::~EvalFunctions () = default;
```
kmokstad, line 388: Inline default destructor.

src/Utility/ExprFunctions.C
```diff
@@ -87,7 +92,56 @@ static void ExprException (const ExprEval::Exception& exc, const char* task,
...
+      func += functions.substr(pos2,pos1-pos2);
+    comps.push_back(func);
+    pos2 = pos1 > 0 && pos1 < std::string::npos ? pos1+1 : pos1;
+    pos1 = functions.find("|",pos1+1);
+  }
+
+  return comps;
+}
+
+
```
kmokstad, line 143: Maybe enough with one blank line between the two `}` lines here

kmokstad (approved):
I accept these "modernizations" but with some few suggestions.

src/Utility/ExprFunctions.C
akva2: ok.

src/Utility/ExprFunctions.C
akva2: if i put it in the header we need the definition of the types of the pointers being deleted in the header. which we do not currently have, and i prefer to keep the forwarding. as for the doxy it's still technically correct since we are running the destructors of the unique_ptr's, but i can rephrase it.

src/Utility/ExprFunctions.C
akva2: see above.

src/Utility/ExprFunctions.C
akva2: see above.

src/Utility/ExprFunctions.C
akva2: ok.

akva2:

src/Utility/ExprFunctions.C
kmoks: OK, I was unaware of that. Then keep it in the C-file, we want to have the header-files as slim as possible wrt. includes.

And yes, the destructors will now still delete the heap-allocated items, but it is now done behind the scene (in uniq_ptr) and not in "our" code which the doxy is referring to. This can be said about many of the other "empty" destructors as well. I have used to write only what is done explicitly by it - not what is done via members or parent class implementation. But no big deal - kan leave it for now.

akva2: it's the same as bare pointers, there is nothing magical about unique_ptrs in this respect.

i have rewritten to say defaulted constructors with a details tag to explain why they are not inlined.

PR #640, opened 2023-09-22
RealFunc: add hessian() method

akva2:
For the reasons described in https://github.com/OPM/IFEM/pull/639 this adds the same for the second derivatives.

Sits on top of https://github.com/OPM/IFEM/pull/639

### DIFF
--- a/src/Utility/ExprFunctions.C
+++ b/src/Utility/ExprFunctions.C
@@ -133,7 +133,7 @@ std::pair<size_t,size_t> getNoDims(size_t psize);
 */
 
 template<>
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this with downstreams ifem-stokes=245 ifem-navierstokes=197 ifem-boussinesq=99 ifem-burgers=4 please

akva2:
turns out there was a reason i had voigtIdx as a anon helper and not a lambda..

akva2:

kmokstad (approved):
Yea, that's the downside of reviewing things in bits an pieces: I easily miss couplings like this. So back to file scope then.

PR #638, opened 2023-09-22
ExprFunctions: separate out setNoDims specializations

akva2:
For the same reason as explained in https://github.com/OPM/IFEM/pull/637 separate out the setNoDims implementations so we can use a common implementation in the EvalMultiFunction

Sits on top of https://github.com/OPM/IFEM/pull/637

PR #641, opened 2023-09-22
added: (Tensor|Vec)Func::gradient

akva2:
returns the gradient of the function

Sits on top of https://github.com/OPM/IFEM/pull/638

### DIFF
--- a/src/Utility/Function.h
+++ b/src/Utility/Function.h
@@ -110,6 +110,9 @@ namespace utl
     virtual Result dderiv(const Vec3&, int, int) const { return zero; }
 
   protected:
+    //! \brief Returns the gradient of the function as a 1D array.
+    virtual std::vector<Real> evalGradient(const Vec3&) const { return {}; }
+
     Result zero; //!< Return value for default implementations of derivatives
   };
 }
@@ -250,6 +253,14 @@ public:
   virtual Real getScalarValue(const Vec3& X) const
   {
     return this->evaluate(X).length();
+  }
+
+  //! \brief Evaluates first derivatives of the function.
+  Tensor gradient(const Vec3& X) const
+  {
+    Tensor result(ncmp);
+    result = this->evalGradient(X);
+    return result;
   }
 };
 
--- a/src/Utility/TensorFunction.h
+++ b/src/Utility/TensorFunction.h
@@ -15,6 +15,7 @@
 #define _TENSOR_FUNCTION_H
 
 #include "Function.h"
+#include "matrixnd.h"
 #include "Tensor.h"
 
 
@@ -49,6 +50,9 @@ public:
   {
     return this->evaluate(X).trace();
   }
+
+  //! \brief Evaluates first derivatives of the function.
+  utl::matrix3d<Real> gradient(const Vec3& X) const;
 };
 
 
@@ -59,6 +63,9 @@ public:
 class STensorFunc : public utl::SpatialFunction<SymmTensor>,
                     public FunctionBase
 {
+  //! \brief Returns the flat indices of the symmetric tensor.
+  size_t index(size_t nsd, size_t i, size_t j) const;
+
 protected:
   //! \brief The constructor is protected to allow sub-class instances only.
   STensorFunc(size_t n = 0, bool with33 = false)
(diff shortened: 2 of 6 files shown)


akva2:
jenkins build this with downstreams ifem-stokes=245 ifem-navierstokes=197 ifem-boussinesq=99 ifem-burgers=4 please

akva2:
jenkins build this with downstreams ifem-stokes=245 ifem-navierstokes=197 ifem-boussinesq=99 ifem-burgers=4 please

akva2:
jenkins build this with downstreams ifem-stokes=245 ifem-navierstokes=197 ifem-boussinesq=99 ifem-burgers=4 please

src/Utility/Function.h
```diff
@@ -251,6 +251,21 @@ class VecFunc : public utl::SpatialFunction<Vec3>, public FunctionBase
...
+    result = this->evalGradient(X);
+    return result;
+  }
+
+protected:
+  //! \brief Returns the gradient of the function as a 1D array.
+  virtual std::vector<Real> evalGradient(const Vec3& X) const
+  {
+    return {};
+  }
```
kmokstad, line 268: The argument `X` is not used in this dummy implementation, and I prefer one-liner implementation:
`virtual std::vector<Real> evalGradient(const Vec3&) const { return {}; }`

kmokstad: Another point, couldn't this implementation be placed in the parent class `SpatialFunction` instead? So you don't have to repeat it for each sub-class. Or are there some template issues that makes that difficult.

src/Utility/TensorFunction.h
```diff
@@ -49,6 +50,23 @@ class TensorFunc : public utl::SpatialFunction<Tensor>, public FunctionBase
   {
     return this->evaluate(X).trace();
   }
+
+  //! \brief Evaluates first derivatives of the function.
+  utl::matrix3d<Real> gradient(const Vec3& X) const
+  {
+    const size_t nsd = sqrt(this->dim());
```
kmokstad, line 57: `dim()` just returns the protected member `ncmp` so you could just write `sqrt(ncmp)` here.

src/Utility/TensorFunction.h
```diff
@@ -49,6 +50,23 @@ class TensorFunc : public utl::SpatialFunction<Tensor>, public FunctionBase
...
+    utl::matrix3d<Real> result(nsd, nsd, nsd);
+    result.fill(this->evalGradient(X).data());
+
```
kmokstad, line 60: remove this blank line

src/Utility/TensorFunction.h
```diff
@@ -85,6 +117,28 @@ class STensorFunc : public utl::SpatialFunction<SymmTensor>,
...
+    return result;
+  }
+
+protected:
+  //! \brief Returns the gradient of the function as a 1D array.
+  virtual std::vector<Real> evalGradient(const Vec3& X) const
+  {
+    return {};
+  }
```
kmokstad, line 141: Se previous comments on the `evalGradient()` method.

src/Utility/TensorFunction.h
```diff
@@ -49,6 +50,23 @@ class TensorFunc : public utl::SpatialFunction<Tensor>, public FunctionBase
...
```
kmokstad, line 69: See two comments for VecFunc which applies here as well.

kmokstad (approved):
I think also we should introduce the TensorFunction.C file containing the non-trival member function implementasjons.

src/Utility/Function.h
akva2: Ok and yeah, I initial was working towards removing the SpatialFunction layer but ended up not doing it, so I have moved the slot into that thus dropping the duplicates of the dummy.

src/Utility/TensorFunction.h
akva2: ok.

src/Utility/TensorFunction.h
akva2: ok, other projects wants blank before return so I mix things up sometimes.

akva2:
added TensorFunction.C

akva2:

PR #644, opened 2023-09-25
Deactivate multi-threading for patches with rigid coupling

kmokstad:
That is the last commit, using the oneGroup() method. Hopefully this will remove the data race encountered now and then for the SSUprofil test.

The rest here is some cleaning/fixing of the `getBoundaryElms()` method, including implementation for Lagrange. Notice that this method now also is used when setting up the thread group for boundary integrals. Think it is equivalent to the code it replaces.

### DIFF
--- a/src/ASM/ASMs1D.C
+++ b/src/ASM/ASMs1D.C
@@ -1893,10 +1893,7 @@ void ASMs1D::getElmConnectivities (IntMat& neigh) const
 }
 
 
-void ASMs1D::getBoundaryElms (int lIndex, int, IntVec& elms) const
+void ASMs1D::findBoundaryElms (IntVec& elms, int lIndex, int) const
 {
-  if (lIndex == 1)
-    elms = {0};
-  else
-    elms = {MLGE[nel-1]-1};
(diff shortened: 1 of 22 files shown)


kmokstad:

src/ASM/ASMs1D.C
```diff
@@ -1893,10 +1893,7 @@ void ASMs1D::getElmConnectivities (IntMat& neigh) const
...
 
 
-void ASMs1D::getBoundaryElms (int lIndex, int, IntVec& elms) const
+void ASMs1D::findBoundaryElms (IntVec& elms, int lIndex, int) const
 {
-  if (lIndex == 1)
-    elms = {0};
-  else
-    elms = {MLGE[nel-1]-1};
+  elms = { lIndex == 1 ? 0 : (int)nel-1 };
```
akva2, line 1898: please use c++ cast

akva2 (approved):
Looks good to me

src/ASM/ASMs1D.C
kmokstad: ok, using static_cast

PR #646, opened 2023-09-25
added: allow explicitly specifying time derivative of function

akva2:
Consistent with spatial derivatives.

### DIFF
--- a/src/Utility/ExprFunctions.C
+++ b/src/Utility/ExprFunctions.C
@@ -286,7 +286,7 @@ void EvalFunction::addDerivative (const std::string& function,
                                   const std::string& variables,
                                   int d1, int d2)
 {
-  if (d1 > 0 && d1 <= 3 && d2 < 1) // A first derivative is specified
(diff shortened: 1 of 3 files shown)


akva2:

akva2:

src/Utility/ExprFunctions.C
```diff
@@ -342,6 +342,9 @@ Real EvalFunction::deriv (const Vec3& X, int dir) const
   }
   else if (!IAmConstant)
   {
+    if (gradient[--dir])
+      return gradient[dir]->evaluate(X);
```
kmokstad, line 346: This will crash if someone accidentally invokes this method with dir > 4 (unlike before). So better replace by the equivalent, but safer
```
  if (gradient[3])
    return gradient[3]->evaluate(X);
```

kmokstad (approved):
with a small sanity improvement

src/Utility/ExprFunctions.C
akva2: indeed.

akva2:

PR #643, opened 2023-09-23
added: (Tensor|Vec)Func::hessian 

akva2:
returns the second derivatives of the function

Sits on top of #642

PR #647, opened 2023-09-25
EvalFunc: add automatic differentiation support

akva2:
Sits on top of https://github.com/OPM/IFEM/pull/643

### DIFF
--- a/src/Utility/ExprFunctions.C
+++ b/src/Utility/ExprFunctions.C
@@ -19,8 +19,10 @@
 #include <omp.h>
 #endif
 
+#include <autodiff/reverse/var.hpp>
 
-int EvalFunc::numError = 0;
+
+template<> int EvalFunc::numError = 0;
 
 
 namespace {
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this with downstreams ifem-stokes=245 ifem-navierstokes=197 ifem-boussinesq=99 ifem-burgers=4 ifem-elasticity=150 please

akva2:

kmokstad (approved):
Lots of template magic hwew which is hard to follow, but I trust to does the job.
I assume the `if constexpr (std::is_same_v<Scalar,Real>)` statement is used the check for equality of template arguments?

akva2:
yeah. it's a compile time if. std::is_same_v<a,b> checks if a and b are the same types and in that case returns true, else false. since it's compile time, this means only one of the code paths are chosen when instancing the template for that type.

kmokstad:
right, which only could be achived using some #ifdef expressions before modern times. Neat

PR #648, opened 2023-09-25
EvalMultiFunction: add automatic differentiation support

akva2:
Sits on top of https://github.com/OPM/IFEM/pull/647

### DIFF
--- a/src/Utility/ExprFunctions.C
+++ b/src/Utility/ExprFunctions.C
@@ -201,7 +201,7 @@ int voigtIdx (int d1, int d2)
 
 
 template<class Scalar>
-EvalFuncImpl<Scalar>::EvalFuncImpl (const char* function, const char* x, Real eps)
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this with downstreams ifem-stokes=245 ifem-navierstokes=197 ifem-boussinesq=99 ifem-burgers=4 ifem-elasticity=150 please

kmokstad (approved):
Looks OK. However, I'm not too fond of class names `<whatever>Impl` only for making unique names. (All classes implement something.) Propose the following names instead:

- EvalFuncScalar instead of EvalFuncImpl
- EvalFuncSpatial instead of EvalFunctionImpl

Then the names more indicate that the former is function with a scalar argument, whereas the latter has a spatial argument. You can still keep the type aliases to avoid effects elsewhere.

akva2:
Fair enough, I'm not at all tied to the impl.

For the record it is meant to imply that it's the implementation, not that it implements something :).

akva2:

PR #649, opened 2023-09-27
Analytic solutions: Allow derivation through AD or FD

akva2:
This allows deriving secondary solutions from the primary solution using AD or FD.

Sits on top of https://github.com/OPM/IFEM/pull/648

### DIFF
--- a/src/Utility/AnaSol.C
+++ b/src/Utility/AnaSol.C
@@ -17,11 +17,49 @@
 #include "Utilities.h"
 #include "IFEM.h"
 #include "tinyxml.h"
+
+#include <autodiff/reverse/var.hpp>
+
 #ifdef HAS_HDF5
-#include "ProcessAdm.h"
-#include "HDF5Writer.h"
 #include "FieldFunctions.h"
 #endif
+
+
+namespace {
+
+//! \brief Function returning the gradient of a scalar or vector function.
+template<class FromFunc, class ToFunc, class Ret>
+class GradientFunc : public ToFunc
+{
+public:
+  //! \brief The constructor initializes the function to use.
+  GradientFunc(const FromFunc& f) : func(f) {}
+
+protected:
+  //! \brief Evaluates the gradient in a point.
+  Ret evaluate(const Vec3& X) const { return func.gradient(X); }
+
+private:
+  const FromFunc& func; //!< Reference to scalar/vector function
+};
+
+
+//! \brief Specialization for symmetric tensors.
+//! \details We need to construct a symmtensor from the returned tensor.
+template<>
+SymmTensor GradientFunc<VecFunc,STensorFunc,SymmTensor>::evaluate(const Vec3& X) const
+{
+  Tensor tmp = func.gradient(X);
+  const size_t nsd = tmp.dim();
+  SymmTensor ret(nsd);
+  for (size_t i = 1; i <= nsd; ++i)
+    for (size_t j = i; j <= nsd; ++j)
+      ret(i,j) = tmp(i,j);
+
+  return ret;
+}
+
+}
 
 
 AnaSol::AnaSol (RealFunc* s1, VecFunc* s2,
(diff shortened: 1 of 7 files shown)


akva2:
jenkins build this with downstreams ifem-stokes=245 ifem-navierstokes=197 ifem-boussinesq=99 ifem-burgers=4 ifem-elasticity=151 ifem-poisson=57 please

akva2:
jenkins build this with downstreams ifem-elasticity=151 please

akva2:
jenkins build this with downstreams ifem-elasticity=151 please

akva2:
jenkins build this with downstreams ifem-elasticity=151 please

akva2:
jenkins build this with downstreams ifem-elasticity=151 please

akva2:
jenkins build this with downstreams ifem-elasticity=151 please

src/Utility/AnaSol.C
```diff
@@ -17,13 +17,54 @@
...
+public:
+  //! \brief The constructor initializes the function to use.
+  GradientFunc(const FromFunc& f) : func(f) {}
+
+protected:
+  //! \brief Evaluates the gradient in a point.
+  Ret evaluate(const Vec3& X) const
+  {
+    return func.gradient(X);
+  }
```
kmokstad, line 43: This single-statement method can also be written on one line.

src/Utility/AnaSol.C
```diff
@@ -17,13 +17,54 @@
...
+
+
+//! \brief Specialization for symmetric tensors.
+//! \details We need to construct a symmtensor from the returned tensor.
+template<>
+SymmTensor GradientFunc<VecFunc,STensorFunc,SymmTensor>::evaluate(const Vec3& X) const
+{
+  Tensor tmp = func.gradient(X);
+  const size_t nsd = tmp.dim();
+  SymmTensor ret(nsd, nsd == 2 && tmp.size() == 4);
```
kmokstad, line 57: This is a bit confusing. `func.gradient()` will here always return a standard tensor (2x2, size=4) or (3x3, size=9), afaik. So in the 2D case you are here constructing a symmetric tensor with `with33=true`, which is intended for axisymmetric stress tensors only. That does not look right. Maybe you intended to symmetrize the sensor instead, i.e., invoke the `symmetrize()` method?

src/Utility/ExprFunctions.h
```diff
@@ -99,6 +99,8 @@ class EvalFuncSpatial : public RealFunc
     Scalar* z; //!< Z-coordinate
     Scalar* t; //!< Time
 
+    //! \brief Returns a const ref to a member.
+    //! !\param dir One-based index to member
```
kmokstad, line 103: Doxy error, an `!` too much

kmokstad (changes requested):
Check the SymmTensor specialization for GradientFunc::evaluate()

src/Utility/ExprFunctions.h
```diff
@@ -99,6 +99,8 @@ class EvalFuncSpatial : public RealFunc
...
     const Scalar& get(int dir) const
     {
```
kmokstad, line 105: (This comment applies to Line 107 below): Isn't it proper to have the default: case as the last case and not the first? Guess it is because you want *x to be the default, but I also thought such fall-through case statements (without break or return before the next), generated annoying warnings in the later compilers. Or have they perhaps explicitly been deactivated in the build set-up.

akva2: Empty fallthroughs do no warn. There is no proper or not about where you put the default label, but if you want I can of course put default at the end or duplicate the line.

src/Utility/AnaSol.C
```diff
@@ -17,13 +17,54 @@
...
+#include <autodiff/reverse/var.hpp>
+
 #ifdef HAS_HDF5
-#include "ProcessAdm.h"
-#include "HDF5Writer.h"
 #include "FieldFunctions.h"
 #endif
 
 
+namespace {
```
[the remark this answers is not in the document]
akva2: The symbols will be exported from the compile unit and can cause clashes. As GradientFunc isn't entirely exotic I prefer to be on the safe side.

src/Utility/AnaSol.C
akva2: Ok

src/Utility/ExprFunctions.h
akva2: ok.

akva2:
jenkins build this with downstreams ifem-elasticity=151 please

kmokstad (approved):

src/Utility/AnaSol.C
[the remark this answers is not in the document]
kmokstad: OK, in the case someone defines a GradientFunc elsewhere. Then keep it. No problem.

PR #650, opened 2023-10-16
EqualOrderOperators: cosmetics

akva2:
also use MxV instead of loops in WeakOps::Divergence and ResidualOps::Divergence


akva2:

PR #651, opened 2023-10-16
vector::add: also expose stridey and ofsy

akva2:
I need these as well to tidy up some code in NavierStokes.

### DIFF
--- a/src/LinAlg/matrix.h
+++ b/src/LinAlg/matrix.h
@@ -129,7 +129,8 @@ namespace utl //! General utility classes and functions.
     vector<T>& operator-=(const vector<T>& X) { return this->add(X,T(-1)); }
     //! \brief Add the given vector \b X scaled by \a alfa to \a *this.
     vector<T>& add(const std::vector<T>& X, const T& alfa = T(1),
-                   int ofsx = 0, int stridex = 1);
+                   unsigned int ofsx = 0, int stridex = 1,
+                   unsigned int ofsy = 0, int stridey = 1);
 
     //! \brief Perform \f${\bf Y} = \alpha{\bf Y} + (1-\alpha){\bf X} \f$
     //! where \b Y = \a *this.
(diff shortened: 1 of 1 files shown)


akva2:

src/LinAlg/matrix.h
```diff
@@ -1367,11 +1370,13 @@ namespace utl //! General utility classes and functions.
 
   template<class T> inline
   vector<T>& vector<T>::add(const std::vector<T>& X, const T& alfa,
-                            int ofsx, int stridex)
+                            int ofsx, int stridex,
+                            int ofsy, int stridey)
   {
-    T* p = this->ptr();
+    T* p = this->ptr() + ofsy;
```
kmokstad, line 1376: Now use `data()` instead of `ptr()` here as well.

src/LinAlg/matrix.h
```diff
@@ -1367,11 +1370,13 @@ namespace utl //! General utility classes and functions.
...
     const T* q = X.data() + ofsx;
-    for (size_t i = 0; i < this->size() && i < X.size(); i++, p++, q += stridex)
+    size_t size = std::min(this->size() / stridey, X.size() / stridex);
+    for (size_t i = 0; i < size; ++i, p += stridey, q += stridex)
```
kmokstad, line 1379: If either ofsx or ofsy are non-zero (positive), I think the loop bounds will be wrong here?

src/LinAlg/matrix.h
```diff
@@ -129,7 +129,8 @@ namespace utl //! General utility classes and functions.
     vector<T>& operator-=(const vector<T>& X) { return this->add(X,T(-1)); }
     //! \brief Add the given vector \b X scaled by \a alfa to \a *this.
     vector<T>& add(const std::vector<T>& X, const T& alfa = T(1),
-                   int ofsx = 0, int stridex = 1);
+                   int ofsx = 0, int stridex = 1,
+                   int ofsy = 0, int stridey = 1);
```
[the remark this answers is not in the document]
akva2: changed to unsigned.

src/LinAlg/matrix.h
akva2: there's a theoretical issue yes, but not with sane input data. the default blas implementations will crash in the same way so i think i'll just keep the assumptions on input data.

akva2:

src/LinAlg/matrix.h
kmokstad: ok, since this is only for the (never used) no-blas fallback anyway

src/LinAlg/matrix.h
```diff
@@ -129,7 +129,8 @@ namespace utl //! General utility classes and functions.
     vector<T>& operator-=(const vector<T>& X) { return this->add(X,T(-1)); }
     //! \brief Add the given vector \b X scaled by \a alfa to \a *this.
     vector<T>& add(const std::vector<T>& X, const T& alfa = T(1),
-                   int ofsx = 0, int stridex = 1);
+                   unsigned int ofsx = 0, unsigned int stridex = 1,
+                   unsigned int ofsy = 0, unsigned int stridey = 1);
```
kmokstad, line 133: Actually, negative stride should be allowed (in combination of non-negative offset), in case someone wants to loop backward in an array. Think some BLAS routines accounts for that at least.

akva2: right. changed back to signed.

akva2:

PR #653, opened 2023-10-24
Added: Option to save result components on grid files

kmokstad:
Firstly, a new `<grid>` tag within the `<resultpoints>` scope is added, to enable easy definition of result points in a grid pattern. Each result component (primary and secondary variables) is the saved to separate files, one column for each grid point (the physical time as the first column) and the time/load steps as rows. Then these files can be used as input to the python script for calculation of optimal sensor placement.

### DIFF
--- a/src/SIM/SIMoutput.C
+++ b/src/SIM/SIMoutput.C
@@ -37,6 +37,7 @@ SIMoutput::SIMoutput (IntegrandBase* itg) : SIMinput(itg)
   myGeomID = 0;
   myVtf = nullptr;
   logRpMap = false;
+  idxGrid = -1;
 }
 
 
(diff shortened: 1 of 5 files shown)


src/SIM/SIMoutput.C
```diff
@@ -1919,6 +1928,50 @@ bool SIMoutput::savePoints (const Vector& psol, double time, int step) const
...
+      return false;
+
+  if (!sol.augmentRows(sol2))
+    return false;
+
+  // Write one file for each result component
+  for (size_t r = 1; r <= sol.rows(); r++)
+  {
+    std::string fName = fileName + "_" + std::to_string(r) + fileExt;
+    std::ofstream fs(fName.c_str(), step == 1 ? std::ios::out : std::ios::app);
```
akva2, line 1964: ofstream has a string constructor since c++11.

akva2 (approved):
Looks fine to me. I'd take the opportunity to replace typedefs with using but I guess it would be inconsistent with the other statements in the file.

src/SIM/SIMoutput.C
kmokstad: Changed to string constructor here and elsewhere in this file

kmokstad:
jenkins build this with downstreams ifem-elasticity=152 please

kmokstad:
jenkins build this with downstreams ifem-elasticity=152 please

PR #654, opened 2023-10-24
Clang warning and compilation fixes

akva2:
One compilation fix (missing <array> include) and various warning quells.

### DIFF
--- a/Apps/Common/SIMSolver.h
+++ b/Apps/Common/SIMSolver.h
@@ -155,7 +155,7 @@ public:
     this->printHeading(heading);
 
     // Solve for each time step up to final time
-    for (int iStep = 1; this->advanceStep(); iStep++)
+    while (this->advanceStep())
       if (!this->S1.solveStep(tp))
         return saveDivergedSol && !this->S1.saveStep(tp,nBlock) ? 4 : 3;
       else if (!this->saveState(geoBlk,nBlock))
--- a/src/ASM/ASMutils.C
+++ b/src/ASM/ASMutils.C
@@ -176,7 +176,11 @@ bool ASM::readXML (std::istream& is, IntMat& MNPC, std::vector<Vec3>& nodes,
   {
     Vec3 X;
     std::istringstream iss(data);
+#if SP_DEBUG > 1
     for (size_t inod = 0; iss; inod++)
+#else
+    while (iss)
+#endif
     {
       iss >> X;
       if (iss)
@@ -194,7 +198,11 @@ bool ASM::readXML (std::istream& is, IntMat& MNPC, std::vector<Vec3>& nodes,
   {
     IntVec mnpc(nenod);
     std::istringstream iss(data);
+#if SP_DEBUG > 1
     for (size_t iel = 0; iss; iel++)
+#else
+    while (iss)
+#endif
     {
       for (int& n : mnpc) iss >> n;
       if (iss)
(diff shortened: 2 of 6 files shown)


akva2:

Apps/Common/SIMSolver.h
```diff
@@ -155,7 +155,7 @@ template<class T1> class SIMSolver : public SIMSolverStat<T1>
     this->printHeading(heading);
 
     // Solve for each time step up to final time
-    for (int iStep = 1; this->advanceStep(); iStep++)
+    for (; this->advanceStep(); )
```
kmokstad, line 158: Then rather use `while (this->advanceStep())` instead

src/ASM/ASMutils.C
```diff
@@ -176,7 +176,7 @@ bool ASM::readXML (std::istream& is, IntMat& MNPC, std::vector<Vec3>& nodes,
   {
     Vec3 X;
     std::istringstream iss(data);
-    for (size_t inod = 0; iss; inod++)
+    for ([[maybe_unused]] size_t inod = 0; iss; inod++)
```
kmokstad, line 179: This was a new thing I did not know about, but think I would prefer the (maybe old-fashioned):
```
#if SP_DEBUG > 1
  for (size_t inod = 0; iss; inod++)
#else
  while (iss)
#endif
```
However, I don't insist.

src/ASM/ASMutils.C
```diff
@@ -194,7 +194,7 @@ bool ASM::readXML (std::istream& is, IntMat& MNPC, std::vector<Vec3>& nodes,
   {
     IntVec mnpc(nenod);
     std::istringstream iss(data);
-    for (size_t iel = 0; iss; iel++)
+    for ([[maybe_unused]] size_t iel = 0; iss; iel++)
```
kmokstad, line 197: same

src/ASM/BasisFunctionCache.C
```diff
@@ -13,6 +13,8 @@
 
 #include "BasisFunctionCache.h"
 
+#include <array>
```
kmokstad, line 16: But `std::array` is also referred in the header file, so it makes more sense to include it there instead in my view.

src/Utility/HDF5Restart.C
```diff
@@ -69,8 +69,13 @@ bool HDF5Restart::writeData (const SerializeData& data)
 #endif
 
   // Lambda function for writing data to the HDF5 file.
-  auto&& writeGroup = [this,pid,ptot] (hid_t group, const std::string& name,
-                                       int len, const void* data, hid_t type)
+#ifdef HAVE_MPI
+  auto&& writeGroup = [this,pid,ptot]
```
kmokstad, line 73: Hmm, to me it looks like `this` and `ptot` also are referred outside the `#ifdef HAVE_MPI`. Have you tested this with hdf5 support?

kmokstad (approved):
Mostly ok, except for a few questions

src/Utility/HDF5Restart.C
akva2: it's not. it is used in the for immediately after the lambda definition so maybe you got confused by that?

Apps/Common/SIMSolver.h
akva2: alright. you have preferred butchered for's before but i'll gladly use while :)

src/ASM/ASMutils.C
akva2: okay.

src/ASM/ASMutils.C
akva2: okay.

akva2:

src/Utility/HDF5Restart.C
kmokstad: right, the use was only outside the lambda I can se now

PR #664, opened 2023-12-04
Dump system matrices and RHS-vector in the start of each load/time step only

kmokstad:
Instead of dumping them every iteration which is not that useful for Nonlinear/Newmark simulations.
But consider that as an option later, if/when needed.


akva2 (approved):

kmokstad:

PR #669, opened 2024-01-04
fixed: reset error array to appropriate size

akva2:
we have to use getNodes(1) to get the proper size. In general this resulted in the wrong number of functions being refined
for single basis methods and mixed methods projecting on the first basis. It also lead to crashes if 100% of function were to be refined since it selected too many functions / non-existing functions for refinement.


akva2:
jenkins build this with downstreams ifem-stokes=260 ifem-boussinesq=105 please

kmokstad (approved):

PR #670, opened 2024-01-04
ASMuxD::evalSolution(P): do not use MNPC directly if there are multipliers

akva2:
We cannot use the MNPC directly if there are additional DOFs in the model.

### DIFF
--- a/src/ASM/LR/ASMu2D.C
+++ b/src/ASM/LR/ASMu2D.C
@@ -2038,8 +2038,8 @@ bool ASMu2D::evalSolution (Matrix& sField, const Vector& locSol,
 {
   PROFILE2("ASMu2D::evalSol(P)");
 
-  size_t nComp = locSol.size() / this->getNoNodes();
-  if (nComp*this->getNoNodes() != locSol.size())
+  size_t nComp = locSol.size() / this->getNoNodes(-1);
+  if (nComp*this->getNoNodes(-1) > locSol.size())
     return false;
 
   size_t nPoints = gpar[0].size();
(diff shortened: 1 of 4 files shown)


akva2:
jenkins build this please

src/ASM/LR/ASMu2D.C
```diff
@@ -2066,6 +2066,8 @@ bool ASMu2D::evalSolution (Matrix& sField, const Vector& locSol,
     fe.v = gpar[1][i];
     int iel = lrspline->getElementContaining(fe.u,fe.v);
 
+    const LR::Element* el = lrspline->getElement(iel);
+
```
kmokstad, line 2070: Should perhaps move this to after the `if (iel != lel)` check, just in case `lrspline->getElement()` does not handle an invalid argument gracefully.

src/ASM/LR/ASMu3D.C
```diff
@@ -1542,12 +1542,14 @@ bool ASMu3D::evalSolution (Matrix& sField, const Vector& locSol,
     // Fetch element containing evaluation point.
     // Sadly, points are not always ordered in the same way as the elements.
     int iel = lrspline->getElementContaining(gpar[0][i],gpar[1][i],gpar[2][i]);
+    const LR::Element* el = lrspline->getElement(iel);
```
kmokstad, line 1545: Should perhaps move this to _after_ the `if (iel < 0)` check, just in case `lrspline->getElement()` does not handle a negative argument gracefully

akva2: yup

akva2:

PR #673, opened 2024-02-14
Fix xml include files

akva2:
"Small" bug introduced when we moved to tinyxml2. After include was injected a closing tag was added for the currently processed element. This meant we got an early closing tag which broke the structure of the XML document.

### DIFF
--- a/src/SIM/Test/with_include.xml
+++ b/src/SIM/Test/with_include.xml
@@ -1,3 +1,4 @@
 <simulation>
   <include>src/SIM/Test/included.xml</include>
+  <someothertag>is_here</someothertag>
 </simulation>
--- a/src/SIM/XMLInputBase.C
+++ b/src/SIM/XMLInputBase.C
@@ -48,7 +48,6 @@ struct IncludeInjector : public tinyxml2::XMLVisitor
   bool VisitEnter(const tinyxml2::XMLElement& elem,
                   const tinyxml2::XMLAttribute* attribute) override
   {
-    tinyxml2::XMLElement* e;
     if (std::string(elem.Value()) == "include") {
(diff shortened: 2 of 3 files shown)


akva2:

src/SIM/Test/with_include.xml
```diff
@@ -1,3 +1,4 @@
 <simulation>
   <include>src/SIM/Test/included.xml</include>
+   <someothertag>is_here</someothertag>
```
kmokstad, line 3: inconsistent indent also here

src/SIM/XMLInputBase.C
```diff
@@ -56,7 +56,7 @@ struct IncludeInjector : public tinyxml2::XMLVisitor
           return false;
       }
       currElem->InsertEndChild(inc_doc.RootElement()->DeepClone(&new_doc));
-      include_found = true;
+      include_found = include_processed = true;
     } else {
       e = new_doc.NewElement(elem.Name());
```
kmokstad, line 61: The temporary `e` is defined outside this block but only (afics) used inside it. Wouldn't that cause some static code check warnings?

kmokstad (approved):
Looks fine. Although I found it slightly confusing with the two bool variables plus the pointer to judge the status of the include tag. Maybe one int (or char) with value 0, 1 and 2 instead would be more clear? But that's up to you.

src/SIM/Test/with_include.xml
akva2: fixed.

src/SIM/XMLInputBase.C
akva2: possibly. reduced the scope.

akva2:

akva2:
As one of the bools are user facing to I prefer to keep the two variables since the value convention would leak outside the class. I marked one of them private though.

PR #677, opened 2024-03-04
added: implement BasisFunctionCache for ASMu2DLag

### LINKED ISSUE #676: Lagrange patches without an underlying spline object no longer work
After the introduction of the basis function cache feature in #571, it is no longer possible to run models with Lagrange meshes where the mesh is read from a file instead of deriving it from a spline object. We had such a feature in reading matlab files, but without full unit test (we test reading the file itself and setting up the FE datastructure, but not integrating the FE matrices). Therefore this probably slipped.

Try the input file in [L1-50x50.zip](https://github.com/OPM/IFEM/files/14476969/L1-50x50.zip) with LinEl, and it crashes due to de-referencing a null spline object pointer when setting up the cache in ASMs2DLag::integrate(). Not sure how to fix this properly, therefore opening the issue.

akva2:
Closes #676

### DIFF
--- a/src/ASM/ASMs2D.C
+++ b/src/ASM/ASMs2D.C
@@ -1718,7 +1718,7 @@ bool ASMs2D::integrate (Integrand& integrand,
   if (myCache.empty())
     myCache.emplace_back(std::make_unique<BasisFunctionCache>(*this, cachePolicy, 1));
 
-  BasisFunctionCache& cache = *myCache.front();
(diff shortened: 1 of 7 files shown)


akva2:
jenkins build this with downstreams ifem-elasticity=158 please

kmokstad (approved):
Approved, under the condition that the ongoing test completes with no other failures.

akva2:
jenkins build this with downstreams ifem-elasticity=158 please

PR #679, opened 2024-03-16
Added: Alternative constructors for LinearFunc and Interpolate1D

kmokstad:
For the convenience, a constructor taking arrays of the x- and y-values is provided (second commit).
The first commit resolves some missing override of `evalSolution()` methods, after  introduction of the Piola transformation stuff.

The fourth commit adds the `ASMs1D::findElement()` method, used to locate the element and local coordinates of a spatial point on a 1D patch. Although the unit test passes, I think it reveals a bug in GoTools for the quadratic case: When searching for a point lying after the end, i.e., its parameter value should be higher that `endParam()`, it correctly returns the coordinates of the last control point, but the distance seems totally wrong. You'll see that when activating the debug print.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,7 +162,7 @@ else()
                          src/ASM/LocalIntegral.h src/ASM/SAMpatch.h
                          src/ASM/SplineField?D.h src/ASM/SplineFields?D.h
                          src/ASM/TimeDomain.h src/ASM/ASMs?D.h src/ASM/ASM?D.h
-                         src/ASM/ASMs?DLag.h
+                         src/ASM/ASMs?DLag.h src/ASM/ASMutils.h
                          src/ASM/DomainDecomposition.h src/ASM/ItgPoint.h
                          src/ASM/ReactionsOnly.h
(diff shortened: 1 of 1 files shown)


kmokstad:

src/ASM/ASMs1DLag.C
```diff
@@ -7,7 +7,7 @@
 //!
 //! \author Einar Christensen / SINTEF
 //!
-//! \brief Driver for assembly of 1D Lagrange FE models.
+//! \brief Driver for assembly of 1D %Lagrange FE models.
```
akva2, line 10: what does this do?

akva2 (approved):
Seems fine.

src/ASM/ASMs1DLag.C
kmokstad: Without it, doxygen will insert a hyperlink to the class with that name (Lagrange).

kmokstad:

kmokstad:

kmokstad:

kmokstad:

PR #680, opened 2024-03-18
ResidualOps: flip sign of Laplacian term

akva2:
this was inconsistent with every other operator where the negation is implicit


akva2:
jenkins build this with downstreams ifem-burgers=9 ifem-boussinesq=106 ifem-navierstokes=210 ifem-stokes=262 ifem-advectiondiffusion=81 please

PR #681, opened 2024-03-18
ResidualOps: flip sign of Divergence term

akva2:
while technically this makes no difference, it is
better to be consistent


akva2:
jenkins build this with downstreams ifem-stokes=263 ifem-navierstokes=211 please

PR #683, opened 2024-03-21
Fix issues disclosed by debug-iterator build

akva2:
Fixes for issues disclosed by debug-iterator builds.

Yes, will be squashed and commit messages pruned. Just for keeping track of things for now.

### DIFF
--- a/src/ASM/ASMs1D.C
+++ b/src/ASM/ASMs1D.C
@@ -860,9 +860,14 @@ double ASMs1D::findPoint (Vec3& X, double* param) const
 
   // Find the closest point on the spline curve
   double dist;
-  Go::Point Xpt(X.x,X.y,X.z), Xfound;
-  curv->ParamCurve::closestPoint(Xpt, *param, Xfound, dist);
-  for (int i = 0; i < curv->dimension(); i++) X[i] = Xfound[i];
+  Go::Point Xfound;
+  if (curv->dimension() == 3)
+    curv->ParamCurve::closestPoint(Go::Point(X.x,X.y,X.z), *param, Xfound, dist);
+  else
+    curv->ParamCurve::closestPoint(Go::Point(X.x,X.y), *param, Xfound, dist);
+  for (int i = 0; i < curv->dimension(); ++i)
+    X[i] = Xfound[i];
+
   return dist;
 }
 
@@ -912,8 +917,11 @@ std::pair<int,double> ASMs1D::findElement (const Vec3& X) const
   // Find the closest point on the spline curve
   int iknot, ielno;
   double param, dist;
-  Go::Point Xtarget(X.x,X.y,X.z), Xfound;
-  curv->ParamCurve::closestPoint(Xtarget,param,Xfound,dist);
+  Go::Point Xfound;
+  if (curv->dimension() == 3)
+    curv->ParamCurve::closestPoint(Go::Point(X.x,X.y,X.z),param,Xfound,dist);
+  else
+    curv->ParamCurve::closestPoint(Go::Point(X.x,X.y),param,Xfound,dist);
 
   // Check if point is inside parameter domain
   if (param <= curv->startparam() && dist > 0.0)
--- a/src/ASM/FiniteElement.C
+++ b/src/ASM/FiniteElement.C
@@ -87,7 +87,9 @@ bool MxFiniteElement::Jacobian (Matrix& Jac, const Matrix& Xnod,
   if (separateGeometry)
     gBasis = bfs.size();
 
-  detJxW = utl::Jacobian(Jac, this->grad(gBasis), Xnod,
+  // Make sure we point to a valid matrix with separate geometry,
+  // even if it is not used
+  detJxW = utl::Jacobian(Jac, this->grad(separateGeometry ? 1 : gBasis), Xnod,
                          bfs[gBasis-1]->dNdu,
                          !separateGeometry);
   if (detJxW == 0.0) return false; // singular point
--- a/src/LinAlg/matrix.h
+++ b/src/LinAlg/matrix.h
@@ -1007,7 +1007,8 @@ namespace utl //! General utility classes and functions.
                                     unsigned int ofsy, int stridey)
   {
     size_t n = this->size() < X.size() ? this->size() : X.size();
-    cblas_saxpy(n,alfa,X.data()+ofsx,stridex,this->data()+ofsy,stridey);
+    if (n > 0)
+      cblas_saxpy(n,alfa,X.data()+ofsx,stridex,this->data()+ofsy,stridey);
     return *this;
   }
 
@@ -1018,7 +1019,8 @@ namespace utl //! General utility classes and functions.
                                       unsigned int ofsy, int stridey)
   {
     size_t n = this->size() < X.size() ? this->size() : X.size();
-    cblas_daxpy(n,alfa,X.data()+ofsx,stridex,this->data()+ofsy,stridey);
+    if (n > 0)
+      cblas_daxpy(n,alfa,X.data()+ofsx,stridex,this->data()+ofsy,stridey);
     return *this;
   }
 
@@ -1027,7 +1029,8 @@ namespace utl //! General utility classes and functions.
                                             const float& alfa)
   {
     size_t n = this->size() < A.size() ? this->size() : A.size();
-    cblas_saxpy(n,alfa,A.ptr(),1,this->ptr(),1);
+    if (n > 0)
+      cblas_saxpy(n,alfa,A.ptr(),1,this->ptr(),1);
     return *this;
   }
 
@@ -1036,7 +1039,8 @@ namespace utl //! General utility classes and functions.
                                               const double& alfa)
   {
     size_t n = this->size() < A.size() ? this->size() : A.size();
-    cblas_daxpy(n,alfa,A.ptr(),1,this->ptr(),1);
+    if (n > 0)
+      cblas_daxpy(n,alfa,A.ptr(),1,this->ptr(),1);
     return *this;
   }
 
@@ -1392,9 +1396,9 @@ namespace utl //! General utility classes and functions.
                             unsigned int ofsx, int stridex,
                             unsigned int ofsy, int stridey)
   {
-    T* p = this->data() + ofsy;
-    const T* q = X.data() + ofsx;
     size_t size = std::min(this->size() / stridey, X.size() / stridex);
+    T* p = this->elem.data() + ofsy;
+    const T* q = X.elem.data() + ofsx;
     for (size_t i = 0; i < size; ++i, p += stridey, q += stridex)
       *p += alfa*(*q);
     return *this;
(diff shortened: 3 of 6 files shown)


akva2:

akva2:

src/LinAlg/matrix.h
```diff
@@ -1007,6 +1007,8 @@ namespace utl //! General utility classes and functions.
                                     unsigned int ofsy, int stridey)
   {
     size_t n = this->size() < X.size() ? this->size() : X.size();
+    if (n == 0)
+      return *this;
```
kmokstad, line 1017: better with
```
if (n > 0)
  cblas_...(...);
```
such that you don't need the extra return. Same with the following 3 changes below.

src/LinAlg/matrix.h
```diff
@@ -1392,9 +1400,11 @@ namespace utl //! General utility classes and functions.
                             unsigned int ofsx, int stridex,
                             unsigned int ofsy, int stridey)
   {
+    size_t size = std::min(this->size() / stridey, X.size() / stridex);
+    if (size == 0)
+      return *this;
```
kmokstad, line 1411: Is this early return necessary here? Since no blas-call involved or `ptr()` which I though was the problem (`data()` should be allowed also for empty vectors, al long you don't try to dereference the result.

src/LinAlg/matrix.h
```diff
@@ -1403,9 +1413,12 @@ namespace utl //! General utility classes and functions.
   template<class T> inline
   matrixBase<T>& matrixBase<T>::add(const matrixBase<T>& A, const T& alfa)
   {
+    size_t size = std::min(this->size(),A.size());
+    if (size == 0)
+      return *this;
```
kmokstad, line 1424: Same here, you you replace the below with
```
T* p = this->elem.data();
const T* q = A.elem.data();

File: src/Utility/Legendre.C [hunk 4]
```diff
@@ -35,7 +35,7 @@ bool Legendre::GL (RealArray& weights, RealArray& points, int n)
     A(n,n-1) = Real(n-1)/Real(2*n-1);
   }
 
-  RealArray eig_complex, evalpoints;
+  RealArray eig_complex(n), evalpoints(n);
```
kmokstad, line 38: why? this resize is performed inside `solveEigNon()` afics..

src/Utility/HDF5Writer.C
```diff
@@ -359,8 +359,10 @@ void HDF5Writer::writeSIM (int level, const DataEntry& entry,
     else
       sim->extractPatchSolution(glbVec,pchVec,pch,ncmp,1);
 
-    field.resize(ncmp,pchVec.size()/ncmp);
-    field.fill(pchVec.ptr());
+    if (ncmp > 0) {
+      field.resize(ncmp,pchVec.size()/ncmp);
+      field.fill(pchVec.ptr());
+    }
```
kmokstad, line 365: would be better with an early exit if (ncmp == 0) after L351 as the stuff in between will not anything useful in that case.

src/LinAlg/matrix.h
```diff
@@ -171,6 +171,8 @@ namespace utl //! General utility classes and functions.
           size_t off1 = 0, int inc1 = 1,
           size_t off2 = 0, int inc2 = 1) const
     {
+      if (this->empty() || v.empty())
+        return 0.0;
       return this->dot(&v.front(),v.size(),off1,inc1,off2,inc2);
```
kmokstad, line 176: use `v.data()` instead of `&v.front()` and the early exit wouldn't/shouldn't be neccessary

src/LinAlg/matrix.h
```diff
@@ -937,6 +939,8 @@ namespace utl //! General utility classes and functions.
     int n1 = i1 > 1 || i1 < -1 ? this->size()/abs(i1) : this->size()-o1;
     int n2 = i2 > 1 || i2 < -1 ? nv/abs(i2) : nv-o2;
     int n  = n1 < n2 ? n1 : n2;
+    if (n == 0)
+      return 0.0;
     return cblas_sdot(n,this->ptr()+o1,i1,v+o2,i2);
```
kmokstad, line 944: same, use `this->data()` instead of `ptr()`

src/LinAlg/matrix.h
```diff
@@ -947,6 +951,8 @@ namespace utl //! General utility classes and functions.
     int n1 = i1 > 1 || i1 < -1 ? this->size()/abs(i1) : this->size()-o1;
     int n2 = i2 > 1 || i2 < -1 ? nv/abs(i2) : nv-o2;
     int n  = n1 < n2 ? n1 : n2;
+    if (n == 0)
+      return 0.0;
     return cblas_ddot(n,this->ptr()+o1,i1,v+o2,i2);
```
kmokstad, line 956: and same

src/ASM/FiniteElement.C
```diff
@@ -87,7 +87,12 @@ bool MxFiniteElement::Jacobian (Matrix& Jac, const Matrix& Xnod,
   if (separateGeometry)
     gBasis = bfs.size();
 
-  detJxW = utl::Jacobian(Jac, this->grad(gBasis), Xnod,
+  // Make sure we point to a valid matrix with separate geometry,
+  // even if it is not used
+  Matrix& dNdX = separateGeometry ? this->grad(1)
+                                  : this->grad(gBasis);
+
+  detJxW = utl::Jacobian(Jac, dNdX, Xnod,
```
kmokstad, line 95: Replace by more compact
```
detJxW = utl::Jacobian(Jac, this->grad(separateGeometry ? 1 : gBasis), Xnod,
```

src/Utility/Legendre.C
akva2: As the links in commit message shows, it triggers in tests.

src/ASM/ASMs1D.C
```diff
@@ -860,7 +860,9 @@ double ASMs1D::findPoint (Vec3& X, double* param) const
 
   // Find the closest point on the spline curve
   double dist;
-  Go::Point Xpt(X.x,X.y,X.z), Xfound;
+  Go::Point Xpt = curv->dimension() == 3 ? Go::Point(X.x,X.y,X.z)
+                                         : Go::Point(X.x,X.y);
+  Go::Point Xfound;
   curv->ParamCurve::closestPoint(Xpt, *param, Xfound, dist);
```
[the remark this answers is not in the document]
akva2: ok

src/ASM/ASMs1D.C
```diff
@@ -914,7 +914,9 @@ std::pair<int,double> ASMs1D::findElement (const Vec3& X) const
   // Find the closest point on the spline curve
   int iknot, ielno;
   double param, dist;
-  Go::Point Xtarget(X.x,X.y,X.z), Xfound;
+  Go::Point Xtarget = curv->dimension() == 3 ? Go::Point(X.x,X.y,X.z)
+                                             : Go::Point(X.x,X.y);
+  Go::Point Xfound;
   curv->ParamCurve::closestPoint(Xtarget,param,Xfound,dist);
```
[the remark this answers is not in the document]
akva2: ok

src/LinAlg/matrix.h
akva2: ok

src/LinAlg/matrix.h
akva2: used data()

src/LinAlg/matrix.h
akva2: ok

src/Utility/HDF5Writer.C
akva2: ok

src/LinAlg/matrix.h
akva2: ok

src/LinAlg/matrix.h
akva2: ok

src/ASM/FiniteElement.C
akva2: ok

akva2:

src/LinAlg/matrix.h
```diff
@@ -387,7 +387,7 @@ namespace utl //! General utility classes and functions.
         for (size_t r = 0; r < ncol; r++)
           for (size_t c = 0; c < nrow; c++)
             this->elem[c+nrow*r] = mat.elem[r+ncol*c];
-      else
+      else if (!mat.elem.empty())
         this->elem.fill(mat.elem.ptr());
```
kmokstad, line 391: and maybe use `data()` here as well, for consistency.

akva2: ok

akva2:

src/Utility/HDF5Writer.C
```diff
@@ -349,6 +349,8 @@ void HDF5Writer::writeSIM (int level, const DataEntry& entry,
   {
     Vector pchVec;
     size_t ncmp = sim->getProblem()->getNoFields(2);
+    if (ncmp == 0)
+      return;
```
kmokstad, line 353: This leaves the temporary `pchVec` unused if `ncmp == 0`, so maybe move its definition after the if-statement

src/Utility/Legendre.C
kmokstad: what links? I don't see them. But anyway I let this pass..

kmokstad (approved):
Although I still not see the need for the Legendre fix, this one looks fine now

src/Utility/Legendre.C
akva2: I removed them for final merge, they were in the original commit messages before I started preparing for merge. http://afem:8080/job/IFEM-static-analysis/302/testReport/junit/(root)/debug-iter/TestLegendre_GL/

kmokstad: OK, so yet another case of dereferencing a pointer to an empty array. This could then also more elegantly been resolved by using `r_val.data()`  and `c_val.data()` in the `dgeev_()` in `DenseMatrix::solveEigNon()`.

akva2: it would not solve the underlying problem which is that nothing is resizing those vectors. This is a logic error.

src/Utility/HDF5Writer.C
akva2: done.

src/Utility/Legendre.C
kmokstad: Really? Check again. The first call to `dgeev_()` (with `Lwork=-1`)  is only for obtaining the length of the work array needed. It will _not_ access the `val` and `vec` arguments (we could as well sent in the `nullptr` here instead). Then they are resized before the second call which actually does the work. Many of the LAPACK-subroutines work like this.

akva2: right. in any case i see no problem with resizing up front.

kmokstad: It is not a problem, just unneccessary

PR #674, opened 2024-03-03
Added: ASMbase::addRigidCouplings()

kmokstad:
For use when reading rigid element definitions from external mesh file.
Also fixed a crash if MLGN is empty.
The support for 2D Lagrangian patches with a mix of 3- and 4-noded elements is also included here.
This is sufficient for the regression tests in OPM/IFEM-ShellEx to pass.

PR #686, opened 2024-04-07
Cache simpification for Lagrange patches

kmokstad:
Sits on top of #674 so only look at the last three commits here.

This was discovered during the work on the Gunnerus boat model, which is a large single-patch 2D Lagrange model.
It will simplify the handling of the basis function cache for Lagrange patches. As far I can see the basis functions stored in the cache will be identical for all elements in a Lagrange patch. Therefore it will be a waste to store a copy for each element and better to let each one use the same instance. Might save some memory footprint for large patches, but probably not a critical amount. The main motivation is to make the logic more clear and less prone for misinterpretation, etc.

Can't rule there are some aspects I overlooked such that separate copies are needed anyway, but at least this works for the Gunnerus model it seems.

### DIFF
--- a/src/ASM/ASMs2DLag.C
+++ b/src/ASM/ASMs2DLag.C
@@ -319,7 +319,7 @@ bool ASMs2DLag::integrate (Integrand& integrand,
   if (myCache.empty())
     myCache.emplace_back(std::make_unique<BasisFunctionCache>(*this, cachePolicy, 1));
 
-  ::BasisFunctionCache<2>& cache = static_cast<::BasisFunctionCache<2>&>(*myCache.front());
+  ASMs2D::BasisFunctionCache& cache = *myCache.front();
   cache.setIntegrand(&integrand);
   if (!cache.init(1))
     return false;
@@ -935,7 +935,7 @@ bool ASMs2DLag::assembleL2matrices (SparseMatrix& A, StdVector& B,
                                     const L2Integrand& integrand,
                                     bool continuous) const
 {
-  BasisFunctionCache& cache = static_cast<BasisFunctionCache&>(*myCache.front());
+  ASMs2D::BasisFunctionCache& cache = *myCache.front();
   if (!cache.init(1))
     return false;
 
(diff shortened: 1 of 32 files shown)


kmokstad:

src/ASM/ASMs1DLag.C
```diff
@@ -139,11 +139,12 @@ bool ASMs1DLag::getElementCoordinates (Matrix& X, int iel, bool) const
     return false;
   }
 
-  const IntVec& ien = MNPC[iel-1];
-  X.resize(nsd,ien.size());
+  // Number of nodes per element
+  size_t nen = std::min((size_t)p1,MNPC[--iel].size());
```
akva2, line 143: c++ cast please

src/ASM/ASMs2DLag.C
```diff
@@ -233,8 +233,7 @@ bool ASMs2DLag::getElementCoordinates (Matrix& X, int iel, bool) const
   }
 
   // Number of nodes per element
-  size_t nen = p1*p2;
-  if (nen > MNPC[--iel].size()) nen = MNPC[iel].size();
+  size_t nen = std::min((size_t)p1*p2,MNPC[--iel].size());
```
akva2, line 236: ditto

src/ASM/ASMs3DLag.C
```diff
@@ -258,8 +258,7 @@ bool ASMs3DLag::getElementCoordinates (Matrix& X, int iel, bool) const
   }
 
   // Number of nodes per element
-  size_t nen = p1*p2*p3;
-  if (nen > MNPC[--iel].size()) nen = MNPC[iel].size();
+  size_t nen = std::min((size_t)p1*p2*p3,MNPC[--iel].size());
```
akva2, line 261: ditto

akva2 (approved):
Seems fine. One minor (x3).

src/ASM/ASMs1DLag.C
kmokstad: Fixed.

src/ASM/ASMs3DLag.C
kmokstad: Fixed.

kmokstad:

PR #687, opened 2024-04-30
Add class for common Piola-mapped weak differential operators

akva2:
Similar to what we have have for equal order and div-compatible.

### DIFF
--- a/Apps/Common/PiolaOperators.h
+++ b/Apps/Common/PiolaOperators.h
@@ -1,0 +1,140 @@
+//==============================================================================
+//!
+//! \file PiolaOperators.h
+//!
+//! \date Apr 30 2024
+//!
+//! \author Arne Morten Kvarving / SINTEF
+//!
+//! \brief Various weak, discrete Piola-mapped operators.
+//!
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this with downstreams ifem-stokes=265 please

akva2:
jenkins build this with downstreams ifem-stokes=265 please

Apps/Common/PiolaOperators.h
```diff
@@ -0,0 +1,140 @@
...
+ *  \details The operators use the block ordering used in the BlockElmMats class.
+ */
+
+class PiolaOperators
+{
+public:
+  //! \brief Common weak operators using Piola mapped discretizations.
+  class Weak
+  {
+  public:
```
kmokstad, line 33: Isn't it better and more clear to use nested namespaces instead of classes here? Since they only contain static methods anyway.

akva2: I need the types for templating purposes.

akva2:
jenkins build this with downstreams ifem-stokes=265 please

PR #690, opened 2024-05-09
Added: Accept near-singular matrices

kmokstad:
A left-over from debugging the unstructured Nastran-file models.
The error condition ierr=ncol+1 is now accepted (but with a warning) instead of aborting the calculations.

Also some minor refactoring in here:
Moved error reporting outside the ifdef to avoid some duplication.
Let the `SuperLUdata` constructor allocate `perm_r` and `perm_c`.


kmokstad:

kmokstad:

PR #692, opened 2024-05-15
Modal simulation driver update

kmokstad:
Adding structural damping parameters, RHS and solution vectors as members in class SIMmodal.
Some of the virtual methods then no longer need to be virtual.

### DIFF
--- a/src/SIM/SIMmodal.C
+++ b/src/SIM/SIMmodal.C
@@ -15,6 +15,9 @@
 #include "AlgEqSystem.h"
 #include "NewmarkMats.h"
 #include "SAM.h"
+#include "Utilities.h"
+#include "IFEM.h"
+#include "tinyxml2.h"
 #include <numeric>
 #include <algorithm>
 #ifdef HAS_CEREAL
@@ -53,6 +56,9 @@ public:
 
 SIMmodal::SIMmodal (std::vector<Mode>& modes) : myModes(modes)
 {
+  parsed = false;
+  alpha1 = alpha2 = 0.0;
+
   modalSys = nullptr;
   modalSam = nullptr;
   myElmMat = nullptr;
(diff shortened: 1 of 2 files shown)


src/SIM/SIMmodal.C
```diff
@@ -78,46 +100,57 @@ bool SIMmodal::swapSystem (AlgEqSystem*& sys, SAM*& sam)
...
     }
 
-  return true;
+  return sol;
+}
+
+
+const Vector& SIMmodal::expandedSolution (int idx) const
+{
+  if (idx >= 0 && idx < (int)sol.size())
```
akva2, line 149: c++ cast

src/SIM/SIMmodal.C
```diff
@@ -67,6 +73,22 @@ SIMmodal::~SIMmodal ()
 }
 
 
+bool SIMmodal::parseParams (const tinyxml2::XMLElement* elem)
+{
+  if (parsed)
+    IFEM::cout <<"\t(skipped)"<< std::endl;
+  else if (!strcasecmp(elem->Value(),"newmarksolver"))
```
akva2, line 80: may as well check for non-null elem or some sca might get nervous.

akva2 (approved):
Two small suggestions.

src/SIM/SIMmodal.C
kmokstad: done

src/SIM/SIMmodal.C
kmokstad: ok, done that too. But there are any other places around where we don't do this check.

kmokstad:
jenkins build this with downstreams ifem-elasticity=164 ifem-shellex=4 please

src/SIM/SIMmodal.C
akva2: yeah i'm aware. sometimes i've had to add checks to quell some warnings so just a defensive knee-jerk.

PR #693, opened 2024-05-15
Fix VTF output of element results for mixed Lagrange meshes

kmokstad:
When a Lagrange mesh consists of a mixture of quads and triangles (and possibly beams as well). The elements are written to the VTF-file block-wise, ordered by increasing number of element nodes. Thus, the element output order is different than the internal ordering of the elements in the patch and a permutation of the result arrays is needed. This will take care of it.

The `SIMoutput::writeGlvNo()` method is extended to also write out the global element numbers as element results - useful for debugging complex models.

Finally (last commit), the option to define explicit element sets as topological items is added for `ASMu2DLag`, similar as already available for `ASMu1DLag`.

### DIFF
--- a/src/ASM/ASMbase.C
+++ b/src/ASM/ASMbase.C
@@ -1356,14 +1356,36 @@ void ASMbase::shiftGlobalElmNums (int eshift)
 }
 
 
-void ASMbase::extractElmRes (const Matrix& globRes, Matrix& elmRes) const
+void ASMbase::extractElmRes (const Vector& globRes, Vector& elmRes,
+                             bool internalOrder) const
+{
+  elmRes.clear();
+  elmRes.reserve(MLGE.size());
+
+  size_t jel = 0;
+  for (int iel : MLGE)
+    if (iel > 0)
+    {
+      size_t idx = internalOrder ? jel++ : iel-1;
+      elmRes.push_back(idx < globRes.size() ? globRes[idx] : 0.0);
+    }
+}
+
+
+void ASMbase::extractElmRes (const Matrix& globRes, Matrix& elmRes,
+                             bool internalOrder) const
 {
   elmRes.resize(globRes.rows(),MLGE.size(),true);
 
-  size_t ivel = 0;
+  size_t ivel = 0, jel = 0;
   for (int iel : MLGE)
     if (iel > 0)
-      elmRes.fillColumn(++ivel,globRes.getColumn(iel));
+    {
+      ++ivel;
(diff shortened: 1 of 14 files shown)


src/ASM/ASMbase.C
```diff
@@ -1356,14 +1356,37 @@ void ASMbase::shiftGlobalElmNums (int eshift)
...
+}
+
+
+void ASMbase::extractElmRes (const Matrix& globRes, Matrix& elmRes,
+                             bool internalOrder) const
 {
   elmRes.resize(globRes.rows(),MLGE.size(),true);
 
-  size_t ivel = 0;
+  size_t ivel = 0, c = 0;
```
akva2, line 1380: maybe use same variable name as above for consistency

src/Utility/ElementBlock.C
```diff
@@ -20,11 +20,12 @@
 
 double ElementBlock::eps = 0.0;
 
+static std::array<size_t,7> legalNENs = { 2, 3, 4, 6, 8, 9, 37 };
```
akva2: 37 vs 27 in original code?

src/ASM/ASMu2DLag.C
```diff
@@ -177,6 +177,44 @@ int ASMu2DLag::parseNodeBox (const std::string& setName, const char* data)
...
+IntVec& ASMu2DLag::getElementSet (const std::string& setName, int& idx)
+{
+  idx = 1;
+  for (ASM::NodeSet& es : nodeSets)
+    if (es.first == setName)
+      return es.second;
+    else if (idx)
+      ++idx;
+
+  elemSets.push_back(std::make_pair(setName,IntVec()));
```
akva2, line 213:
```c++
elemSets.emplace_back(setName,IntVec())
```

src/Utility/ElementBlock.C
kmokstad: bummer, already fixed. Should be 27 of course (9^3)

src/ASM/ASMbase.C
kmokstad: sure, using `jel` here also instead of `c`

src/ASM/ASMu2DLag.C
kmokstad: done (still not got used to using emplace_back() instead of push_back()).

kmokstad:

kmokstad:

kmokstad:

PR #697, opened 2024-05-30
Added: support for an extra basis in the mixed formulations

akva2:
This contains preparation work and then support for the extra basis in the mixed formulations.

### DIFF
--- a/src/ASM/ASMs1D.C
+++ b/src/ASM/ASMs1D.C
@@ -62,10 +62,9 @@ ASMs1D::ASMs1D (const ASMs1D& patch, unsigned char n_f)
 bool ASMs1D::read (std::istream& is)
 {
   if (shareFE) return true;
-  if (curv) delete curv;
 
   Go::ObjectHeader head;
-  curv = new Go::SplineCurve;
+  curv = std::make_shared<Go::SplineCurve>();
   is >> head >> *curv;
 
   // Eat white-space characters to see if there is more data to read
@@ -80,16 +79,14 @@ bool ASMs1D::read (std::istream& is)
   if (!is.good() && !is.eof())
   {
     std::cerr <<" *** ASMs1D::read: Failure reading spline data"<< std::endl;
-    delete curv;
-    curv = nullptr;
+    curv.reset();
     return false;
   }
   else if (curv->dimension() < 1)
   {
     std::cerr <<" *** ASMs1D::read: Invalid spline curve patch, dim="
 	      << curv->dimension() << std::endl;
-    delete curv;
-    curv = nullptr;
+    curv.reset();
     return false;
   }
   else if (curv->dimension() < nsd)
@@ -122,9 +119,10 @@ void ASMs1D::clear (bool retainGeometry)
   if (!retainGeometry)
   {
     // Erase spline data
-    if (proj && proj != curv) delete proj;
-    if (curv && !shareFE) delete curv;
-    geomB = projB = curv = proj = nullptr;
+    curv.reset();
+    proj.reset();
+    geomB.reset();
+    projB.reset();
   }
 
   // Erase the FE data
@@ -254,8 +252,10 @@ bool ASMs1D::createProjectionBasis (bool init)
 {
   if (!curv)
     return false;
-  else if (init && !proj)
-    projB = proj = curv->clone();
+  else if (init && !proj) {
+    proj.reset(curv->clone());
+    projB = proj;
+  }
 
   std::swap(geomB,projB);
   std::swap(curv,proj);
@@ -693,7 +693,7 @@ Tensor ASMs1D::getRotation (size_t inod) const
 
 bool ASMs1D::getElementCoordinates (Matrix& X, int iel, bool) const
 {
-  return this->getElementCoordinates(X,iel,MNPC,curv);
+  return this->getElementCoordinates(X,iel,MNPC,curv.get());
 }
 
 
@@ -830,7 +830,7 @@ Vec3 ASMs1D::getElementCenter (int iel) const
   {
     double u[2];
     this->getElementBorders(iel,u);
-    SplineUtils::point(X0,0.5*(u[0]+u[1]),curv);
+    SplineUtils::point(X0,0.5*(u[0]+u[1]),curv.get());
   }
   return X0;
 }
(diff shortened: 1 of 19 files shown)


akva2:
jenkins build this with downstreams ifem-stokes=270 ifem-navierstokes=220 ifem-boussinesq=117 please

src/ASM/ASMs1D.C
```diff
@@ -1956,15 +1956,14 @@ bool ASMs1D::getNoStructElms (int& n1, int& n2, int& n3) const
 bool ASMs1D::evaluate (const FunctionBase* func, RealArray& values,
                        int, double time) const
 {
-  Go::SplineCurve* scrv = SplineUtils::project(curv,*func,func->dim(),time);
+  std::unique_ptr<Go::SplineCurve> scrv(SplineUtils::project(curv.get(),*func,func->dim(),time));
```
kmokstad, line 1959: Don't see what is gained here by the template usage (except for the more lengthy statement), since this pointer is just a local variable which is deleted before returning.

src/ASM/ASMs1D.h
```diff
@@ -466,8 +466,8 @@ class ASMs1D : public ASMstruct, public ASMunstruct, public ASM1D
   virtual void getElmConnectivities(IntMat& neighs) const;
 
 protected:
-  Go::SplineCurve* curv; //!< Pointer to the actual spline curve object
-  Go::SplineCurve* proj; //!< Pointer to spline curve for projection basis
+  std::shared_ptr<Go::SplineCurve> curv; //!< Pointer to the actual spline curve object
+  std::shared_ptr<Go::SplineCurve> proj; //!< Pointer to spline curve for projection basis
```
kmokstad, line 470: Then remove "Pointer to" in the doxy to keep it within 80 columns.

Applies elsewhere in similar declarations as well.

src/ASM/ASMs2D.C
```diff
@@ -155,16 +154,14 @@ bool ASMs2D::read (std::istream& is)
...
+    surf.reset();
     return false;
   }
   else if (surf->dimension() < 2)
   {
     std::cerr <<" *** ASMs2D::read: Invalid spline surface patch, dim="
-	      << surf->dimension() << std::endl;
-    delete surf;
-    surf = nullptr;
+          << surf->dimension() << std::endl;
```
kmokstad, line 163: Wrong indent

src/ASM/LR/ASMu3Dmx.h
```diff
@@ -36,7 +36,7 @@ class ASMu3Dmx : public ASMu3D, private ASMmxBase
   //! \brief Copy constructor.
   ASMu3Dmx(const ASMu3Dmx& patch, const CharVec& n_f = CharVec(3,0));
   //! \brief Empty destructor.
-  virtual ~ASMu3Dmx() {}
+  virtual ~ASMu3Dmx() = default;
```
kmokstad, line 39: Guess these are equivalent, right? Just to be consistent.

src/ASM/ASMs2Dmx.C
```diff
@@ -1045,23 +1045,27 @@ bool ASMs2Dmx::evalSolutionPiola (Matrix& sField, const Vector& locSol,
...
 
-  size_t len = std::accumulate(nb.begin(),nb.end(),0u);
+  size_t len = std::inner_product(nb.begin(), nb.end(), nfx.begin(), 0u);
   bool withPressure = len == locSol.size();
+  size_t nFields;
   if (!withPressure)
-    len = nb[0] + nb[1];
+    len = nb[0] + nb[1], nFields = 2;
+  else
+    nFields = std::accumulate(nfx.begin(), nfx.end(), 0);
```
kmokstad, line 1074: Maybe better with
```
  size_t nFields = 2;
  if (withPressure)
    nFields = std::accumulate(nfx.begin(), nfx.end(), 0u);
  else
    len = nb[0] + nb[1];
```
The logic is in any case hard to grasp

src/ASM/LR/ASMu2Dmx.C
```diff
@@ -900,37 +900,41 @@ bool ASMu2Dmx::evalSolutionPiola (Matrix& sField, const Vector& locSol,
...
-  size_t len = std::accumulate(nb.begin(),nb.end(),0u);
+  size_t len = std::inner_product(nb.begin(), nb.end(), nfx.begin(), 0u);
   bool withPressure = len == locSol.size();
+  size_t nFields;
   if (!withPressure)
-    len = nb[0] + nb[1];
+    len = nb[0] + nb[1], nFields = 2;
+  else
+    nFields = std::accumulate(nfx.begin(), nfx.end(), 0);
+
```
kmokstad, line 930: See previous comment for ASMs2Dmx

kmokstad (approved):
Although I'm not too fond of "smart" pointers, I guess this is fine. Did not go into the logic though which is hard to follow, and now will be even harder since we don't need to understand where the dynamic objects go out of scope and needs to be deleted (but that' my personal view only).

src/ASM/LR/ASMu3Dmx.h
akva2: yes, they are equivalent.

src/ASM/ASMs2D.C
akva2: fixed. my IDE replaced a tab with spaces but my tabs are 2, not 8..

src/ASM/ASMs1D.h
akva2: ok.

src/ASM/ASMs2Dmx.C
akva2: okay, it will cause sca warnings but those are harmless of course.

akva2:

src/ASM/ASMs1D.C
akva2: i now immediately see that this pointer won't leak, as opposed to having to look through the code to find the matching delete, and making sure there are not early exits or the likes where it would leak. in this case, sure, there's none of those cases but in general i think that new and delete should be removed from the user facing language and be reserved to low level libraries.

akva2:

PR #698, opened 2024-06-04
Extend support for analytic boundary conditions

akva2:
- add support for conditions on multiple bases
- add support for analytic robin conditions

also extend the multipatch model generator to generate sets that is used with RT.

### DIFF
--- a/Apps/Common/MultiPatchModelGenerator.C
+++ b/Apps/Common/MultiPatchModelGenerator.C
@@ -455,6 +455,8 @@ bool MultiPatchModelGenerator2D::createTopologySets (SIMinput& sim) const
   TopEntity& e2f = sim.topology("Edge2Patches");
   TopEntity& e3f = sim.topology("Edge3Patches");
   TopEntity& e4f = sim.topology("Edge4Patches");
+  TopEntity& xb  = sim.topology("BoundaryX");
+  TopEntity& yb  = sim.topology("BoundaryY");
 
   auto&& insertion = [&sim, &e5](TopEntity& e, TopEntity& ef, TopItem top)
   {
@@ -469,10 +471,19 @@ bool MultiPatchModelGenerator2D::createTopologySets (SIMinput& sim) const
     insertion(e1, e1f, TopItem(i*nx+1,1,1));
     insertion(e2, e2f, TopItem((i+1)*nx,2,1));
   }
+  for (const TopItem& e : e1)
+    xb.insert(e);
+  for (const TopItem& e : e2)
+    xb.insert(e);
+
   for (size_t i = 0; i < nx; ++i) {
     insertion(e3, e3f, TopItem(i+1,3,1));
     insertion(e4, e4f, TopItem(nx*(ny-1)+1+i,4,1));
   }
+  for (const TopItem& e : e3)
+    yb.insert(e);
+  for (const TopItem& e : e4)
+    yb.insert(e);
 
   TopEntity& c = sim.topology("Corners");
   auto&& insertionv = [&sim, &c](TopEntity& e, TopItem top)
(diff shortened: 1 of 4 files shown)


akva2:

akva2:

src/SIM/SIMinput.C
```diff
@@ -599,13 +599,20 @@ bool SIMinput::parseBCTag (const tinyxml2::XMLElement* elem)
...
-    if (rval)
-      this->setNeumann(rval->Value(),type,direction,code);
-    else
-      this->setNeumann(std::string(),type,direction,code);
+    if (type == "anasol")
+    {
+      IFEM::cout <<" (analytic)";
+      this->setPropertyType(code,Property::ROBIN_ANASOL);
+    }
+    else {
```
kmokstad, line 607: Inconsistent style, { after the else here, but on new line i L603.

src/SIM/SIMinput.C
```diff
@@ -599,13 +599,20 @@ bool SIMinput::parseBCTag (const tinyxml2::XMLElement* elem)
...
+      if (!type.empty())
+        IFEM::cout <<" ("<< type <<")";
+      this->setPropertyType(code,Property::ROBIN);
+      if (rval)
+        this->setNeumann(rval->Value(),type,direction,code);
+      else
+        this->setNeumann(std::string(),type,direction,code);
```
kmokstad, line 614: or `this->setNeumann(rval ? rval->Value() : "", type, direction,code);`

Apps/Common/MultiPatchModelGenerator.C
```diff
@@ -469,10 +471,19 @@ bool MultiPatchModelGenerator2D::createTopologySets (SIMinput& sim) const
     insertion(e1, e1f, TopItem(i*nx+1,1,1));
     insertion(e2, e2f, TopItem((i+1)*nx,2,1));
   }
+  for (const auto& e : e1)
+    xb.insert(e);
+  for (const auto& e : e2)
+    xb.insert(e);
```
kmokstad, line 477: All the `e` objects here will be a `TopEntity` reference, so I rather would use `for (const TopEntity& e : e1)`, etc. here, and the reader will save a few seconds figuring out that himself ;)

Apps/Common/MultiPatchModelGenerator.C
```diff
@@ -848,6 +859,22 @@ bool MultiPatchModelGenerator3D::createTopologySets (SIMinput& sim) const
       for (size_t i = 0; i < nx; ++i)
         insertion(TopItem(IJK2K(i,j,k),r,2), "Boundary", "Face");
 
+  TopEntity& bx  = sim.topology("BoundaryX");
+  TopEntity& by  = sim.topology("BoundaryY");
+  TopEntity& bz  = sim.topology("BoundaryZ");
+
+  auto side_insertion = [&sim](TopEntity& set, int start)
```
kmokstad, line 866: All the other lambda's in this file are declared `auto&&`, so why not here as well for the consistency? (don't remember if there is a practical difference in this case or not).

kmokstad (approved):
Only a few cosmetic tings

src/SIM/SIMinput.C
akva2: sure. not new code but we can improve while at it.

Apps/Common/MultiPatchModelGenerator.C
akva2: they are Topitem, but sure.

Apps/Common/MultiPatchModelGenerator.C
akva2: changed.

akva2:

PR #700, opened 2024-06-06
Fix VTF output for staggered solvers sharing VTF

akva2:
sharing VTF output between different ASMs, element orders
may differ. we need to order things in the VTF in the correct
output by using the correct gridParameters and MLGE.

for element norm outputs these have to externally been
reordered to the master patch order before being sent
to writeGlvN.

Also move the copyRefinement implementation to free functions to facility reuse in such staggered solvers.
I need to reuse these functions in adaptive staggered equal-order Boussinesq.


akva2:

PR #701, opened 2024-06-06
Element matrix cache fix

kmokstad:
In case the external element numbers are truly external, and not just a generated sequence [1,nel], like when importing Nastran input files, we need to account for that when allocating the buffers.

Also use element matrix buffers when doing linear multi-load-case simulation using the NonLinSIM driver.


kmokstad:

kmokstad:

kmokstad:
jenkins build this with downstreams ifem-elasticity=167 ifem-shellex=6 please

PR #255, opened 2018-03-15
added: class for simple global node number establishment for LR splines

akva2:
use this to enable adaptive refinement for mixed multi-patch models.
in particular, the refinements basis differ from the finite element
bases for some formulations, so a simple node numbering scheme for
the refinement basis is required to perform the multi-patch refinements.

PR #704, opened 2024-06-21
added: support for mixed multipatch adaptive simulations

akva2:
Extracted from https://github.com/OPM/IFEM/pull/255

I left out the dd part since it's not yet working and not yet required.

It's time to get this in.

### DIFF
--- a/src/ASM/LR/GlobalNodes.C
+++ b/src/ASM/LR/GlobalNodes.C
@@ -1,0 +1,186 @@
+// $Id$
+//==============================================================================
+//!
+//! \file GlobalNodes.C
+//!
+//! \date Mar 13 2018
+//!
+//! \author Arne Morten Kvarving / SINTEF
+//!
+//! \brief Simple global node establishment for unstructured FE models.
+//!
+//==============================================================================
+
+#include "GlobalNodes.h"
+
+#include "ASMLRSpline.h"
+#include "Utilities.h"
+
+#include <numeric>
+
+namespace {
+
+/*!
+  \brief Class for ordering interfaces in processing order.
+*/
+
+class InterfaceOrder {
+public:
+  //! \brief Comparison operator for interfaces
+  //! \param A First interface
+  //! \param B Second interface
+  bool operator()(const ASM::Interface& A, const ASM::Interface& B) const
+  {
+    if (A.master != B.master)
+      return A.master < B.master;
+
+    if (A.slave != B.slave)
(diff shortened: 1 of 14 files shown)


akva2:
jenkins build this with downstreams ifem-stokes=280 please

src/ASM/LR/GlobalNodes.C
```diff
@@ -0,0 +1,187 @@
...
+  for (const auto& pch : pchs) {
+    it->resize(pch->nBasisFunctions());
+    std::iota(it->begin(), it->end(), nNodes);
+    nNodes += pch->nBasisFunctions();
+    ++it;
+  }
+
+  // remap common nodes
+  InterfaceOrder ifOrder;
+  std::set<ASM::Interface, InterfaceOrder> ifset(ifOrder);
```
kmokstad, line 158: Is it necessary with the `ifOrder` instance here? Thought the following would work fine:

    std::set<ASM::Interface,InterfaceOrder> ifset;

We have a similar construct for the `MPC` sets.
Even better, could the `InterfaceOrder` class be replaced by a local lambda here?

src/SIM/SIMinput.C
```diff
@@ -1604,9 +1604,16 @@ bool SIMinput::refine (const LR::RefineData& prm, Vectors& sol)
...
-        if ((locId = myModel[i]->getNodeIndex(k+1)) > 0)
-          if (refineIndices[i].insert(locId-1).second)
+        if (!prm.MLGN.empty()) {
+          const auto it = std::find(prm.MLGN[i].begin(), prm.MLGN[i].end(), k);
+          if (it != prm.MLGN[i].end())
+            if (refineIndices[i].insert(std::distance(prm.MLGN[i].begin(), it)).second)
+              changed = true;
+        } else {
+         if ((locId = myModel[i]->getNodeIndex(k+1)) > 0)
+           if (refineIndices[i].insert(locId-1).second)
```
kmokstad, line 1614: check indents (probably some tabs' involved)

src/ASM/LR/GlobalNodes.C
akva2: removed the instance. cannot use a lambda as a tempalte parameter.

src/SIM/SIMinput.C
akva2: fixed

akva2:
jenkins build this with downstreams ifem-stokes=280 please

PR #694, opened 2024-05-21
Small beam extensions

kmokstad:
Basically, adding nodal eccentricity vectors as member in class `FiniteElement`.

Also in this PR (totally unrelated, I know) is the option to specify an element set (for unstructured 2D Lagrange meshes) through one or more bounding boxes (third commit).

PR #706, opened 2024-08-29
Added: virtual method ASMbase::isInElementSet().

kmokstad:
Fixed: Wrong set referred in the non-const getElementSet() methods.

- continuation of the last commit of the now-merged #694


kmokstad:
jenkins build this with downstreams ifem-shellex=10 please

PR #707, opened 2024-09-18
Update interface to SPR equation solver

kmokstad:
In the early(est) days of IFEM, the SPR equation solver (from the SAM library) was the only direct solver available. Later when SuperLU was interfaced its coupling was totally neglected - until now, when the SAM library is available as open source (https://github.com/openfedem/sam-lib). This will therefore reenable the option to use that solver (for symmetric positive-definite systems only), as an alternative to SuperLU.

Also added the option to use the LAPACK eigen solver or SLEPC, for Dense and PETSc matrices, respectively, as well as the Lanczos eigensolver for SPR-matrices, by specifying opt.eig > 10.


kmokstad:

kmokstad:
jenkins build this with downstreams ifem-elasticity=169 please

kmokstad:
jenkins build this with downstreams ifem-elasticity=169 please

PR #710, opened 2024-10-16
 Facilitate splitting the assembly step in two or more parts

kmokstad:
Added a virtual method `getElementCenter()` in `ASMbase` for calculation of geometric center of a specified element. Implemented for all Lagrange patches. In addition the LRSpline patches already had these implemented, but they are now also available via the `ASMbase` interface.

Also added methods to assign a list of currently active elements. This can be used to split the assembly stage into several sub-steps, in case some manipulation is desired in between the assembly of various element groups, for instance.

Also added missing newline in the matrix_market output format for system matrices.

Finally, a  consistent copy constructor and assignment operator are added for `AlgEqSystem`. This includes some fixes of the copy constructors for `SystemMatrix` sub-classes.

### DIFF
--- a/src/ASM/ASMLagBase.C
+++ b/src/ASM/ASMLagBase.C
@@ -34,3 +34,15 @@ bool ASMLagBase::nodalField (Matrix& field, const Vector& sol, size_t nno) const
 
   return true;
 }
+
+
+Vec3 ASMLagBase::getGeometricCenter (const std::vector<int>& MNPC) const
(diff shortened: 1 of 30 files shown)


kmokstad:

akva2 (approved):
Looks fine. For the record I still think the alloca usage is over-complication for absolutely no gain.

kmokstad:

PR #712, opened 2024-11-19
Janitoring: Quell gcc13 warnings

akva2:


akva2:

src/SIM/AdaptiveSetup.C
```diff
@@ -431,7 +431,7 @@ int AdaptiveSetup::calcRefinement (LR::RefineData& prm, int iStep,
 
   if (!errPrefix.empty())
   {
-    char suffix[9];
+    char suffix[17]{};
```
kmokstad, line 434: Ok, but weird that 9 bytes is not enough here. The fixed format string should always produce a 8-byte string here + the 0-terminator. And why 17? If more than 9 actually is needed, I feel better when a string size is 2^(something), i.e. 16 in this case.

akva2: With 9:
```
warning: ‘%03d’ directive writing between 3 and 11 bytes into a region of size 8 [-Wformat-overflow=]
```
with 16:
```
warning: ‘__builtin___sprintf_chk’ may write a terminating nul past the end of the destination [-Wformat-overflow=]
```

17 is the first that shuts it up.

akva2: yes, it's probably called a bug.. anyways, keep 17 or make it 32?

akva2:
jenkins build this please

src/SIM/AdaptiveSetup.C
kmokstad: 17 is fine then, since the value in any case is unlogic. But maybe spare a comment in the code on why this is needed, such that not "someone" tries to be smart later...

akva2:
jenkins build this please

PR #713, opened 2024-11-19
fixed: build without superlu

akva2:
Random discovery..

### DIFF
--- a/src/LinAlg/SparseMatrix.C
+++ b/src/LinAlg/SparseMatrix.C
@@ -1252,6 +1252,7 @@ bool SparseMatrix::solveSLUx (Vector& B, Real* rcond)
   std::cerr <<"SparseMatrix::solve: SuperLU solver not available"<< std::endl;
 #endif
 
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

src/LinAlg/SparseMatrix.C
```diff
@@ -1272,6 +1273,7 @@ bool SparseMatrix::solveSLUx (Vector& B, Real* rcond)
     if (rcond && slu)
       *rcond = slu->rcond;
   }
+#endif
 
   return true;
```
kmokstad, line 1278: Ok, this was due to the refactoring I did in b32d2cd9d159dab424099d6fe54ffbda64de4ad8. Another "error" introduced there is that this method is supposed to return `false` when no superlu, i.e., L178 should be replaced by
```
return ierr == 0;
```

akva2: ok.

kmokstad:
you could put all these (#712 - #715) into one PR instead then, since they all are about gcc13 porting.

akva2:
jenkins build this please

PR #714, opened 2024-11-19
avoid VLAs

akva2:
these are not portable and causes warnings with newer compilers.

In particular VLAs is a C extension that just happens to be available in gcc and clang. However, never versions warns when they are used, so let's not.

### DIFF
--- a/src/ASM/LR/ASMLRSpline.C
+++ b/src/ASM/LR/ASMLRSpline.C
@@ -350,10 +350,10 @@ bool ASMLRSpline::doRefine (const LR::RefineData& prm, LR::LRSpline* lrspline)
 
 Go::BsplineBasis ASMLRSpline::getBezierBasis (int p, double start, double end)
 {
-  double knot[2*p];
-  std::fill(knot,   knot+p,   start);
-  std::fill(knot+p, knot+2*p, end);
-  return Go::BsplineBasis(p,p,knot);
+  std::vector<double> knot(2*p);
+  std::fill(knot.begin(),     knot.begin() + p,   start);
+  std::fill(knot.begin() + p, knot.begin() + 2*p, end);
+  return Go::BsplineBasis(p, p, knot.data());
 }
 
 
--- a/src/ASM/LR/ASMu3D.C
+++ b/src/ASM/LR/ASMu3D.C
@@ -1751,12 +1751,10 @@ bool ASMu3D::evalSolution (Matrix& sField, const IntegrandBase& integrand,
     fe.v   = gpar[1][i];
     fe.w   = gpar[2][i];
 
-    double u[2*p1];
-    double v[2*p2];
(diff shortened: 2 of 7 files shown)


akva2:

akva2:
if this was inner-loopy I might be concerned but this is called extremely seldom and any optimization concerns here is completely premature. we allocate tens of vectors per integration point, once-ish per simulation stuff really makes absolutely no difference.

src/ASM/LR/ASMLRSpline.C
```diff
@@ -350,10 +350,10 @@ bool ASMLRSpline::doRefine (const LR::RefineData& prm, LR::LRSpline* lrspline)
...
 Go::BsplineBasis ASMLRSpline::getBezierBasis (int p, double start, double end)
 {
-  double knot[2*p];
-  std::fill(knot,   knot+p,   start);
-  std::fill(knot+p, knot+2*p, end);
-  return Go::BsplineBasis(p,p,knot);
+  std::vector<double> knot(2*p);
+  std::fill(knot.begin(),     knot.begin() + p,   start);
+  std::fill(knot.begin() + p, knot.begin() + 2*p, end);
+  return Go::BsplineBasis(p, p ,knot.data());
```
kmokstad, line 356: Misplaced comma, should be `p, p, knot...` not `p, p ,knot...` if you want space between the parameters

src/ASM/LR/ASMu3D.C
```diff
@@ -1751,12 +1751,12 @@ bool ASMu3D::evalSolution (Matrix& sField, const IntegrandBase& integrand,
...
 
-    double u[2*p1];
-    double v[2*p2];
-    double w[2*p3];
-    basis1.computeBasisValues(gpar[0][i], u, 1);
-    basis2.computeBasisValues(gpar[1][i], v, 1);
-    basis3.computeBasisValues(gpar[2][i], w, 1);
+    std::vector<double> u(2*p1);
+    std::vector<double> v(2*p2);
+    std::vector<double> w(2*p3);
```
kmokstad, line 1756: should then have the definition on one line `std::vector<double> u(2*p1), v(2*p2), w(2*p3)` like in L2451 below

kmokstad (approved):
With a couple of minor comments.
It is now very tempting to rewrite `SIMbase::solutionNorms()` to accept `std::vector` arguments instead of the pointers, to avoid the `data()` invokes.

src/ASM/LR/ASMLRSpline.C
akva2: fixed

src/ASM/LR/ASMu3D.C
akva2: done

akva2:
jenkins build this please

PR #718, opened 2024-11-26
fixed: check for zero size vectors in send/recv/allReduce

akva2:
Causes errors with debug iter build


akva2:

PR #719, opened 2024-11-26
C++20 fixes

akva2:
While bumping to c++-20 won't happen before Ubuntu 26.04 (ie when 22.04 is out of support), I tried and it's mostly straight forward. One fix and one warning quell which are already valid under c++-17 so we may as well merge them now.

Also avoid some VLAs in HDF5Writer

### DIFF
--- a/src/Utility/HDF5Restart.C
+++ b/src/Utility/HDF5Restart.C
@@ -78,13 +78,13 @@ bool HDF5Restart::writeData (const SerializeData& data)
        int len, const void* data, hid_t type)
   {
 #ifdef HAVE_MPI
-    int lens[ptot], lens2[ptot];
-    std::fill(lens,lens+ptot,len);
-    MPI_Alltoall(lens,1,MPI_INT,lens2,1,MPI_INT,*m_adm.getCommunicator());
-    hsize_t siz   = (hsize_t)std::accumulate(lens2,lens2+ptot,0);
-    hsize_t start = (hsize_t)std::accumulate(lens2,lens2+pid,0);
+    std::vector<int> lens(ptot), lens2(ptot);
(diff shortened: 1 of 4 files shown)


src/Utility/HDF5Restart.C
```diff
@@ -78,13 +78,13 @@ bool HDF5Restart::writeData (const SerializeData& data)
...
-    int lens[ptot], lens2[ptot];
-    std::fill(lens,lens+ptot,len);
-    MPI_Alltoall(lens,1,MPI_INT,lens2,1,MPI_INT,*m_adm.getCommunicator());
-    hsize_t siz   = (hsize_t)std::accumulate(lens2,lens2+ptot,0);
-    hsize_t start = (hsize_t)std::accumulate(lens2,lens2+pid,0);
+    std::vector<int> lens(ptot), lens2(ptot);
+    std::fill(lens.begin(), lens.end(),len);
+    MPI_Alltoall(lens.data(),1,MPI_INT,lens2.data(),1,MPI_INT,*m_adm.getCommunicator());
+    hsize_t siz   = std::accumulate(lens2.begin(), lens2.end(), hsize_t{0});
+    hsize_t start = std::accumulate(lens2.begin(), lens2.begin() + pid, hsize_t{0});
```
kmokstad, line 85: any practical difference with `hsize_t{0}` vs. `hsize_t(0)` ? If not, since we elsewhere we have used the latter convention when type-casting constants I'd prefer this here also, for the consistency.

kmokstad (approved):
yes, good to bee a but ahead of upcoming issues.

src/Utility/HDF5Restart.C
akva2: no difference for primitive types so i'll change it.

PR #723, opened 2024-12-17
Fixed: Extraction of patch-level element results with internal ordering

kmokstad:
Was incorrect for multi-patch models, where we need to add an offset to the element index equal to the total number of elements in the previous patches when accessing the global array.


kmokstad:

PR #727, opened 2025-01-13
Small changes regarding element/node sets for Lagrange patches

kmokstad:
Of significance is that the `nodeSets` and `elemSets` containers are now protected (instead of private), such that they can be accessed by sub-classes, and the introduction of the bool second argument for `addTo[Node|Elem]` methods. The rest is cosmetics

### DIFF
--- a/src/Utility/Vec3.h
+++ b/src/Utility/Vec3.h
@@ -97,6 +97,9 @@ public:
     return *this;
   }
 
+  //! \brief Negation operator.
+  Vec3 operator-() const { return Vec3(-x,-y,-z); }
+
   //! \brief Return the sum of the vector.
   Real sum() const { return x+y+z; }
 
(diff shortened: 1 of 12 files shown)


kmokstad:

src/Utility/Vec3.h
```diff
@@ -97,6 +97,9 @@ class Vec3
     return *this;
   }
 
+  //! \brief Negation operator.
+  Vec3 operator-() const { return Vec3(-x,-y,-z); }
```
akva2, line 101: should this go in Vec3Oper like the other operators?

kmokstad:

src/Utility/Vec3.h
kmokstad: Vec3Oper only contains global operators, i.e., functions taking the Vec3 object(s) they operate on as arguments. Vec3.h contains all the other operators, like assignment, +=, indexing, etc. Think the unary operator fits well here too.

PR #729, opened 2025-01-22
utl::vector refactoring without inheritance of std::vector

kmokstad:
Changing the `std::vector` relationship from "is-a" to "has-a" since `std::vector` is not supposed to be inherited. A bit more changes than initially planned, but this is basically what it takes. The commits are quite fragmented to better see how things are related. The first 9 commits are actually independent of the `utl::vector` refactor in the 10th commit and could be implemented first without changing the vector class. But they are necessary to make things work after the changes in commit 10. The last commit is side effects of the changes made in commit 10, but is kept separately for clarity.

Quite some downstream affections of course, which will be added subsequently.

PR #730, opened 2025-01-25
Use RealArray as type for reaction forces containers

kmokstad:
These are typically smaller-sized containers which does not require the algebra operations as typical NDOF-length vectors. A `utl::vector` copy constructor is added for easy conversion in the rare event we actually do want to do some algebra stuff. Also fixed a harmless memory leak in `HHTSIM` on destruction.

This used to be part of #729 but is taken out as a prelude for it, as it is a separate issue.

### DIFF
--- a/src/ASM/ReactionsOnly.C
+++ b/src/ASM/ReactionsOnly.C
@@ -19,7 +19,7 @@
 
 
 ReactionsOnly::ReactionsOnly (const SAM* sam, const ProcessAdm& adm,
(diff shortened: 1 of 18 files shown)


kmokstad:
jenkins build this with downstreams ifem-poisson=61 ifem-darcy=69 ifem-elasticity=175

kmokstad:
jenkins build this with downstreams ifem-poisson=61 ifem-darcy=69 ifem-elasticity=175 please

PR #733, opened 2025-01-28
Added: virtual method IntegrandBase::suppressOutput()

kmokstad:
This can be used to filter out specific result components for output to VTF.
The suppressOutput() method needs to be overridden by a app-specific subclass, default is not filtering.


kmokstad:

PR #735, opened 2025-01-31
Account for global Lagrange multipliers in Lagrange patches (single-field)

kmokstad:
We do this by ensuring the size of the temporaries fe.N, fe.Xn and dXu match the MNPC array for the elements, which may contain an extra node representing the LM. This node has no coordinates associated with it (set to zero).

PR #737, opened 2025-02-04
Fixed: Use nen=p1*p2(*p3) as loop range in case MNPC also contains GLMs.

kmokstad:
Changed: Check for element index out-of-range before using as array index.

This is a fixup for #735 which failed when exporting to VTF for models with Gls and Lagrange interpolation.


kmokstad:

kmokstad:

PR #738, opened 2025-02-06
Some utl::vector extensions

kmokstad:
This used to be part of #729 but is taken out in a separate (and smaller) PR as a prerequisite for that one.

We here extend the `utl::vector` class with some overloaded `push_back()` methods, and add more type-casting operators for the `utl::matrix` class. Also extending the `utl::matrix::sum()` method to optionally return the sum of a specified column when the argument is a negative value (column index).

### DIFF
--- a/Apps/Common/PiolaOperators.C
+++ b/Apps/Common/PiolaOperators.C
@@ -204,7 +204,7 @@ void PiolaOperators::Copy (Matrices& EM,
 {
   const size_t nsd = fe.dNdX.cols();
   if (nsd < 1 || nsd > 3)
-      return;
+    return;
   size_t ofs = 1;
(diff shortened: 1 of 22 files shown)


kmokstad:
jenkins build this with downstreams ifem-elasticity=176 ifem-finitedeformation=45 ifem-poroelasticity=64 ifem-openfrac=83 ifem-stokes=310 ifem-navierstokes=257 ifem-boussinesq=147 simra-postproc=17 please

kmokstad:
jenkins build this with downstreams ifem-elasticity=176 ifem-finitedeformation=45 ifem-poroelasticity=64 ifem-openfrac=83 ifem-stokes=310 ifem-navierstokes=257 ifem-boussinesq=147 simra-postproc=17 please

PR #739, opened 2025-02-10
fixed: use getNoProjectionNodes() instead of getNodes()

akva2:
the latter includes GLMs. causes missing output in VTF files


akva2:

kmokstad (approved):

PR #740, opened 2025-02-11
State dependency for property functions

kmokstad:
With this one can assign state variables (displacement and velocity for now) to any of property functions of the integrand, such that they can depend on the current solution in addition to spatial location and time.
So far for 2D Lagrange patches only.


kmokstad:

kmokstad:
jenkins build this with downstreams ifem-shellex=19 please

PR #741, opened 2025-02-14
Fixed: BLAS version of utl::vector::add with stride larger than 1

kmokstad:
The `n` argument to the `[sd]axpy` calls was too large resulting in invalid memory access.
Also adding some sanity checking on the stride arguments, they can be negative (at least for BLAS version), but not zero.
Also adding an overloaded `utl::matrix::fill()` method.

### DIFF
--- a/src/LinAlg/Test/MatrixTests.h
+++ b/src/LinAlg/Test/MatrixTests.h
@@ -10,46 +10,54 @@
 //!
 //==============================================================================
 
-#include "MatVec.h"
+#include "matrixnd.h"
 
-#include "gtest/gtest.h"
+#include <gtest/gtest.h>
 
 #include <limits>
 #include <numeric>
+
 
 namespace {
 
 template<class Scalar>
 void vectorAddTest()
 {
-  constexpr size_t size = 10;
-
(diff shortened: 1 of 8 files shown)


kmokstad:

src/LinAlg/Test/MatrixTests.h
```diff
@@ -10,54 +10,62 @@
...
+    EXPECT_FLOAT_EQ(d2[i], d[i]+(i-4)/3);
 }
 
 
 template<class Scalar>
 void vectorDotTest()
 {
   constexpr size_t size = 10;
-  constexpr Scalar max = Scalar(size-1) * Scalar(size) / 2.0;
+  constexpr Scalar max = Scalar((size-1)*size/2);
```
akva2, line 68: if you cast after the division you're doing the division as integer.

kmokstad: yes, and that is totally fine here, since `(size-1)*size` always will be an even number no-matter the value on `size` such that dividing by two will/should always result in the same as when using float arithmetics

kmokstad:

PR #745, opened 2025-02-28
Small change regarding use of nViz

kmokstad:
This option is for Splines patches only, and don't change it for Lagrange


kmokstad:
jenkins build this nodoxy with downstreams please

kmokstad:
jenkins build this with downstreams ifem-elasticity=179 please

PR #746, opened 2025-03-10
Changed: Skip basis function derivatives when integrand says NO_DERIVATIVES

kmokstad:
E.g, for the Andes shell elements, which have their internal basis function evaluations, it is wasted to evaluate derivatives, unless we need them for load calculation. This was already implemented for 1D Lagrange, but 2D and 3D was missing. So for the consistency..

Also in this PR, some refactoring of the Gauss point parameter calculations, using a one-dimensional array instead of a matrix.


kmokstad:
jenkins build this with downstreams nodoxy please

kmokstad:

kmokstad:

PR #747, opened 2025-03-28
Let the immersedGeometry() method define the name of the added geometry

kmokstad:
Such that this method can be used for other kinds of additional geometries also (point masses, etc.)


kmokstad:

PR #749, opened 2025-04-02
More flexible specification of modes for EigenModeSIM

kmokstad:
Added the option:

    <eigenmodes>
      <modes scale="1.23">1 2 3 5 7 8</modes>
    </eigenmodes>

which is more compact than specifying mode by mode using `<mode number="1">1.23</mode>`.
The latter syntax is still valid though.


kmokstad:
jenkins build this with downstreams nodoxy please

PR #753, opened 2025-05-15
VTF output of external element numbers

kmokstad:
The first commit is unrelated - an extension of the `Parametric` function class defining element activation histories.
The second commit assigns external element numbers to `ElementBlock` objects, for unstructured Lagrange patches (this was erroneously left out initially).

Then, in the third commit the `SIMoutput::writeGlvNo()` method is extended to also output the original external element numbers as a scalar field to `VTF`. This is needed when the external numbers do not consist of a monotonic sequence 1...nel, but contains "holes". The elements are then renumbered such that they can be used as indices in global element arrays. Therefore we need to do this in order to visualize the original numbers.

The last two commits are also unrelated: Signature change for virtual method `SIMbase::printProblem()` with downstreams implications, and some new node set methods in class `ASMs1D`.

### DIFF
--- a/src/ASM/ASMs1D.C
+++ b/src/ASM/ASMs1D.C
@@ -468,6 +468,40 @@ const IntVec& ASMs1D::getNodeSet (int iset) const
 }
 
 
+bool ASMs1D::isInNodeSet (int iset, int inod) const
+{
+  if (iset < 1 || iset > static_cast<int>(nodeSets.size()))
+    return false;
+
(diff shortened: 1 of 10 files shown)


kmokstad:

kmokstad:
jenkins build this with downstreams ifem-elasticity=182 ifem-finitedeformation=48 ifem-poroelasticity=65 ifem-openfrac=87 please

PR #755, opened 2025-05-22
Option to override conflickting Dirchlet conditions

kmokstad:
If you specify two different Dirichlet conditions for the same node, the latter last one will be silently ignored, always.
With the attribute `override="true"` on the `<dirichlet>` tag it is now possible to make the last one override the former.
This in the second commit. The remaining are unrelated smaller things.


kmokstad:

kmokstad:
jenkins build this with downstreams ifem-elasticity=183 please

kmokstad:
jenkins build this with downstreams ifem-elasticity=183 ifem-shellex=24 please

PR #757, opened 2025-06-06
Added: Optionally build SPRMatrix with 64-bit integer version of SPR

kmokstad:
Configure with cmake option `-DIFEM_USE_SPR=I64` to try this.
With `-DIFEM_USE_SPR=<anything-else>` you get the default 32-bit integer version.
With `-DIFEM_USE_SPR=OFF`  you configure without SPR support (same as not specifying IFEM_USE_SPR at all).


kmokstad:

PR #758, opened 2025-06-10
Fix PETSc eigenvalue support

akva2:
Fix using PETSc matrices with ARPACK, and fix SLEPc support


akva2:
Replaced by PR series

PR #759, opened 2025-06-18
Base class for integrands with gravity ++

kmokstad:
The new class `HasGravityBase` derived from `IntegrandBase` is introduced, with methods for assigning and parsing the gravity vector from file. Such that this does not need to replicated for all integrands having gravity as a property.

Also (first commit), the virtual method `SIMbase::addConstraints()` is given a default implementation dealing with explicit nodal constraints as well as entire patch constraints. Since this is a common feature for all sim's it does not need to be replicated in the sub-classes but instead invoking the `SIMbase` method from the overridden methods.

Then (third commit), the method `SIMsupel::recoverInternalDOFs()` is added, containing the general parts of the things earlier found in the class `SIMLinElSup` of [OPM/IFEM-Elasticity](https://github.com/OPM/IFEM-Elasticity).

The other three commits are smaller fixes related to superelement analysis.

### DIFF
--- a/src/ASM/ASMsupel.C
+++ b/src/ASM/ASMsupel.C
@@ -12,6 +12,7 @@
 //==============================================================================
 
 #include "ASMsupel.h"
+#include "Integrand.h"
 #include "GlobalIntegral.h"
 #include "ElementBlock.h"
 #include "Utilities.h"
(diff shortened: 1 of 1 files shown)


kmokstad:
jenkins build this with downstreams ifem-elasticity=185 ifem-stokes=323 ifem-thm=26 please

kmokstad:
jenkins build this with downstreams ifem-elasticity=185 ifem-stokes=323 ifem-thm=26 please

PR #761, opened 2025-06-25
EigSolver: implement support for using PETSc in ARPACK solver

akva2:
Extracted from https://github.com/OPM/IFEM/pull/758


akva2:

src/Eig/EigSolver.C
```diff
@@ -190,6 +190,20 @@ static void _eig_ax(const SystemMatrix* A, const double* x, double* y)
...
+#if HAS_PETSC
+  const PETScMatrix* pA = dynamic_cast<const PETScMatrix*>(A);
+  if (pA) {
+    PETScVector X(pA->getAdm(), x, A->dim());
+    X.endAssembly();
+    PETScVector Y(pA->getAdm(), A->dim());
+    pA->multiply(X, Y);
+    PetscScalar* yv;
+    VecGetArray(Y.getVector(), &yv);
+    memcpy(y,yv,Y.dim()*sizeof(double));
```
kmokstad, line 202: Looks like a memory leak to me. Where is `yv` released?

src/Eig/EigSolver.C
```diff
@@ -211,6 +225,19 @@ extern "C" void eig_sol_(const double* x, double* y, int& ierr)
...
 
+#ifdef HAS_PETSC
+  const PETScMatrix* pA = dynamic_cast<const PETScMatrix*>(AM);
+  if (pA) {
+    PETScVector RHS(pA->getAdm(), x, AM->dim());
+    RHS.endAssembly();
+    ierr = AM->solve(RHS) ? 0 : -1;
+    PetscScalar* yv;
+    VecGetArray(RHS.getVector(), &yv);
+    memcpy(y,yv,RHS.dim()*sizeof(double));
```
kmokstad, line 236: Same here. Need to release `yv`? Moreover, the memcpy will work only work if PetsCScalar and double are the same type (or at least same number of bytes per word). Is that always true? Otherwise, maybe std::copy would be more safe.

kmokstad (commented):
Not sure of this was ready for merge or still some WIP.
Just spotted a few suspicious things..

src/Eig/EigSolver.C
akva2: yv is owned by Y, we borrow a reference to its internal buffer.

src/Eig/EigSolver.C
akva2: Same reason. And yes petsc scalars are always double / we only support petsc with double scalars.

kmokstad: Got it, PetscScalar=double, always

akva2:

akva2:

PR #763, opened 2025-06-27
fixed: profiler does MPI calls

akva2:
need to cycle LinAlgInit ref count


akva2:

PR #764, opened 2025-07-02
Changed: Use SIM::RHS_ONLY also in NewmarkSIM

kmokstad:
To avoid that the system Newton matrix is assembled unnecessarily when iterating the RHS vector only.
This will require some downstream updates though (for users of the NewmarkSIM driver).


kmokstad:
jenkins build this with downstreams ifem=elasticity186 ifem-shellex=27 please

kmokstad:
jenkins build this with downstreams ifem=elasticity186 ifem-shellex=27 please

PR #766, opened 2025-07-10
ThreadGroups: add oneStripe(list)

akva2:
this adds the specified elements in one stripe,
so they can all be safely assembled in parallel
in structured asms.

to be used for threadsafe integrand handling


akva2:

PR #770, opened 2025-07-14
Profiler: output to IFEM::cout

akva2:
Output to the per-process output instead of std::cout.

In particular this fixes unstable regression tests in parallel, where the profiler output from all processes has the potential to end up interleaved with output we are actually checking for (e.g. recently introduced eigenvalue regression tests).

### DIFF
--- a/src/Utility/Profiler.h
+++ b/src/Utility/Profiler.h
@@ -57,7 +57,7 @@ public:
   void stop(const std::string& funcName);
 
   //! \brief Prints a profiling report for all tasks that have been measured.
-  void report(std::ostream& os) const;
+  template<class Stream> void report(Stream& os) const;
   //! \brief Clears the profiler.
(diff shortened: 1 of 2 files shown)


akva2:

src/Utility/Profiler.h
```diff
@@ -57,7 +57,8 @@ class Profiler
   void stop(const std::string& funcName);
 
   //! \brief Prints a profiling report for all tasks that have been measured.
-  void report(std::ostream& os) const;
+  template<class Stream>
+  void report(Stream& os) const;
```
kmokstad, line 61: Prefer to have the template declaration on the same line, when everything fits within the '80 column standard"

kmokstad (approved):
Yep, this was probably the reason for the failure I mentioned in #764

PR #769, opened 2025-07-11
L2 projection: Add interface for assembling

akva2:
So that we can override this for other matrix types (e.g. PETSc) in the future.

PR #774, opened 2025-07-15
preAssemble: remove unnecessary parameter and virtualize on SystemMatrix level

akva2:
This is a little more to-the-point than (that part of) https://github.com/OPM/IFEM/pull/769 so do this one first.


akva2:

kmokstad (changes requested):
This change will break L2-projection for Mortar-based contact problems.
Suggest to replace it by #780 instead.

akva2:
Replaced by #780

PR #775, opened 2025-07-15
graphPartition: take element graph as parameter

akva2:
instead of a SIM. for upcoming alternative use purposes


kmokstad (approved):
Looks fine (a bit outside my domain), but one suggestion

akva2:

PR #768, opened 2025-07-11
mnpc generation: refactor to facilitate reuse

akva2:
This sticks the MNPC generation in a separate method for reuse purposes.

PR #779, opened 2025-07-28
Refactor MNPC generation for reuse

kmokstad:
This should replace #768.
The main difference is that the class-dependent implementation is now in the static method `createMNPC()` for each class, which then is used by the the virtual `getElmNodes()` methods. The static methods are also used internally instead of going via the virtual method. In addition, the Lagrange implementation of `getElmNodes()` was probably buggy, in that the `nx`, `ny` and `nz` members where used also when calculating for the projection basis. Now they are recalculated using the correct spline object before invoking the static method for generating  the MNPC.


kmokstad:

PR #780, opened 2025-07-29
Added: virtual method SystemMatrix::preAssemble()

kmokstad:
This sits on top of #779 and replaces #774 which will fail if doing contact analysis and L2-projection (see my comment in #769 for further elaboration). Unfortunately, there is currently no regression test catching this flaw.


kmokstad:

akva2 (approved):
Tested, works fine.

PR #784, opened 2025-08-20
fixed: multithreaded face assembly for structured splines

akva2:
we need to use the highest order to determine strip sizes, like we do for volume integrals


kmokstad (approved):
Looks fine.

PR #785, opened 2025-08-21
Fix VTF output for Lagrange models with extra-ordinary nodes

### LINKED ISSUE #783: VTF export for Lagrange when there are extra-ordinary nodes is broken
Like, for LinearElasticity with rigid couplings, for FiniteDeformation with contact.

Seems the VTF export does not account for the added nodes such that a mismatch between the expected number of nodes and the actual size of the results array occurs. Spent some time locating the source of this bug, and for FiniteDeformation at least it originates from the fix #461. Not sure how to best fix this, and not of high priority either. Therefore logging this issue instead, for now.

kmokstad:
This fixes #783. We now always consider basis 1 only when writing BC codes to VTF-file, such that the extraordinary nodes (either from rigid elements or contact analysis) are excluded. If not, the export would fail for Lagrange meshes where a one-to-one correspondence between the FE nodes and the visualization nodes are assumed. For spline meshes it worked since the fields are interpolated anyway.

Also deactivating multi-threading for Lagrange and mixed models when requested. It was only done for single-basis spline models so far.

### DIFF
--- a/src/ASM/ASMs1DLag.C
+++ b/src/ASM/ASMs1DLag.C
@@ -478,7 +478,7 @@ bool ASMs1DLag::evalSolution (Matrix& sField, const Vector& locSol,
                               const RealArray*, bool, int, int) const
 {
   // Direct nodal evaluation
-  return this->nodalField(sField,locSol,coord.size());
+  return this->nodalField(sField,locSol,nnod);
 }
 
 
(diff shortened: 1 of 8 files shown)


kmokstad:
jenkins build this with downstreams ifem-elasticity=187 ifem-finitedeformation=50 please

akva2 (approved):

PR #794, opened 2025-09-05
fixed: parallel L2 assembly for ASMsxDLag

akva2:
as BsplineBasis::knotInterval() is not thread safe. this is called when evaluating the integrand solution inside the omp loop.

### DIFF
--- a/src/ASM/ASMs3DLag.C
+++ b/src/ASM/ASMs3DLag.C
@@ -947,11 +947,13 @@ int ASMs3DLag::findElement (double u, double v, double w,
     return -1;
   }
 
-  const std::array<std::pair<double,int>,3> knot {{
-    {u, svol->basis(0).knotInterval(u)},
-    {v, svol->basis(1).knotInterval(v)},
-    {w, svol->basis(2).knotInterval(w)}
-  }};
+  std::array<std::pair<double,int>,3> knot;
+#pragma omp critical
+  {
+    knot[0] = {u, svol->basis(0).knotInterval(u)};
+    knot[1] = {v, svol->basis(1).knotInterval(v)};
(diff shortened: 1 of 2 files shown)


akva2:

src/ASM/ASMs3DLag.C
```diff
@@ -947,11 +947,14 @@ int ASMs3DLag::findElement (double u, double v, double w,
...
-    {w, svol->basis(2).knotInterval(w)}
-  }};
+  std::array<std::pair<double,int>,3> knot;
+
+#pragma omp critical
+  {
+    knot[0] = {u, svol->basis(0).knotInterval(u)};
+    knot[1] = {v, svol->basis(1).knotInterval(v)};
+    knot[2] = {w, svol->basis(2).knotInterval(w)};
+  }
```
kmokstad, line 957: Will not not work just to insert the `#pragma` before L950? Since it is just a single statement anyway, or..?

src/ASM/ASMs3DLag.C
```diff
@@ -947,11 +947,14 @@ int ASMs3DLag::findElement (double u, double v, double w,
     return -1;
   }
 
-  const std::array<std::pair<double,int>,3> knot {{
-    {u, svol->basis(0).knotInterval(u)},
-    {v, svol->basis(1).knotInterval(v)},
-    {w, svol->basis(2).knotInterval(w)}
-  }};
+  std::array<std::pair<double,int>,3> knot;
+
```
kmokstad, line 951: So just remove the blank line 951 and I'm happy, more clear these statements belong together like in the 2D case above.

akva2: ok

akva2:

PR #795, opened 2025-09-05
SparseMatrix: re-implement and re-enable thread-parallel MxV

akva2:
This time using openmp array reduction


kmokstad (approved):
Great :+1:

PR #796, opened 2025-09-05
MultiPatchModelGenerator: remove subdivision support

akva2:
This was prototype functionality to enable full DD single-patch parallelization through Bezier extraction. Largely unused, and since graph partitioning can be used instead, it's not worth carrying any longer


akva2:
jenkins build this with downstreams ifem-poisson=73 please

kmokstad (approved):
Don't know much about it, so go ahead.

PR #797, opened 2025-09-07
Added: virtual method SystemMatrix::compressPattern()

kmokstad:
Implemented for SparseMatrix only.
Use this in eig::solver() to speed up the matrix-vector multiplication in case of single-thread.


akva2 (approved):
Looks fine

PR #798, opened 2025-09-10
Additional app-dependent displacement blocks

kmokstad:
The protected member `addDisBlk` is added to class `SIMoutput`, through which downstream apps may specify additional displacement field blocks - not belonging to the FE model itself - to be written to the VTF file. handy for visualization of sea surface elevation, etc.

Extended the virtual `writeGlvA()` method with the time argument.

Also (second commit) adding profiling of the `SystemMatrix::initAssembly()` call.


kmokstad:
jenkins build this with downstreams ifem-elasticity=189 ifem-finitedeformation=51 please

PR #799, opened 2025-09-10
Added: Option to indent the profiling report based on level

kmokstad:
To easier spot the level one tasks (no indent) which together make upt the different between the "Total" and "Other".


kmokstad:
jenkins build this with downstreams ifem-elasticity=190 please

PR #802, opened 2025-09-13
Added: Pass global 0-based element index to the integrand

kmokstad:
Used for allocating global element matrix cache for integrands.

Also adding some (missing?) isElementInPartition() calls in integrate() methods.


kmokstad:

akva2 (approved):
Strictly speaking there are downstreams that use the wrong index now, but obviously it's fine for those cases so not necessary to address them now.

PR #803, opened 2025-09-16
changed: don't call preallocate for all system matrices

akva2:
right now the matrices have the same sparsity pattern, so we can just preallocate for the first matrix and use a copy for the rest


akva2:

akva2:

src/ASM/AlgEqSystem.C
```diff
@@ -66,7 +66,10 @@ bool AlgEqSystem::init (LinAlg::MatrixType mtype, const LinSolParams* spar,
     }
 
     PROFILE("Pre-assembly");
-    A[i]._A->initAssembly(sam,preAssemblyFlag);
+    if (i == 0)
+      A[i]._A->initAssembly(sam,preAssemblyFlag);
+    else
+      A[i]._A = A[0]._A->copy();
```
kmokstad, line 72: So this seems to work, but the _A matrix for i > 0 has then already been created unnecessarily, so maybe L59 should be

    if (i == 0 && !A[0]._A)

or put the create step before the loop to make it even more clear

akva2:

kmokstad (approved):
The test on A.empty() is probably not necessary (or just consider it an error condition to bee fool-proof), but that's not important.

akva2:
Yeah maybe. It used to implicitly do the same test in the loop, so added it just to be sure..

PR #807, opened 2025-09-17
export MPIEXEC_EXECUTABLE in IFEMConfig

akva2:
to avoid having to look for it in downstream apps


akva2:

PR #808, opened 2025-09-22
fixed: copy-paste error in TestVector.Norm / TestVectorFallback.Norm

akva2:
Add some sauce to the copy-pasta.

### DIFF
--- a/src/LinAlg/Test/MatrixTests.h
+++ b/src/LinAlg/Test/MatrixTests.h
@@ -93,12 +93,12 @@ template<class Scalar>
 void vectorNormTest()
 {
   constexpr size_t size = 10;
-  constexpr Scalar max = Scalar(size-1) * Scalar(size) / 2.0;
+  const Scalar max = sqrt(size*(size+1)*(2*size+1) / 6);
 
   utl::vector<Scalar> d(size);
-  std::iota(d.begin(), d.end(), 0.0);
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

src/LinAlg/Test/MatrixTests.h
```diff
@@ -93,12 +93,12 @@ template<class Scalar>
 void vectorNormTest()
 {
   constexpr size_t size = 10;
-  constexpr Scalar max = Scalar(size-1) * Scalar(size) / 2.0;
+  const Scalar max = sqrt(size*(size+1)*(2*size+1) / 6);
```
kmokstad, line 96: Don't see why the previous was wrong and the current is right,
but what's the magic with `constexpr size_t size = 120;` why not just `const size_t` like i L96

kmokstad (approved):
Guess it's fine, only need some clarifying on the commented.

src/LinAlg/Test/MatrixTests.h
akva2: constexpr means there is no .rodata entry, ie, the value is inlined as a value in user code. sadly sqrt is not constexpr in clang on jenkins, so there I had to use a const to workaround.

I simply screwed up earlier, we're summing squares of numbers, the previous formula for for \sum n, not \sum n^2, and since the code was never actually executed..

PR #810, opened 2025-09-22
convert unit-tests to catch 

akva2:
catch has much better support for data-parametrized and templated tests
and integrates better with ctest


akva2:
jenkins build this please

Apps/Common/Test/LR/TestMultiPatchLRRefine.C
```diff
@@ -19,10 +19,12 @@
 #include "SIM3D.h"
 #include "tinyxml2.h"
 
-#include "gtest/gtest.h"
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/generators/catch_generators.hpp>
 
 #include <sstream>
 
+namespace {
```
kmokstad, line 27: Don't quite see the point of putting these classes in anonymous namespace in the Test C-file. Is it because they are also defined in other Test-files (but with different content) compiled into the same binary? Or some other reason.

kmokstad (approved):
Since this only affects the units tests themselves, and the changes are that big. there is no way I can judge if something is not ok. Therefore approved.

akva2:
jenkins build this please

PR #820, opened 2025-10-03
added: (advanced) support for LTO in buildsystem

akva2:
while cmake has a generic CMAKE_INTERPROCEDURAL_OPTIMIZATION flag this is problematic since it uses all cores for each linker job, overwhelming the system when there are many linking stages.

so we add our own, mostly "borrowed" from opm-common and simplified, with support for setting the number of jobs and for using LTO cache where available.

- -DIFEM_USE_LTO to enable
- -DIFEM_LTO_JOBS=x to set number of linker optimization job
- -DIFEM_LTO_CONFIGURATION_TYPES=xx to enable for a configuration (defaults to Release and RelWithDebubInfo)
- -DIFEM_LTO_CACHE_POLICY to override default lto cache policies


akva2:
jenkins build this please

akva2:
Extensively tested on my box and non-intrusive. I'll just merge.

PR #821, opened 2025-10-06
Optionally save velocity and accelerations (default off) to VTF

kmokstad:
Now you need to specify `<saveVelAcc/>`  inside the` <postprocessing>` scope to save the velocity and acceleration fields to VTF from NewmarkSIM and sub-classes. To reduce file size when only interested in the deflection.
Also omit surface traction vectors (in Gauss points) unless `<saveTrac\>`is specified.
Then some build system fixups.


kmokstad:
jenkins build this with downstreams ifem-elasticity=202 ifem-beamex=47 please

PR #823, opened 2025-10-06
changed: remove buildsystem support for static analyzers

akva2:
now that clang-tidy supports the functionality of clang-check, this is better handled by external tools


akva2:
jenkins build this with downstreams ifem-costa=33 please

PR #824, opened 2025-10-06
Changed: When the fe.G array is used to carry the surface tangents...

kmokstad:
...let the temporary Jac be a reference to this array such that it is not necessary to copy it afterwards, or use fe.G directly instead when no other use of this array.

Added: Update fe.G to reflect the deformed state when available.

The latter is needed when doing linear dynamics with surface normal pressure and the surface rotates.


kmokstad:
jenkins build this with downstreams ifem-shellex=33 please

PR #836, opened 2025-11-21
Refactor: Reuse common ASMxD test code for ASMuxD and ASMsxD

akva2:
You know what they say, duped twice you ignore, duped thrice you can't ignore any more.


kmokstad (changes requested):
Needs to be rebased after my force-push. Apart from that this is good for me.

PR #845, opened 2025-12-10
Fixed: Write displacement blocks for the additional geometries only...

kmokstad:
..when writing deformations (and not velocities and accelerations).
Swap variables idBlock and jBlk in `SIMoutput::writeGlvS()` for convenience.


kmokstad:
jenkins build this with downstreams ifem-shellex=44 please

PR #847, opened 2026-03-13
Handle empty matrix/vectors on multiply/add

kmokstad:
The title applies to the second commit only. This will ensure the empty objects are automatically initialized when adding new values to them, instead of crashing.

The other commit is totally unrelated, only some BDF clean-up which was lying on my local branch.


kmokstad:

PR #848, opened 2026-03-17
Fix initElement call for mixed Lagrange

kmokstad:
Discovered this when running the DarcyCorrection solver with a Lagrange patch (easier to compare Penalty and AL methods on similar discretizations). Its initElement method needs the element index which is stored in the fe object.


kmokstad:
jenkins build this with downstream please

kmokstad:
jenkins build this with downstream please

PR #849, opened 2026-03-23
fixed: hdf5 norm output with filter

akva2:
there is still a row for the element norms, even if they are not valid. thus the l variable is not necessary, we can use k directly


akva2:

kmokstad (approved):
Assume this is is fine. I've not tested norm output to hdf5 in  a long time (if at all).

PR #852, opened 2026-04-11
Set evaluation time for secondary solutions

kmokstad:
When some secondary solution fields have direct dependency on the current time, this can be is used to pass the time on to the integrand level, since the solution evaluation framework does not have the time as argument.


kmokstad:
jenkins build this with downstreams ifem-elasticity=207 ifem-finitedeformation=53 please

PR #859, opened 2026-05-04
Missing test and minor tolerance adjustments for arm64

akva2:


akva2:

PR #863, opened 2026-05-07
adjustments for removed IFEM-FiniteDeformation

akva2:
The finite deformation solver is going public. This adjusts the Apps/ build system and jenkins build scripts for this.


akva2:
jenkins build this with downstreams ifem-elasticity=211 ifem-fsi=94 please

PR #864, opened 2026-05-07
Fixed: AlgEqSystem::d was nullptr when using copy() in build with OPENMP

kmokstad:
...but run with single thread


kmokstad:

PR #869, opened 2026-05-12
Refresh Doxyfile.in

akva2:
Update to remove outdated tags. Enable QUIET to reduce terminal spam. Fix an outdated setting value.


akva2:
jenkins build this please

PR #870, opened 2026-05-12
AdaptiveSetup: add tests for error limits code

akva2:
Reorganize the code for calculating error limits in AdaptiveSetup, and add a test.

The dörfel regression test in IFEM-Poisson is unstable across machines. Replace with a unit test,
then we can drop the unstable regression test without coverage loss.

### DIFF
--- a/src/SIM/AdaptiveSetup.h
+++ b/src/SIM/AdaptiveSetup.h
@@ -16,6 +16,8 @@
 
 #include "MatVec.h"
 
+#include <utility>
+
 class ScalarFunc;
 class SIMoutput;
 namespace tinyxml2 { class XMLElement; }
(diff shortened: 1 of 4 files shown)


akva2:
jenkins build this please

src/SIM/AdaptiveSetup.h
```diff
@@ -95,6 +97,32 @@ class AdaptiveSetup
...
+  //! internally defined greater-than operator can be used when sorting the
+  //! error+index pairs in decreasing error order.
+  using DblIdx = std::pair<double,int>;
+
+  //! \brief Struct holding info about errors.
+  struct ErrorInfo
+  {
+      double limit = 0.0; //!< Limit for errors
+      double sumErr = 0.0; //!< Sum of errors, beta-scaled target for DORFEL
+      double curErr = 0.0; //!< Current error
```
kmokstad: Indent doesn't look right here.

src/SIM/Test/TestAdaptiveSetup.C
```diff
@@ -0,0 +1,133 @@
...
+#include <numeric>
+#include <string>
+
+namespace {
+
+class SIMA : public SIMdummy<SIMgeneric>
+{
+public:
+    SIMA() {}
+};
```
kmokstad, line 29: The indentation in this file looks awkward.

kmokstad (approved):
I guess the Copilot does not care much about indentation, but I do ;)

src/SIM/AdaptiveSetup.h
akva2: fixed

src/SIM/Test/TestAdaptiveSetup.C
akva2: fixed

akva2:
jenkins build this please

PR #871, opened 2026-05-12
Fixed: Missing call to IntegrandBase::initResultPoints()

kmokstad:
To set evaluation time for secondary solution when doing Greville projection.
Without this, the output typically fails for models with time-dependent element activation.


kmokstad:
jenkins build this with downstreams ifem-elasticity=217 please

kmokstad:

PR #874, opened 2026-05-19
Added: Account for linear couplings when flagging non-zero terms

kmokstad:
And call flagNonZeroEq() also for the SAM-less assembly methods (used in L2-projection v2).


kmokstad:

kmokstad:

PR #890, opened 2026-06-23
Fix: 3D edge 8 mapping in GlobalNodes::getBoundaryNodes

akva2:
Edge 8 (TOP|EAST) was incorrectly mapped to TOP|WEST (duplicate of edge 7).


akva2:

PR #891, opened 2026-06-23
Couple of fixes for multipatch adaptivity

akva2:
Two copilot identified issues;
1) need to loop again if domain is extended
2) improve getBoundaryCovered - use getOverlappingNodes to also grab detect nodes further "in" from the boundary, which is in particular important for higher polynomial order.


akva2:

PR #897, opened 2026-07-07
Option to consider newly activated elements stress-free

kmokstad:
This is a continuation of 6cc02418b597d6aaf731971f9482a93c1039148f which was the last (unrelated) commit in #896.
When explicitly moving the control points to avoid fictitious large deformations in newly activated elements, we mus also update the control points in associated patches not-yet-activated, sharing common node number with the updated point. otherwise the model will exhibit gaps when adding more and more patches and the simulation results will probably be wrong as well.


kmokstad:

PR #904, opened 2026-08-21
Changed: Return type of virtual method NormBase::initIntegration()

kmokstad:
This is used by integrands with internal integration point buffers, to check that the specified number of quadrature points does not change when integrating norms.


kmokstad:
jenkins build this with downstreams ifem-elasticity=232 please

kmokstad:
jenkins build this with downstreams ifem-elasticity=232 ifem-openfrac=109 please

PR #906, opened 2026-09-16
added: Stress formulation for the Piola mapped laplacian operator

akva2:
Allow for stress formulation in Stokes with piola


akva2:
jenkins build this with downstreams ifem-stokes=351 please

PR #908, opened 2026-09-17
add: support for weak conditions with Piola mapped RT

akva2:
For Piola there are two complications for boundary conditions;

1) With fully-strong dirichlet conditions, we have to prescribe the values in the piola mapped space. This does not map to regular component-based dirichlet enforcement, Piola mixes the components.
2) With weak conditions we have to only prescribe the normal component. This requires pullback to Piola space, again requiring access to both conditions even though it eventually boils down to single-component conditions
Encode this using existing conventions as
```xml
<dirichlet set="Boundary" direction="0" type="anasol"/>
```
The direction usage mirrors the robin and neumann input syntax; direction="0" is normal.


akva2:
jenkins build this with downstreams ifem-stokes=354 please

src/ASM/ASMbase.h
```diff
@@ -517,6 +517,16 @@ class ASMbase
   bool hasTimeDependentDirichlet(const std::map<int,RealFunc*>& func,
                                  const std::map<int,VecFunc*>& vfunc);
 
+  //! \brief Flags that a Dirichlet code prescribes the normal component.
+  //! \param[in] code The in-homogeneous Dirichlet condition code
+  //! \details The prescribed value is then the normal velocity, and for a
```
kmokstad, line 523: Think you need an empty comment line, i.e., `//!`, before the `\details` tag, or else doxygen will think the details is associated with the `code` input parameter only, and not the method itself.

Alternatively, but the details in the C-file in front of the method definition.

src/SIM/SIMinput.C
```diff
@@ -549,6 +549,17 @@ bool SIMinput::parseBCTag (const tinyxml2::XMLElement* elem)
       utl::getAttribute(elem,"axes",axes,true);
       utl::getAttribute(elem,"component",comp);
       utl::getAttribute(elem,"comp",comp);
+      // As for tractions, direction 0 denotes the normal direction. For a
+      // div-compatible basis the normal component is carried by one basis
+      // alone, namely the one whose parameter direction the boundary is
+      // constant in, so which basis that is depends on the boundary itself.
```
kmokstad: Think I know the meaning, but a bit creaky written in my opinion. Better with:
"..., namely the one whose parameter is constant on the boundary, so which ..." ?

src/SIM/SIMinput.C
```diff
@@ -1443,7 +1454,23 @@ size_t SIMinput::setPropertyType (int code, Property::Type ptype,
...
+        if (basis >= 0)
+          p->basis = basis;
+        else if (ASMbase* pch = this->getPatch(p->patch); pch &&
+                 abs(p->ldim)+1 == pch->getNoParamDim())
+          // Normal-direction condition, see parseBCTag. The normal component
+          // on a boundary which is constant in one parameter direction is
+          // carried by the basis of that direction.
+          // Negated to mark it as a normal-direction condition; the
+          // simulator normalises it in preprocessA once it has seen it.
+          p->basis = -((abs(p->lindx)+1)/2);
```
kmokstad: Unnecessary paranthesis?  `-(abs(p-lindx)+1)/2;` should give the same result.

src/SIM/SIMinput.C
```diff
@@ -1443,7 +1454,23 @@ size_t SIMinput::setPropertyType (int code, Property::Type ptype,
...
+        else
+        {
+          std::cerr <<" *** SIMinput::setPropertyType: Direction 0 is only"
+                    <<" available on a boundary of one less dimension than"
+                    <<" the patch."<< std::endl;
```
kmokstad: " the patch itself."

src/SIM/SIMbase.C
```diff
@@ -338,8 +338,18 @@ bool SIMbase::preprocessC (const IntVec& ignored, bool fixDup, double time0)
...
         {
+          // A negated basis index marks a normal-direction condition, see
+          // SIMinput::setPropertyType. The basis it resolved to is the one
+          // carrying the normal component, and the patch is told that the
+          // prescribed value for this code is a normal velocity.
+          char pbasis = p.basis;
+          if (pbasis < 0)
+          {
+            pbasis = -pbasis;
+            myModel[p.patch-1]->addNormalDirichlet(abs(code));
```
kmokstad: Don't need the `abs()` here, since you already do that inside the `addNormalDiriclet()` method.

src/ASM/ASMs2D.h
```diff
@@ -134,13 +134,15 @@ class ASMs2D : public ASMstruct, public ASM2D
...
+    int                dir;   //!< Parameter direction of the boundary {-2,-1,1,2}
     int                dof;   //!< Local DOF to constrain along the boundary
     int                code;  //!< Inhomogeneous Dirichlet condition code
     std::vector<Ipair> nodes; //!< Nodes subjected to projection on the boundary
 
     //! \brief Default constructor.
-    DirichletEdge(Go::SplineCurve* sc = nullptr, int d = 0, int c = 0)
-    : curve(sc), dof(d), code(c) {}
+    DirichletEdge(Go::SplineCurve* sc = nullptr, int dr = 0,
+                  int d = 0, int c = 0)
```
kmokstad: Purely cosmetic, but put the `dr` argument on L144 too together with the other `int` arguments.

src/ASM/ASMs2D.C
```diff
@@ -1192,6 +1195,141 @@ void ASMs2D::setNodeNumbers (const IntVec& nodes)
...
+
+  double par[2] = { 0.0, 0.0 };
+  par[ndir-1] = dedge.dir < 0 ? (ndir == 1 ? geo->startparam_u()
+                                           : geo->startparam_v())
+                              : (ndir == 1 ? geo->endparam_u()
+                                           : geo->endparam_v());
+
+  // The prescribed value converted to a coefficient of the reference basis
+  std::vector<Go::Point> pts(3);
+  auto&& refValue = [&](double t, double& value) -> bool
```
kmokstad: A lambda that captures everything (I guess the `[&]` means that) and is only used at one place does not increase the readability in my opinion. Then it is better to inline the code where it is used directly instead. When I see a lambda definition, I automatically think it is for reducing code duplication, but when I realize it is only used once I get disappointed ;)

src/ASM/ASMs2D.C
```diff
@@ -1192,6 +1195,141 @@ void ASMs2D::setNodeNumbers (const IntVec& nodes)
...
+      const double dS = tv.length();
+      if (dS <= 0.0)
+      {
+        std::cerr <<" *** ASMs2D::projectNormalDirichlet: Degenerate boundary"
+                  <<" at parameter "<< t << std::endl;
+        return false;
+      }
+      value = nsign * dS * (tangent ? sf->timeDerivative(Xt) : (*sf)(Xt));
+    }
+    else
```
kmokstad:
```
else if (vf)
```

src/ASM/ASMs2D.C
```diff
@@ -1192,6 +1195,141 @@ void ASMs2D::setNodeNumbers (const IntVec& nodes)
...
+    {
+      // The whole velocity is prescribed, so pull it back to the reference
+      // basis and keep the component this boundary constrains. The adjugate
+      // of the jacobian is its inverse times the determinant, which is
+      // exactly the factor of the contravariant Piola transform, leaving no
+      // determinant and no inverse to compute.
+      const Vec3 u = tangent ? vf->timeDerivative(Xt) : (*vf)(Xt);
+      value = ndir == 1 ? dXdv.y*u.x - dXdv.x*u.y
+                        : dXdu.x*u.y - dXdu.y*u.x;
+    }
```
kmokstad:
```
else
  return false;
```

src/ASM/ASMs2D.C
```diff
@@ -18,12 +18,15 @@
 #include "GoTools/geometry/RectDomain.h"
 
 #include "ASMs2D.h"
+
+#include <memory>
```
kmokstad, line 22: Do we need this one here? Can't see there is any new spart-pointer usage or such things added in this commit.

src/ASM/ASMs2D.h
```diff
@@ -390,6 +392,18 @@ class ASMs2D : public ASMstruct, public ASM2D
...
+  //! \param[in] dedge The boundary to interpolate along
+  //! \param[in] sf Prescribed normal velocity, if given as a scalar function
+  //! \param[in] vf Prescribed velocity, if given as a vector function
+  //! \param[in] time Current time
+  //! \param[in] tangent If \e true, use time-derivatives of prescribed values
+  //! \return Spline curve with the resulting control point values
+  Go::SplineCurve* projectNormalDirichlet(const DirichletEdge& dedge,
+                                          const RealFunc* sf,
+                                          const VecFunc* vf,
+                                          double time, bool tangent) const;
```
kmokstad: Isn't this method relevant for mixed only with the div-compatible basis? Then consider if not ASMs2Dmx is not the right place for it, maybe with a virtual interface in ASMs2D.

src/ASM/ASMbase.C
```diff
@@ -1265,6 +1265,9 @@ bool ASMbase::updateDirichlet (const std::map<int,RealFunc*>& func,
       return false;
     }
 
+    if (this->isNormalDirichlet(cit.second))
+      continue; // Prescribed by the patch, direct evaluation would be wrong
```
kmokstad, line 1269: This check can be put before L1260 since we then don't need the `inod` index and can save the look-up.

src/ASM/ASMs2D.C
```diff
@@ -860,42 +860,64 @@ void ASMs2D::constrainEdge (int dir, bool open, int dof, int code, char basis)
...
     {
     case  1: // Right edge (positive I-direction)
       node += n1-1;
     case -1: // Left edge (negative I-direction)
       if (!open)
-	this->prescribe(node,dof,bcode);
+      {
+        this->prescribe(node,dof,bcode);
+        if (normalBC)
+          dirich.back().nodes.emplace_back(1,node);
```
kmokstad, line 877: Thought it then should be:
```
        if (normalBC)
          dirich.back().nodes.emplace_back(1,node);
        else
          this->prescribe(node,dof,bcode);
```
since the `prescribe()` call will add an MPC for the same DOF setting its value directly.

src/ASM/ASMs2D.C
```diff
@@ -860,42 +860,64 @@ void ASMs2D::constrainEdge (int dir, bool open, int dof, int code, char basis)
...
+        if (this->prescribe(node,dof,-code) == 0 && code > 0)
+          dirich.back().nodes.emplace_back(i2,node);
       }
       if (!open)
-	this->prescribe(node,dof,bcode);
+      {
+        this->prescribe(node,dof,bcode);
+        if (normalBC)
+          dirich.back().nodes.emplace_back(n2,node);
+      }
```
kmokstad, line 893: same here

src/ASM/ASMs2D.C
```diff
@@ -860,42 +860,64 @@ void ASMs2D::constrainEdge (int dir, bool open, int dof, int code, char basis)
...
     case  2: // Back edge (positive J-direction)
       node += n1*(n2-1);
     case -2: // Front edge (negative J-direction)
       if (!open)
-	this->prescribe(node,dof,bcode);
+      {
+        this->prescribe(node,dof,bcode);
+        if (normalBC)
+          dirich.back().nodes.emplace_back(1,node);
+      }
```
kmokstad, line 904: and here

src/ASM/ASMs2D.C
```diff
@@ -860,42 +860,64 @@ void ASMs2D::constrainEdge (int dir, bool open, int dof, int code, char basis)
...
+        if (this->prescribe(node,dof,-code) == 0 && code > 0)
+          dirich.back().nodes.emplace_back(i1,node);
       }
       if (!open)
-	this->prescribe(node,dof,bcode);
+      {
+        this->prescribe(node,dof,bcode);
+        if (normalBC)
+          dirich.back().nodes.emplace_back(n1,node);
+      }
```
kmokstad, line 919: and here as well

src/ASM/ASMbase.h
```diff
@@ -517,15 +517,16 @@ class ASMbase
...
-  bool isNormalDirichlet(int code) const
-  { return normalDirich.find(abs(code)) != normalDirich.end(); }
+  //! \details The basis is then Piola mapped, so the prescribed value lives in
+  //! the physical frame while the degrees of freedom are coefficients of the
+  //! reference basis. The patch has to pull the value back before prescribing
+  //! it, rather than use it as it stands.
+  void addPiolaDirichlet(int code) { piolaDirich.insert(abs(code)); }
+  //! \brief Returns \e true if the given code prescribes a physical velocity.
+  bool isPiolaDirichlet(int code) const
+  { return piolaDirich.find(abs(code)) != piolaDirich.end(); }
```
kmokstad, line 529: OK, so here changed stuff added in a previous commits. But think my comments there still apply.

kmokstad (commented):
This was a bit overwhelming. I guess the AI-agents has generated many of the comments and commit logs? Have tried to address a few things as a human reader. Also consider, we do have support for Dirichlet BC's on local axis directions (normal and/or tangent) from before, so this is perhaps complementary to that, but probably not totally equivalent.

src/ASM/ASMs2D.C
akva2: it's strictly needed, as there is pre-existing smart pointer use, we only get it transitionally.

akva2:
jenkins build this with downstreams ifem-stokes=354 please

akva2:
jenkins build this with downstreams ifem-stokes=354 please

src/ASM/ASMs2D.C
kmokstad: ok, it should have been added before then and not related to this patch as such (the way I see it).

akva2: yes. but it was flagged and is correct, so figured it's okay to include it.

